PR #25, opened 2021-12-22
added appimage icon

adam4nj:


DavidoTek:
Good idea. I changed the height of both to 56px so it doesn't look blurry.

PR #57, opened 2022-04-18
Added polish translation

StarterX4:


DavidoTek:
Thanks for the translation! I will include it in the next release.

Would be great if you could translate the [appdata summary](https://github.com/DavidoTek/ProtonUp-Qt/blob/933191a62a46c640b031f4005e3b59672fdb6a5f/share/metainfo/net.davidotek.pupgui2.appdata.xml#L7) and the [description for the compatibility tools in line 11](https://github.com/DavidoTek/ProtonUp-Qt/tree/main/share/pupgui2/ctmods) too.

i18n/pupgui2_pl.ts
```diff
@@ -0,0 +1,370 @@
...
+    </message>
+    <message>
+        <location filename="../pupgui2/pupgui2aboutdialog.py" line="61"/>
+        <source>system (restart required)</source>
+        <translation>Systemowy (wymagany restart)</translation>
+    </message>
+    <message>
+        <location filename="../pupgui2/pupgui2aboutdialog.py" line="70"/>
+        <source>Update Steam game list</source>
+        <translation>Zaktualizuj liste gier Steam</translation>
```
RDKRACZ, line 160: nitpick: `liste` should use the letter "ę" instead.

StarterX4: done

DavidoTek:
Great, thanks!

PR #81, opened 2022-06-20
it's known as GE-Proton now, adjust readme

LiamDawe:
Should probably use the correct naming :)


DavidoTek:
> (previously known as Proton-GE)

I'd omit that part to keep it simple.

Should we change the name inside ProtonUp-Qt as well? A "replace all" should do it...

LiamDawe:
I would say to replace all instances of it yes. I did try looking to find in the code where it is but couldn't find it so thought I would at least get the ball rolling here.

About the "previously" though, that's your call :)

PR #97, opened 2022-07-23
feat: add support for DXVK git version.

### LINKED ISSUE #96: DXVK and TKG git (Nightly CI)
DXVK is having a big update in next release, for users who want to try out the new features first this can be useful for them.

We can use GitHub API to get repository commit and use https://nightly.link to download the CI file (since GitHub requires signing in to download CI artifacts).

Manually downloading it does the trick, but I just like it being integrated into ProtonUp-Qt.

**Additional context**
I may be able to write that feature and make a PR, but it'll take some time to do it.

teppyboy:
This PR adds support for DXVK git version in master branch as "DXVK (nightly)" in compatibility tool list, fixes #96. It'll install DXVK nightly as `dxvk-master-<commit long hash>` in Lutris.

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_z2dxvknightly.py
+++ b/pupgui2/resources/ctmods/ctmod_z2dxvknightly.py
@@ -1,0 +1,151 @@
+# pupgui2 compatibility tools module
+# DXVK for Lutris (nightly version): https://github.com/doitsujin/dxvk/
+# Copyright (C) 2022 DavidoTek, partially based on AUNaseef's protonup
+
+import os, shutil, zipfile, requests
+from PySide6.QtCore import *
+
+CT_NAME = 'DXVK (nightly)'
+CT_LAUNCHERS = ['lutris']
+CT_DESCRIPTION = {}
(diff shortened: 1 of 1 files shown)


teppyboy:
Also not related to this PR but with the ability to install git artifacts, we can now also install `wine-tkg-git` (and its variants like valvexbe, etc). If you like that feature, HMU for another PR with that feature :P

DavidoTek:
Thanks! I will test it and see how I can add an "advanced mode" and then merge it.

> HMU for another PR with that feature

Also, should us give more options, right? If you want, you can help implement that.

teppyboy:
> Thanks! I will test it and see how I can add an "advanced mode" and then merge it.
> 
> > HMU for another PR with that feature
> 
> Also, should us give more options, right? If you want, you can help implement that.

yep, all in that "advanced mode", but for now we need to fix proton tkg because tkg switched to CI builds

DavidoTek:
You're right. Can you try to add your code from DXVK to the existing TKG module, so that it will offer both the old GitHub releases as well as the new CI builds? Or do you think having only the new builds would be enough?

teppyboy:
> You're right. Can you try to add your code from DXVK to the existing TKG module, so that it will offer both the old GitHub releases as well as the new CI builds? Or do you think having only the new builds would be enough?

I think we should have both new builds and old builds, about the code I'll try to add ASAP!

teppyboy:
Check #99 for Proton Tkg support I mentioned earlier

PR #99, opened 2022-07-25
feat: support Proton Tkg CI builds 

teppyboy:
It was a tough work, but it can download both new builds from CI and old builds from GitHub releases now. Currently CI builds is named after the workflow runs id (which doesn't really make sense for end user except that the largest number is the latest), and it's using Proton with Valve bleeding edge Wine (because I think Proton is best suited with Valve's Wine which contain patches for gaming), if you want to switch to Proton with Wine latest commit, tell me and I'll change it :P (Although I don't think there will be benefits in changing, since Wine master is just bleeding edge as Valve Wine)

Proton CI version when installed will be named as `proton_tkg_experimental.bleeding.edge.x.x.xxxx.YYYYMMDD`, this name is from the CI, I can't do anything to get this name before extracting (since the CI archive is literally a .tar archive packed inside a .zip archive called `proton-tkg-build.zip` LOL) and have no plan to since this name looks ok for me as a proton package and not really possible to.

Also I got [rate limited](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting) by GitHub, so can't test properly but at least installing both new proton from CI and old proton from releases and the button to show info works (tested twice before getting rate limited lol).

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_protontkg.py
+++ b/pupgui2/resources/ctmods/ctmod_protontkg.py
@@ -202,7 +202,7 @@ class CtInstaller(QObject):
         Get link with info about version (eg. GitHub release page)
         Return Type: str
         """
-        if self.__get_artifact_from_commit(version):
+        if self.__get_artifact_from_id(version):
             return self.CT_INFO_URL_CI + version
 
         return self.CT_INFO_URL + version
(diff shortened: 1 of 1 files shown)


DavidoTek:
Thank you very much!!

> Currently CI builds is named after the workflow runs id ([...] largest number is the latest)
> ...named as `proton_tkg_experimental.bleeding.edge.x.x.xxxx.YYYYMMDD`

I guess that should not be a problem. We can change it later if we want.

> it's using Proton with Valve bleeding edge Wine

Yes, I think that's what we want.

> test properly

I will do some more testing, but the code looks promising.

PR #111, opened 2022-08-21
make single instance check more accurate

yurikaka:
after a reboot, the previous pid written in the lockfile can be taken by other process, in this case, user will keep getting the "Second instance of ProtonUp-Qt found!" but actually there's no other instance. Setting custom process name and checking the process name will make it more accurate.

### DIFF
--- a/pupgui2/util.py
+++ b/pupgui2/util.py
@@ -322,7 +322,7 @@ def single_instance() -> bool:
         f = open(lockfile, 'r')
         pid = f.read()
         f.close()
-        cmdline_file = os.path.join('/proc/' + pid + '/cmdline')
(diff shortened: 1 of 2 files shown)


yurikaka:
> Thanks, that's an important addition!
> 
> The solution requires two additional dependencies though, maybe you can get rid of them.
> 
> 1. Setting a custom process name isn't needed as the command should contain `pupgui2` already (e.g. `python -m pupgui2`)
> 2. The command can be read from the file `/proc/<pid>/cmdline`

yes, in most cases, the cmdline should be enough
user can still change cmdline if they run it manually, but if they do that, they should be able to figure it out themselves

PR #134, opened 2022-11-06
Add Comment to .desktop file

sonic2kk:
Adds a `Comment` line to the `.desktop` file which will show a small description in the Application Launcher. This Comment and its translations were taken from the existing translation strings in the `net.davidotek.pupgui2.appdata.xml`.

Here is an example of how it looks on the KDE Plasma Application Launcher (the same one the Steam Deck uses as far as I know):

image

I am not sure how this `.desktop` file is used *exactly*, but I would assume that it is simply copied on installation of the Flatpak/AppImage. So I don't think this needs updated anywhere else, but if it does please let me know.

Thanks!


DavidoTek:
Thanks, useful addition.

I think you have to use short language codes like `de` instead of `de_DE`. Using the underscore is only used if there are larger differences, like in Chinese. (Both should work afaik but I think the first one is more common)

sonic2kk:
Really sorry I didn't get a chance to edit this today! But I just had a look at some other .desktop files on my system and you're absolutely correct. I'll get it updated ASAP :-)

DavidoTek:
No worries, here you go.

PR #137, opened 2022-11-16
Use QTableView in CtInfoDialog

sonic2kk:
This replaces the QListView in the CfInfoDialog with a QTableView. For Steam games, the column headings are the AppID and the Game Name. For Lutris, it uses the slug and the game name.

image

This has a couple of benefits:
- Matches more closely with the GamesListDialog
- Allows for sorting entries by AppID/Slug and name using the column headings
- In my opinion, looks a little neater :sweat_smile: 

I could not test Lutris as I use Lutris-git and this does not seem to be returning game/configuration information correctly (the `cfg` value is `{}`, and "Games using compatibility tool:" does not show a value at all -- not even 0). But I will open a separate issue for that when I have time :smiley: I only noticed when working on this PR.

Thanks!


DavidoTek:
Thanks!

> ```python
> self.ui.listGames.setItem(i, 0, QTableWidgetItem(self.tr #...
> self.ui.listGames.setItem(i, 1, QTableWidgetItem(self.tr #...
> ```
> lines 67, 68, 79, 80

You can omit the `self.tr` here. Only needed for translating static strings 😉

Just a small thing, but maybe rename `list_games_item_double_clicked` to `list_games_cell_double_clicked`.

> In my opinion, looks a little neater 😅

Agreed.

> I could not test Lutris

Code looks fine. I will do some testing tomorrow.

sonic2kk:
Ah! Sorry for the delay, and thanks for merging in the tweaks :heart:

PR #140, opened 2022-11-18
Fix DXVK 2.0 downloading DXVK native release

### LINKED ISSUE #138: DXVK 2.0
Download for DXVK 2.0 for Steam is pullig the wrong folder structure.
DXVK 2.0 Async is downloading corrrectly....

sonic2kk:
Fixes #138 

DXVK 2.0's release ships a new tarball for DXVK native. The logic in the DXVK ctmod doesn't account for this (as I believe previous releases just shipped the regular DXVK release). As a result, when downloading DXVK 2.0 with ProtonUp-Qt, the different folder structure incorrectly shows `lib` and `lib32` - the two folders in the DXVK native tarball - as compatibility tools.

This adds a check for the GitHub asset name having `native` anywhere in it, as with a name like that we can assume (at least currently) that this will only refer to the DXVK native builds.

image

According to the linked issue, DXVK Async doesn't have this issue. I am not sure about DXVK nightly, either. If we need to make changes here too, I can add those as well :-)

Thanks!

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_z0dxvk.py
+++ b/pupgui2/resources/ctmods/ctmod_z0dxvk.py
@@ -82,7 +82,7 @@ class CtInstaller(QObject):
 
         values = {'version': data['tag_name'], 'date': data['published_at'].split('T')[0]}
         for asset in data['assets']:
-            if asset['name'].endswith('tar.gz'):
+            if asset['name'].endswith('tar.gz') and 'native' not in asset['name']:
                 values['download'] = asset['browser_download_url']
                 values['size'] = asset['size']
         return values


DavidoTek:
Thank you soo much, that was a quick fix (even before I could take a look at it :smile:)!  :heart_decoration: 

I will merge it (and also look into the other issues) when I'm home.

sonic2kk:
Awesome :smile: In my haste I forgot to actually give some links, oops...

Here is the DXVK 2.0 releases page: https://github.com/doitsujin/dxvk/releases

You can see the asset with `native` in the name is the Native DXVK release, and if you download it, it has a different folder structure.

## DXVK
image

## DXVK Native
image

I suppose ProtonUp-Qt *could* have support for downloading DXVK Native, however this is really only intended for developers and not really for users. There used to be a note about this somewhere in the DXVK documentation I believe, but I actually cannot find anything about this now. Either way, I do not think it's something intended for *users*.

Issue #73, opened 2022-05-29
[Request] VKD3D


Sethox:
**NOTE:** I searched first between the issues (if it was already requested).

So the request is just that if ProtonUp-Qt can add [VKD3D](https://github.com/HansKristian-Work/vkd3d-proton) project up for Steam. 

Other than that, great project and I hope the best for the neat features it already has.

teppyboy:
Steam uses Proton, and Proton already has VKD3D. Do you meant VKD3D for Lutris?

sonic2kk:
Got a WIP implementation for this for Lutris - vkd3d-proton from [Hans-Kristan / "Valve"](https://github.com/HansKristian-Work/vkd3d-proton/), and from [Lutris](https://github.com/lutris/vkd3d) (who seem to maintain a mostly/completely plain fork?).

Will hopefully get a PR up for this soon :-)

sonic2kk:
Should be implemented with #142 :smiley:

Issue #87, opened 2022-07-05
When Downloading the latest version of Proton GE (7-24), the download progress freezes at 93%


WhiSach:
**Describe the bug**  
When Downloading the latest version of Proton GE (7-24), the download progress freezes at 93%. 
 
**To Reproduce**  
1. Go to "Add Version"
2. Select the "GE-Proton7-24" version of GE Proton
3. Select Install
4. As you wait, the install will progress until it freezes at 93%

**Screenshots**  
![Screenshot from 2022-07-05 17-07-40](https://user-images.githubusercontent.com/47545327/177319441-4bb23e1b-cd6e-4585-8de1-3fdf69001c19.png)

**Desktop (please complete the following information):**  
- Linux Mint 20.3
- Flatpak
- Steam install from the official Steam Website

**Terminal output**  
Failed to load module "xapp-gtk3-module"
Loaded ctmod Proton-GE
Loaded ctmod Wine-GE
Loaded ctmod Boxtron
Loaded ctmod Kron4ek Wine-Builds Vanilla
Loaded ctmod Lutris-Wine
Loaded ctmod Luxtorpeda
Loaded ctmod Proton Tkg
Loaded ctmod Roberta
Loaded ctmod DXVK
Loaded ctmod DXVK Async
qt.pysideplugin: Environment variable PYSIDE_DESIGNER_PLUGINS is not set, bailing out.
qt.pysideplugin: No instance of QPyDesignerCustomWidgetCollection was found.
[Errno 28] No space left on device

DavidoTek:
Most likely a problem with the temporary directory, I have to make some improvements for the Flatpak version.
Can you send me the output of running `df -h` in the Terminal?

PS: The AppImage version should work fine if the problem is caused by the Flatpak temporary directory.

Whipshadow:
**After running `df -h` with the frozen download**
```
Filesystem      Size  Used Avail Use% Mounted on
udev            1.9G     0  1.9G   0% /dev
tmpfs           382M  1.7M  381M   1% /run
/dev/sda2       117G   66G   45G  60% /
tmpfs           1.9G  149M  1.8G   8% /dev/shm
tmpfs           5.0M  4.0K  5.0M   1% /run/lock
tmpfs           1.9G     0  1.9G   0% /sys/fs/cgroup
/dev/loop1       62M   62M     0 100% /snap/core20/1518
[...]

PR #142, opened 2022-11-22
Lutris: Add vkd3d sources

sonic2kk:
Implements #73 

## Overview
Adds support for downloading vkd3d for Lutris. Like with manual DXVK version downloads, this has to be entered manually in the Advanced Options for a Lutris game, but this is general Lutris usage and would apply if this was done manually as well :-)

Two sources are added here: [vkd3d-proton](https://github.com/HansKristian-Work/vkd3d-proton/) (the project used by Valve in Proton) and [vkd3d-lutris](https://github.com/lutris/vkd3d/) (a fork of this project with seemingly minimal changes, at least at the time of writing) - The latter is just a Lutris fork of upstream vkd3d-proton simply called "vkd3d" - This is slightly further behind than regular vkd3d-proton, and releases as a `.tar.xz` instead of a `.tar.zst` like vkd3d-proton does. 

The benefit of having both is the ability to use the latest vkd3d release from upstream vkd3d-proton  and using it "raw", in case there are any changes in the downstream Lutris build. Upstream is on v2.7, Lutris is on v2.6. A brief skim of the commits appears like there are no real changes in the fork, maybe they just host separately for packaging? I am not sure of the reason, but both should work. The folder structure is identical.

This change does **not** impact Steam, as this is irrelevant to Steam.

image
image

### Two Ctmods
I considered taking a similar approach to what I did with SteamTinkerLaunch: Having one "main" ctmod for vkd3d, and then having a separate one for vkd3d-lutris. Since I thought the code would be much the same, just different URLs.

It turns out the URLs are different, and some of the code needed for vkd3d-proton was different. The release asset extensions are different and the extraction and paths would be different. Since they are in reality two separate repositories, I figured having two separate ctmods made the most sense in the end and would be cleaner as there wouldn't be a bunch of conditional checks.

### Zstandard Library
Okay, the main "problem" with this implementation. Upstream vkd3d-proton is releases as a `.tar.zst` file, and this is **not** supported by the Python tarball library. There is a [stale request](https://bugs.python.org/issue37095) for this upstream, but I wouldn't hold my breath on this. And even so, ProtonUp-Qt does not use the latest Python version, so a version bump would be required to make use of that.

To implement this, I used the [Zstandard Python library](https://pypi.org/project/zstandard/) and mostly followed an implementation from a [GitHub Gist](https://gist.github.com/scivision/ad241e9cf0474e267240e196d7545eca). It basically uses this Zstandard library to extract the tar data to a temporary tarfile, which is then extracted properly using the tarfile library.

The problem is that this adds an extra dependency. Creating a PR that adds an extra dependency is probably not very helpful, and I am really sorry about that. I did investigate to see if there was a way to do this without pulling in an extra dependency, but that is how I came across the upstream Python issue.

If there is a way to fix this, I would be happy to remove this dependency.

## Remaining Issues
The only remaining issue with this PR is that the vkd3d downloads don't show up in the compatibility tool list. I figure the Lutris runner `vkd3d` directory just needs to be added somewhere in the code, but I couldn't figure out where to put it.

Apart from this issue, I am happy with the state of the PR and think it's ready for review.

<hr>

Phew, I was meant to take a break but I couldn't resist trying to tackle this :sweat_smile: As always, feedback is appreciated! :-)

### DIFF
--- a/pupgui2/pupgui2.py
+++ b/pupgui2/pupgui2.py
@@ -177,13 +177,13 @@ class MainWindow(QObject):
         self.ui.listInstalledVersions.clear()
         self.compat_tool_index_map = get_installed_ctools(install_directory())
 
-        # Launcher specific (Lutris): Show DXVK
+        # Launcher specific (Lutris): Show DXVK and vkd3d-proton
         if install_loc.get('launcher') == 'lutris':
             dxvk_dir = os.path.join(install_directory(), '../../runtime/dxvk')
-            for ct in get_installed_ctools(dxvk_dir):
-                if not 'dxvk' in ct.get_displayname().lower():
-                    ct.displayname = 'DXVK ' + ct.displayname
-                self.compat_tool_index_map.append(ct)
+            vkd3d_dir = os.path.join(install_directory(), '../../runtime/vkd3d')
+
+            self.get_installed_versions('dxvk', dxvk_dir)
+            self.get_installed_versions('vkd3d', vkd3d_dir)
 
         # Launcher specific (Steam): Number of games using the compatibility tool
         if install_loc.get('launcher') == 'steam' and 'vdf_dir' in install_loc:
@@ -210,6 +210,12 @@ class MainWindow(QObject):
         #    self.ui.btnShowGameList.setVisible(True)
         else:
             self.ui.btnShowGameList.setVisible(False)
+
+    def get_installed_versions(self, ctool_name, ctool_dir):
+        for ct in get_installed_ctools(ctool_dir):
+            if not ctool_name in ct.get_displayname().lower():
+                ct.displayname = f'{ctool_name} {ct.displayname}'
+            self.compat_tool_index_map.append(ct) 
 
     def install_compat_tool(self, compat_tool):
         """ install compatibility tool (called by install dialog signal) """
--- a/pupgui2/resources/ctmods/ctmod_vkd3dlutris.py
+++ b/pupgui2/resources/ctmods/ctmod_vkd3dlutris.py
@@ -2,10 +2,8 @@
 # vkd3d-lutris for Lutris: https://github.com/lutris/vkd3d/
 # Copyright (C) 2022 DavidoTek, partially based on AUNaseef's protonup
 
-import os, shutil, tarfile, requests, tempfile
-import zstandard
+import os, shutil, tarfile, requests
 
-from pathlib import Path
 from PySide6.QtCore import *
 
 CT_NAME = 'vkd3d-lutris'
@@ -120,16 +118,14 @@ class CtInstaller(QObject):
 
         vkd3d_dir = os.path.join(install_dir, '../../runtime/vkd3d')
 
-        destination = temp_dir
-        destination += data['download'].split('/')[-1]
-        destination = destination
+        temp_download = os.path.join(temp_dir, data['download'].split('/')[-1])
 
-        if not self.__download(url=data['download'], destination=destination):
+        if not self.__download(url=data['download'], destination=temp_download):
             return False
 
         if os.path.exists(vkd3d_dir + 'vkd3d-' + data['version'].lower()):
             shutil.rmtree(vkd3d_dir + 'vkd3d-' + data['version'].lower())
-        tarfile.open(destination, "r:xz").extractall(vkd3d_dir)        
+        tarfile.open(temp_download, "r:xz").extractall(vkd3d_dir)        
 
         self.__set_download_progress_percent(100)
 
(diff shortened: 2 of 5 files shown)


DavidoTek:
> Two Ctmods

I think it makes sense to have two ctmods.

> Zstandard Library

I don't think we have an alternative to using the Zstandard Python library. It will increase the package a bit, but the binary inside the package contains debug info which can be stripped to reduce the size to around 940KB (Flatpak and AppImage Builder automatically strip the binaries AFAIK). I think that is fine

> The only remaining issue with this PR is that the vkd3d downloads don't show up in the compatibility tool list

It's a bit hacky, but DXVK tools are separately added to the list here: https://github.com/DavidoTek/ProtonUp-Qt/blob/99b0c15a15288a256dd847aa9c1b33258104f424/pupgui2/pupgui2.py#L181-L186  
(If this gets "too hacky" we may need add a `get_installed_versions` method to each ctmod, but I think it's fine for now)

-------------------

> Phew, I was meant to take a break but I couldn't resist trying to tackle this :sweat_smile: As always, feedback is appreciated! :-)

I know that all to well. Once you've started something, it's hard to take a break for a while until every idea is put down :smile: As always, thanks!!

sonic2kk:
Thank you! I did an initial draft by just copying the code to display DXVK in the list, but it seemed very repetitive and also fairly straightforward to break out into a function, so I did just that :-)

```python
def get_installed_versions(self, ctool_name, ctool_dir):
    for ct in get_installed_ctools(ctool_dir):
        if not ctool_name in ct.get_displayname().lower():
            ct.displayname = f'{ctool_name} {ct.displayname}'
        self.compat_tool_index_map.append(ct)
```

I created a function called `get_installed_versions` that takes the `ctool_name` (i.e., `dxvk`, `vkd3d`) and  the `ctool_dir` and adds them to the ctool list. This seems to work, but if there is any problem with this I can revert this and just do a straight copy of the DXVK code like:

```python
# DXVK
for ct in get_installed_ctools(dxvk_dir):
    if not 'dxvk' in ct.get_displayname().lower():
        ct.displayname = 'DXVK ' + ct.displayname
    self.compat_tool_index_map.append(ct)

# vkd3d-proton
for ct in get_installed_ctools(vkd3d_dir):
    if not 'vkd3d' in ct.get_displayname().lower():
        ct.displayname = 'vkd3d ' + ct.displayname
    self.compat_tool_index_map.append(ct)
```

DavidoTek:
Thanks, that looks good.
I will take a final look if something else needs to be considered and merge it tomorrow.

DavidoTek:
I'm not sure how well `tempfile` works inside Flatpaks. There was an issue a while back where the [temporary directory has no space left](https://github.com/DavidoTek/ProtonUp-Qt/issues/87).

I will test it and if needed change the folder, that should be simple to do.

PS: Is tempfile actually creating a file on disk or using some sort of ramdisk?

sonic2kk:
Ah sorry, that could be a problem, I didn't even think of that!

From research, `tempfile` creates an actual file I believe. The default location varies between systems but on my system (and I presume the vast majority of distros) it uses `/tmp`. You can change this with `tempfile.tempdir` (https://docs.python.org/3.8/library/tempfile.html#tempfile.tempdir). It also seems like `TemporaryFile` can take a `dir` parameter: https://docs.python.org/3.8/library/tempfile.html#tempfile.TemporaryFile 

So if there is a preferred Flatpak location or a better location, we could set it in either of these ways. Maybe setting the tempfile location would be better if we want different locations for Flatpaks vs everything else, and passing it in the `TemporaryFile` constructor could be the cleanest if we want the same location all the time :-)

This StackOverflow answer has some more background too: https://stackoverflow.com/questions/18280245/where-does-python-tempfile-writes-its-files

DavidoTek:
All right. I will modify the code later.

DavidoTek:
I made some changes to the code:
- I removed the `tempfile` and `pathlib` dependency and made sure the temporary `.tar` file is created in the `temp_dir` provided by the `get_tool` method
- I cleaned the code a bit, for instance removed `destination=destination` which is remnant from the original ProtonUp-Qt ctmods (and is present in almost all ctmod :laughing:)

Will merge. Thanks! :tada:

sonic2kk:
> I removed the `tempfile` and `pathlib` dependency and made sure the temporary `.tar` file is created in the `temp_dir` provided by the `get_tool` method

Nice! :smile: 

> I cleaned the code a bit, for instance removed `destination=destination` which is remnant from the original ProtonUp-Qt ctmods (and is present in almost all ctmod :laughing:)

Hehe, I noticed this but I didn't want to touch too much of the other code *just* in case it was added as a specific workaround or for another reason.

Thanks and glad I could help out some more :smile:

Issue #152, opened 2022-12-01
Add Steam-Play-None compatibility tool


puddydoo:
Make [Steam-Play-None](https://github.com/Scrumplex/Steam-Play-None) installable from ProtonUp-Qt

DavidoTek:
That should be easily doable.

Just wondering, are there any other examples of games than `CrossCode` that benefit from `Steam-Play-None`?

puddydoo:
Apparently it helps with modded slay the spire where mod the spire has black screen issues on proton and the game crashes on steam linux runtime

sonic2kk:
There is no release for this tool, but the repository from `git` can simply be cloned and moved to `compatibilitytools.d`. A concern here could be file size / excess files, since it'll have the `.git` and potentially `.github` file.

Also, to my understanding, this is really only useful on Steam Deck where there is unfortunately no easy way to do what this compatibility tool does. On Linux Desktop, you can simply run your game's Linux port as-is. It might be worth noting that this is mainly useful on Steam Deck in the compatibility tool description.

DavidoTek:
> There is no release for this tool [...] the .git and potentially .github file.

We can download the repo as a tar.gz, that should not include the .git file. Not sure if it's best to download the `main` branch as we cannot go back this way but on the other hand I don't think there is much that will change.

https://github.com/Scrumplex/Steam-Play-None/archive/refs/heads/main.tar.gz

> mainly useful on Steam Deck

We can add it to the description. Maybe this should only be available in `advanced mode`.

PR #153, opened 2022-12-04
Add Steam-Play-None compatibility tool

DavidoTek:
Adds [Scrumplex's Steam-Play-None](https://github.com/Scrumplex/Steam-Play-None) compatibility tool.

Description:
```
Run Linux games as is, even if 
Valve recommends Proton for a game.
Created by Scrumplex.

Useful for Steam Deck.
```

Closes https://github.com/DavidoTek/ProtonUp-Qt/issues/152.

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_steamplaynone.py
+++ b/pupgui2/resources/ctmods/ctmod_steamplaynone.py
@@ -12,7 +12,7 @@ from PySide6.QtCore import QObject, QCoreApplication, Signal, Property
 
 CT_NAME = 'Steam-Play-None'
 CT_LAUNCHERS = ['steam', 'advmode']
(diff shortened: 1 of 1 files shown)


DavidoTek:
Just noticed, the internal/`display_name` is different from the folder name.

`display_name` | `None`
-------------|----------
folder name | `Steam-Play-None`

sonic2kk:
SteamTinkerLaunch has this too and I'm pretty sure we just hacked in a couple of hardcoded checks to get it to display "SteamTinkerLaunch" in the tool list and to manage the game list menu.

DavidoTek:
I've decided to change the internal name of Steam-Play-None from `none` to `Steam-Play-None` when installing the tool. This way it is consistent with the folder name and no modifications to ProtonUp-Qt need to be done.

As a side effect it's now also clear, when selecting a tool that `Steam-Play-None` is a tool whereas `None` sound more like "No tool used".

Are there any objections why this could be a bad idea?

PR #157, opened 2022-12-11
Bump steam library version

### LINKED ISSUE #155: Application wasn't able to open game list, now cannot reinstall
**Describe the bug**  
May be two bugs...

I was running protonupqt successfully for 2 months but when I went to check for new proton updates today the app would not open the game list when the button was pressed. I rebooted and it was the same. I'm using the dec 8 preview build on my steamdeck.

Next I attempted to remove and reinstall. The discover app aborts the install, complaining about the flatpak having a bad commit signature key with a bad date associated. Removing local data from previous install did not resolve.

**To Reproduce**  
1. Open app
2. Click 'show games list
3. Nothing happens

1. Search app in Discovery, install
2. See error

**Expected behavior**  
Loads games list for proton selection.

Installs via discovery without error.

**Screenshots**  
N/a

 - **Platform**: Steam Deck
 - **System**: SteamOS 3.4 preview Dec 8
 - **Version**: ProtonUp-Qt 2.7.6
 - **How did you install ProtonUp-Qt?**: Discover app store

**Additional context**  

**Terminal output**  
```
Open the Terminal app and type one of the following commands and press Enter (The app is called "Konsole" on Steam Deck):

If you installed ProtonUp-Qt via Flatpak (Discover on Steam Deck), type:
flatpak run net.davidotek.pupgui2

If you have downloaded the AppImage, navigate to the folder containing the AppImage file and type:
./ProtonUp-Qt*.AppImage

Try to replicate the bug and paste the text output of the command here afterwards.
```

sonic2kk:
Fixes #155.

The `steam` library for Python was updated yesterday to address an issue with the `parse_appinfo` function. The format of this changed on Steam's side and so the library had to be updated to accommodate this following commit https://github.com/ValvePython/steam/commit/abf65abff15385deb8ba05adaf4e312254fb435c.

Honestly I don't really understand what the changes here did but I can confirm it fixes the issue for me when running from source. Also from my armchair glance at the commit this should not have any breakages on Flatpak or require any sort of Python version bump. It is a fix but it doesn't seem to do anything *too* "magical" or extravegant that might break compatibility anywhere. My concern here is bumping from a minimum of `1.1.0` to a minimum of `1.4.4`. In testing *from source* I could not see anything that has broken (yet, anyway) but I'm also not sure what the previous version of the library I was using was. I would guess that this is probably fine but the changes from 1.1.0 to 1.4.4 may want to be reviewed with more scrutiny first.

Currently the `requirements.txt` mandate `steam>=1.1.0`, however since this new version of the `steam` library fixes a critical bug, I set the version as `>=1.4.4`, forcing this newer version since without it, various pieces of functionality will not work without this version or higher.

Opening this as a PR for transparency on the research process for this fix, as well as discussion. Merging a 3 character change on its own is rarely worth a PR but the commentary around the change might be worth it :-)

Thanks!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,6 @@ requests>=2.27.0
 vdf>=3.4
 inputs==0.5
 pyxdg>=0.27
-steam>=1.1.0
+steam>=1.4.4
 PyYAML==6.0
 zstandard>=0.19.0


DavidoTek:
Thanks!

Looks like they changed the beginning "magic" of the file for some reason and the library wasn't able to detect the file as a vdf.
Not sure why they would do this, maybe this is a new version. [Looks like they changed `'DV\\x07` to `(DV\\x07`](https://github.com/ValvePython/steam/issues/418) and `(` comes after `'` in the ASCII table after all.

> My concern here is bumping from a minimum of 1.1.0 to a minimum of 1.4.4. 

[It was using 1.3.0 before](https://github.com/DavidoTek/ProtonUp-Qt/blob/9f187e3a66ff31ee37a9f652a82738f57141c408/requirements.txt#L6), 1.4.4 should be fine. I changed it to 1.1.0 as it ~is~was the minimum version required.

> Opening this as a PR for transparency on the research process for this fix, as well as discussion. Merging a 3 character change on its own is rarely worth a PR but the commentary around the change might be worth it :-)

I think having a PR for discussions is the way to go. As you said, it will make the development of ProtonUp-Qt more transparent and as a side effect also provides an information base useful for other FOSS projects.

PR #159, opened 2022-12-17
Add ProtonDB.com status

DavidoTek:
This PR adds functionality to show the status from [ProtonDB](https://www.protondb.com/) for a game.

- It will download the status summary for all installed games on each startup of ProtonUp-Qt
- Currently, you can display the "ProtonDB tier" by hovering the SteamDeck Compatibilty in the game list
- TODO: Maybe display the status in a separate column?

Screenshot

### DIFF
--- a/pupgui2/constants.py
+++ b/pupgui2/constants.py
@@ -64,6 +64,7 @@ STEAM_API_GETAPPLIST_URL = 'https://api.steampowered.com/ISteamApps/GetAppList/v
 STEAM_APP_PAGE_URL = 'https://store.steampowered.com/app/'
 AWACY_GAME_LIST_URL = 'https://raw.githubusercontent.com/Starz0r/AreWeAntiCheatYet/master/games.json'
 LOCAL_AWACY_GAME_LIST = os.path.join(TEMP_DIR, 'awacy_games.json')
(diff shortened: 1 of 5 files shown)


sonic2kk:
ProtonDB also returns a "confidence" rating, this isn't very important to show in a column but it could be neat to show in the tooltip maybe? Just my thoughts :-) I also think having a separate column for the ProtonDB rating is a good idea, it would go hand-in-hand with the anti-cheat rating imo

DavidoTek:
I've added another column to the game list and modified the tooltip to show the confidence.

It looks like this now. Not sure about the colors/contrast, but looks fine with both Adwaita light and dark:
screenshot2

sonic2kk:
> If someone has 200 games in their Steam library that means 200 requests will be made to the protondb.com API.

Oh gosh, that is a scary thought! Some of us (*cough*) could have over 1,000 more than that which could be very problematic for requests haha

Having a button for this instead makes more sense.

> I think that API is meant for "internal" purposes.

I also think it is as well, I didn't even know about it until I saw that STL used it :sweat_smile: Though it seems to have been "public" for a while so I don't think there's any risk of it going private anytime soon or anything. Even Steam has a few undocumented but rate-limited endpoints as well

Issue #161, opened 2022-12-19
Add UI indicator for GitHub API limit


dysonsphere-startmail:
Please fill out following when reporting a new bug:

**Describe the bug**  
When choosing to install a compatibility tool the drop-down list does not populate with available versions

**To Reproduce**  
1. Click Add version
2. Choose a comparability tool

**Expected behavior**  
A list of versions of the selected compatability tool should be displayed in the version drop-down list.

**Screenshots**  
![Screenshot from 2022-12-19 13-36-21](https://user-images.githubusercontent.com/22195888/208498248-229be7ba-d3aa-44eb-b984-7f1a74689b29.png)

**Desktop (please complete the following information):**  
 - **Platform**: Laptop
 - **System**: Pop!_OS 22.04 LTS x86_64
 - **Version**: For example ProtonUp-Qt 2.7.6
 - **How did you install ProtonUp-Qt?**: Flatpak from the Pop!_Shop

**Terminal output**  
```ProtonUp-Qt 2.7.6 by DavidoTek. Build Info: DavidoTek Flathub build.
Python 3.10.6 (main, Nov 10 2011, 15:00:00) [GCC 12.1.0], PySide 6.4.0.1
Platform: Linux-6.0.6-76060006-generic-x86_64-with-glibc2.35
Gtk-Message: 13:45:56.375: Failed to load module "appmenu-gtk-module"
Gtk-Message: 13:45:56.415: Failed to load module "canberra-gtk-module"
Gtk-Message: 13:45:56.415: Failed to load module "canberra-gtk-module"
Qt: Session management error: None of the authentication protocols specified are supported
Loading locale en / en_CA
Loaded ctmod GE-Proton
Loaded ctmod Wine-GE
Loaded ctmod Boxtron
Loaded ctmod Kron4ek Wine-Builds Vanilla
Loaded ctmod Lutris-Wine
Loaded ctmod Luxtorpeda
Loaded ctmod Proton Tkg
Loaded ctmod Roberta
Loaded ctmod SteamTinkerLaunch
Loaded ctmod SteamTinkerLaunch-git
Loaded ctmod DXVK
Loaded ctmod DXVK Async
Loaded ctmod DXVK (nightly)
qt.pysideplugin: Environment variable PYSIDE_DESIGNER_PLUGINS is not set, bailing out.
qt.pysideplugin: No instance of QPyDesignerCustomWidgetCollection was found.
Gamepad error: No gamepad found.
```

[...]

PR #163, opened 2022-12-20
Add GitHub API rate limit warning

DavidoTek:
- Adds a warning that is printed to console when the GitHub API rate limit is exceeded
- Fixes crashes that occur when the rate limit is exceeded

See https://github.com/DavidoTek/ProtonUp-Qt/issues/161

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_00protonge.py
+++ b/pupgui2/resources/ctmods/ctmod_00protonge.py
@@ -10,6 +10,8 @@ import hashlib
 
 from PySide6.QtCore import QObject, QCoreApplication, Signal, Property
 
+from pupgui2.util import ghapi_rlcheck
+
 
 CT_NAME = 'GE-Proton'
 CT_LAUNCHERS = ['steam', 'heroicproton', 'bottles']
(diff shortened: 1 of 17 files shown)


sonic2kk:
Maybe it would be useful to show this in the status bar as well. The `ghapi_rlcheck` method would need to be updated or potentially a new method could be created to return a boolean to allow for checks on the ratelimit. Maybe returning a list/dictionary with one element being a boolean to represent whether the rate limit has been reached.

```python3
# pupgui2.py
def setup_ui(self):
    # ...
    if ghapi_rlcheck():
        self.ui.statusBar().showMessage(f'{APP_NAME} {APP_VERSION}')
    else:
        self.ui.statusBar().showMessage(f'{APP_NAME} {APP_VERSION} - GitHub Rate Limit Exceeded')
    # ...
```

Could always be done in a separate PR :smile: I also would like to say maybe showing a dialog would be useful, but I think only showing it when a user tries to download a compatibility tool would make the most sense. Users may open PUPQT offline for example on Steam Deck to remove or bulk update Proton versions. So the rate-limit warning dialog could only be shown if a user is online.

Just a small idea I had :-)

PR #167, opened 2022-12-27
fixed spelling issue, add German translation string

Lamdarer:
Hi, 
I noticed a spelling issue when downloading the Proton version(Tita**l**Fall instead of Tita**n**Fall).

I corrected it and while I was on it I also added a German translation and added a `the` to the English version.
I hope this is fine I didnt contribute to ProtonUp-Qt before.

Also Thanks for the cool Program I really like it.


DavidoTek:
Thanks for the contribution, looks good!
Also, great to hear that you like ProtonUp-Qt!

I will merge your changes soon.

PS: I have changed the format of the translation as I recently changed the way they work. I haven't added it to NorthstarProton until now, just so you don't wonder.

Issue #60, opened 2022-04-25
Very long start


VladimirMrzv:
Program start long time, this is reading folder proton from steam. If change to lutris or heroic than start faster. 
Version 2.5.0 and 2.6.0 start faster.

**Desktop (please complete the following information):**  
Ubuntu 22.04
ProtonUp-Qt 2.6.1 AppImage or Flatpak

Testing AppImage and flatpak versions. Old version only Appimage.

DavidoTek:
How many different Proton-GE versions do you have installed? A lot of them may increase the startup time.

Nevertheless, I will try to fix the issue in the next version of ProtonUp-Qt.

VladimirMrzv:
Thank you.
Wait for next release.

VladimirMrzv:
@DavidoTek 
Hope you see this comment.

In some last builds was broken "Show Game List" and at app begun run fast and operate in fast too. I think something wrong happened with read this list at startup.

sonic2kk:
I actually looked into this long start-up time a while ago just before the [`steam` dependency issue](https://github.com/DavidoTek/ProtonUp-Qt/pull/157). I believe the issue is related to the appinfo parser, I think there is some sort of bottleneck. This would also explain why Lutris is faster. Switching to Lutris, closing ProtonUp-Qt and then opening it again will save that Lutris was selected and so it will open faster. Switching to Steam freezes ProtonUp-Qt for around the same amount of time that it takes to open when Steam is selected, meaning it's fetching all of this data again most likely from the VDF file. 

Even on my Steam Deck it takes a long while for ProtonUp-Qt to open. In my case, the `appinfo.vdf` file is almost 25mb, so it's not hard to see why parsing this file could take time. I had a quick look in the file and it seems to have information about every game I own, as it has games I recently bought but that I never downloaded.

In the case of my Steam library, Steam says I have 1,514 games. There are probably a small handful of forgotten Steam shortcuts here but I would say that would only be 2 or 3. I got this figure by going to the library filters beside the search bar and for Steam Deck compatibility, I selected "All Games". This figure includes games that don't count towards the library count on your profile (games that the store lists as "Profile Features Limited"). 

If the `appinfo.vdf` contains information about 1,500+ games it would be no surprise why it takes so long for ProtonUp-Qt to open.

DavidoTek:
Okay, I took a look at the code.

[...]

PR #171, opened 2023-01-16
Optimize start time

DavidoTek:
Close https://github.com/DavidoTek/ProtonUp-Qt/issues/60

A few changes for improving the start time of ProtonUp-Qt.
Mainly some optimizations and avoidance of loops.


sonic2kk:
Improves load times on my PC from ~10 seconds to ~3 seconds. As a result as well, switching from Lutris to Steam is even faster, a slight delay but *much* faster. In the same vein, the games list dialog now also only takes ~3 seconds to load.

Major speedups here on my end :+1:

DavidoTek:
Awesome :tada:

> the games list dialog now also only takes ~3 seconds to load

Great to hear that even the small optimizations made a difference here.

I will see if there is any more performance to squeeze out here, then I will merge it.

PR #174, opened 2023-01-21
Add SteamTinkerLaunch Version in Compat Tool List

sonic2kk:
Some other tools, like Luxtorpeda I believe, show a version string beside their entry on the compatibility tools list. SteamTinkerLaunch does not and I *believe* this was somewhat intentional, as there can only be one version of SteamTinkerLaunch installed at a time.

However for users switching between versions, it may be helpful to see which version ProtonUp-Qt is detecting. It should always detect the currently "active" version that Steam is seeing, and a user can in most cases just run `steamtinkerlaunch --version` from the command line to get the current version. Still, I think it might be at least somewhat useful to show the current version.

stl stable

stl git

We fetch the version from the STL script file by reading the first 20 lines and grabbing the one that has `PROGVERS`. My reason for doing this is that it may be faster than using something like `subprocess.run` to get the version. This is then written out to a `VERSION.txt` file in the `steamdir/compatibilitytools.d/SteamTinkerLaunch` directory. ProtonUp-Qt looks for and reads the `VERSION.txt` for all compatibility tools if its present.

I also fixed a very small issue with one of the dialogs, where a period came after some `\n\n`, putting it on a newline instead of having the newlines start after the period.

<hr>

Just a minor cosmetic change I wanted to make to hopefully help further improve the SteamTinkerLaunch support :-)


DavidoTek:
Awesome!

I think this addition is quite useful, I also thought about it a while ago. Now the user doesn't need to use the terminal.

> reading the first 20 lines and grabbing the one that has `PROGVERS` [...] than using something like subprocess.run

Agreed. Although this isn't implemented in the current code :wink:
`for i, line in enumerate(stl_script):` and `if i == 19:` should work

> I also fixed a very small issue with one of the dialogs, where a period came after some \n\n, putting it on a newline instead of having the newlines start after the period.

Note to myself @DavidoTek : I think we need to update the translations files as well. I will run a simple `replace all` over the source

sonic2kk:
> Agreed. Although this isn't implemented in the current code

Oh my bad you're absolutely right! I wrote most of this yesterday and I guess somewhere along the way I changed how I did this, and it didn't check the first X lines anymore :sweat_smile: 

It should be fixed now! :smile:

DavidoTek:
No worries :smile:

I will do the translations part (probably tomorrow) and then merge it.

Thanks!

Issue #175, opened 2023-01-24
Non-Steam games are not listed in the games list.


marco-calautti:
When clicking the games list button, I can only see steam games, but not the external ones added as shortcut.

1. Open ProtonUp-Qt
2. Select steam (Flatpak in my case)
3. Click on "Show game list"

**Expected behavior**  
All games, both steam and non-steam should be shown

Having also non-steam games listed would allow to automatically bump the proton version for *all* games configured in Steam.

sonic2kk:
ProtonUp-Qt reads the `~/.local/share/Steam/appcache/appinfo.vdf` file to determine which games are owned and installed I believe. However from what I can see from the Non-Steam Games that I added, these entries are not written out to the appinfo file.

I believe they are written elsewhere if memory serve right from working on STL (logically, Steam has to store it somewhere I would guess, how would it remember the given name, compat tool, etc?), but I cannot find or remember where at the moment.

Couple of small notes around this issue that come to mind:
- Not all Non-Steam Games will have a compatibility tool, but not all compatibility tools without a compatibility tool will be native.
- Some tools on Steam Deck like EmuDeck add a lot of Non-Steam shortcuts, which could make the list quite long. Maybe an option to filter Non-Steam games would be helpful.

I'll do a bit of investigation into this and see if I can figure out a way to implement it :-)

sonic2kk:
Information about shortcuts is stored at `~/.local/share/Steam/config/<steam_userid>/config/shortcuts.vdf`. I think the Steam library can parse this but I haven't checked yet.

There's a different problem though. This file, as far as I know, doesn't store information about the compatibility tool used for the file. When updating the compatibility tool used in Steam, this file is not modified.

Another couple of files are modified, including a licensecache file. The more interesting file is the `localconfig.vdf` file. This file is fairly big, and contains lots of misc information about the Steam installation - your friends list, some CDN information, and a list of entries under `Software`.

I scrolled through the file and tried to find somewhere that had some information about the compatibility tool, but even under the `Software` entries I could really only see information about cloud saves and some last played time.

marco-calautti:
From what I know, the compatibility tool mappings are all stored in a vdf inside the config folder.

DavidoTek:
Thanks for the investigation.

> Having also non-steam games listed would allow to automatically bump the proton version for all games configured in Steam.

That an interesting idea which might be worth adding to ProtonUp-Qt.

[...]

PR #176, opened 2023-01-25
Add support to non-steam games

marco-calautti:
This fixes bug #175 . I tested it on my machine and everything seems to be working.
I do not see yet why we would actually need to compute the shortcut ids using the CRC-based algorithm, since Steam does it for us already.

So, please verify I did not miss anything. I batch updated all my games with the new change, and everything worked fine.

### DIFF
--- a/pupgui2/steamutil.py
+++ b/pupgui2/steamutil.py
@@ -89,6 +89,8 @@ def get_steam_app_list(steam_config_folder: str, cached=False) -> List[SteamApp]
                 app.deck_compatibility = 'Unknown'
                 if ct := c.get(str(appid)):
                     app.compat_tool = ct.get('name')
+                    app.ctool_name = ct.get('name')
+                    app.ctool_from_oslist = ct.get('from_oslist')
                 apps.append(app)
 
     except Exception as e:
--- a/pupgui2/steamutil.py
+++ b/pupgui2/steamutil.py
@@ -33,12 +33,16 @@ def get_steam_app_list(steam_config_folder: str, cached=False) -> List[SteamApp]
 
     libraryfolders_vdf_file = os.path.join(os.path.expanduser(steam_config_folder), 'libraryfolders.vdf')
     config_vdf_file = os.path.join(os.path.expanduser(steam_config_folder), 'config.vdf')
+    users_folder = os.path.realpath(os.path.join(os.path.expanduser(steam_config_folder), '../userdata/'))
 
     apps = []
 
+    
     try:
         v = vdf.load(open(libraryfolders_vdf_file))
         c = vdf.load(open(config_vdf_file)).get('InstallConfigStore').get('Software').get('Valve').get('Steam').get('CompatToolMapping')
+        
+        #get steam apps
         for fid in v.get('libraryfolders'):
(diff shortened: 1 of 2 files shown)


DavidoTek:
Thanks! :tada:

> I do not see yet why we would actually need to compute the shortcut ids using the CRC-based algorithm, since Steam does it for us already.

Yes, we do not need to do that. It was only necessary in an older version of Steam.

> So, please verify I did not miss anything. I batch updated all my games with the new change, and everything worked fine.

Did some tests, everything seems to work fine.

We may want to split up the `#get non-steam apps`-part into a separate function (e.g. `get_steam_shortcuts_list`) to keep the code clean and keep it outside the try-catch of the `#get steam apps`-part.
We can call it inside `get_steam_app_list` though. Maybe add another try-catch with `apps = get_steam_shortcust_list(apps)`.

pupgui2/steamutil.py
```diff
@@ -55,6 +59,40 @@ def get_steam_app_list(steam_config_folder: str, cached=False) -> List[SteamApp]
...
+                app.app_id = appid
+                app.shortcut_id = sid
+                app.shortcut_path = svalue.get('StartDir')
+                app.app_type='game'
+                app.game_name = svalue.get('AppName')
+                app.deck_compatibility = 'Unknown'
+                if ct := c.get(str(appid)):
+                    app.compat_tool = ct.get('name')
+                    app.ctool_name = ct.get('name')
+                    app.ctool_from_oslist = ct.get('from_oslist')
```
DavidoTek, line 93: This is only needed if the app is a Steam-own compatibility tool like Proton. It can be removed.

pupgui2/steamutil.py
```diff
@@ -55,6 +59,40 @@ def get_steam_app_list(steam_config_folder: str, cached=False) -> List[SteamApp]
...
+                appid = svalue.get('appid')
+                if appid < 0:
+                    appid = appid +(1 << 32) #convert to unsigned
+                
+                app.app_id = appid
+                app.shortcut_id = sid
+                app.shortcut_path = svalue.get('StartDir')
+                app.app_type='game'
+                app.game_name = svalue.get('AppName')
+                app.deck_compatibility = 'Unknown'
```
DavidoTek, line 89: app.deck_compatibility will default to {} which is interpreted as unknown. It can be removed.

DavidoTek (changes requested):

pupgui2/steamutil.py
marco-calautti: Fixed

pupgui2/steamutil.py
marco-calautti: Fixed

marco-calautti:
> We may want to split up the `#get non-steam apps`-part into a separate function (e.g. `get_steam_shortcuts_list`) to keep the code clean and keep it outside the try-catch of the `#get steam apps`-part. We can call it inside `get_steam_app_list` though. Maybe add another try-catch with `apps = get_steam_shortcust_list(apps)`.

I refactored the code in a separate function.

DavidoTek:
Thanks! Will merge it.

Issue #156, opened 2022-12-11
add support for wine-tkg for lutris (if possible)


DarkReaperBoy:
https://github.com/Frogging-Family/wine-tkg-git

DavidoTek:
I think Wine-Tkg is mainly intended for system wide installation. It doesn't ship with many required libraries (e.g. SDL, ...).
It probably can be made to work, but afaik this isn't the correct way. Let me know if it works fine or there is a way to make it work properly!

Just wondering, is there any specific problem for a game that occurs with Lutris Wine/Wine-GE that doesn't occur with Wine-Tkg?

sonic2kk:
I have used Wine-tkg with Lutris before. I don't recall if I used a specific patch or not, but you can use it with Lutris, which has its own Lutris Runtime as well which may have these libraries? I'm not sure, to be honest.

It would need re-tested and probably input from someone more familiar with using CI builds.

> Just wondering, is there any specific problem for a game that occurs with Lutris Wine/Wine-GE that doesn't occur with Wine-Tkg?

It wouldn't apply here but one of the main benefits of Wine-tkg (and Proton-tkg) is being able to build it with custom Wine patches, which *could* fix issues. Also, I think builds like Lutris Wine are based on a much older Wine version, and GE is based on Valve's Wine maybe? Whereas Wine-tkg is, as far as I know (which may be wrong) based on a much newer version of Wine. Newer Wine builds *may* in some cases have some fixes for a small handful of games, but generally afaik these are patches that were already applied to other Wine builds and have just managed to make their way into mainline Wine.

DavidoTek:
We could modify the [Protok Tkg Ctmod](https://github.com/DavidoTek/ProtonUp-Qt/blob/main/pupgui2/resources/ctmods/ctmod_protontkg.py) to just install Wine-Tkg for Lutris instead of Proton-Tkg for Steam.

The code should work, only `proton` needs to be replaced with `wine`, though I'm not 100% sure about the GitHub CI part/artefact names. The folder structure is quite similar, only the name needs to be changed from `usr` to `wine-tkg-[...]`.

sonic2kk:
Feeling brave, so I'm taking a look at this :eyes: Seems straightforward enough for now. I took a similar approach to STL and STL-git where the Wine-tkg ctmod is a subclass of the Proton-tkg ctmod.

[...]

PR #178, opened 2023-02-02
Add Wine-tkg sources

sonic2kk:
Implements #156.

This PR aims to implement at least one Wine-tkg source into ProtonUp-Qt for use with Lutris :partying_face: If others are desired they should be straightforward to implement too :smile: 

image

As of writing, currently only WIne-tkg based on Valve's Wine has been implemented as a ctmod, based on the so-called "[Exp Bleeding Edge Other distro](https://github.com/Frogging-Family/wine-tkg-git/actions/workflows/wine-valvexbe.yml)". It will download this ctmod and extract the zip, which contains a `.tar.zst` file inside the zip. This extracts to a `usr` folder, which is the actual folder that Lutris will use as its base Wine runner directory, so we rename it to match the name of the zip archive.

On a technical note, this PR implements the Wine-tkg ctmod as a separate ctmod file, but this is just a subclass of the Proton-tkg ctmod as the logic is 95%. The only three places the logic differs is:
- The `PROTON_PACKAGE_NAME`, which in the case of the Valve Wine build is `wine-valvexbe`.
- The `install_folder` should be suffixed with `wine_tkg` instead of `proton_tkg,` so we allow setting a custom suffix using a class variable `TKG_EXTRACT_NAME`.
- Wine-tkg builds hold a `.tar.zst`, so the Proton-tkg `get_tool` method now has a check for a `*tar.zst` archive and then handles the extraction and renaming for that process.

Hopefully this approach to implementation is alright :sweat_smile: I thought it would be way overkill to just copy/paste the Proton-tkg ctmod, especially if we want to implement more Tkg builds! I took a similar approach with SteamTinkerLaunch and SteamTinkerLaunch-git since they both downloaded from the same repo, and I think in this case since the Wine-tkg and Proton-tkg builds share a repo, it makes sense to have subclassed ctmods in this instance as well :-)

<hr>

I guess if we can agree on and finalize the implementation of this Wine-tkg build, we can start to discuss others. To reiterate what I left in a comment on #156, I think we should give the user at least two options for the Wine-tkg build they want by default: One based on Valve's Wine, and another based on vanilla Wine.

I am open to suggestions on which flavour of vanilla Wine to go with, but I figured that the one based on the Ubuntu buildsystem is probably the safest to go with. I have no preference either way :slightly_smiling_face: 

I was also thinking that we could add the League of Legends Wine build under Advanced Mode, since we have the Titanfall 2 build perhaps we could also support the LoL Wine as well. I don't play League of Legends so I can't say how beneficial it may be, and unlike the Titanfall 2 build it was not requested (as far as I know :nerd_face:), it it may be "solving" a problem that doesn't exist :sweat_smile:

<hr>

Thanks as always! :smile:


DavidoTek:
Great!

> this PR implements the Wine-tkg ctmod as a separate ctmod file, but this is just a subclass of the Proton-tkg

Very good, makes sense.

> I think we should give the user at least two options for the Wine-tkg build they want by default: One based on Valve's Wine, and another based on vanilla Wine.

We can do that. Maybe we should hide one of them behind *advanced mode* so it doesn't get too much. I'm open on this.

> but I figured that the one based on the Ubuntu buildsystem is probably the safest to go with

Yeah, Ubuntu uses older/more stable dependencies. Should work on most other distros then.

> I was also thinking that we could add the League of Legends Wine build under Advanced Mode [...] I can't say how beneficial it may be [...]

I also can't say what the advantages of this build are, but nothing speaks against adding it.

---

I will do an in-depth review tomorrow and merge.

Thanks! :tada: :tada:

sonic2kk:
D'oh, just realised this extracts the dotfiles that come with Wine-tkg:
- `.BUILDINFO`
- `.INSTALL`
- `.MTREE`
- `.PKGINFO`

We could put these into a list, and check for and remove them after successful extraction maybe. Will look into it later :-)

DavidoTek:
> Perhaps the vanilla Wine should be kept to advanced mode?

That was my thought too.

> If we go with the Ubuntu one, it might look something like this: [code]
> [...] checked a vanilla Wine-tkg build and it is actually packaged slightly different

That is short. Very nice.
Maybe we can add a check for the `usr` folder in the "main-ctmod" protontkg later if necessary.

> I think that the LoL builds have various patches that are incompatible

All right.

sonic2kk:
Added a small check to remove the listed files. I prefer removing from a list instead of removing dotfiles so we are explicit about what we remove, but if it's overkill we can change it. Also, not sure if I implemented the removal in the most efficient way, but it's a start at least :-)

pupgui2/resources/ctmods/ctmod_protontkg.py
```diff
@@ -189,13 +191,47 @@ def get_tool(self, version, install_dir, temp_dir):
...
-                f_count += 1
-                tarfile.open(f, "r").extractall(install_dir)
-            if f_count > 0:
-                shutil.rmtree(install_folder)
+            
+            # Supports both Wine-tkg and Proton-tkg
+            zst_glob = glob.glob(f'{install_folder}/*.tar.zst')
+            if len(zst_glob) > 0:
+                # Wine-tkg is .tar.zst
+                tkg_dir = os.path.abspath(os.path.join(install_dir, '../../runners/wine'))
```
DavidoTek, line 199: We should remember that this line is for **Lutris** compatibility tools. Hence this comment.

DavidoTek:
> I prefer removing from a list instead of removing dotfiles so we are explicit about what we remove,

Good, that keeps it clean.

---

Thanks, will merge.

PR #179, opened 2023-02-03
Fix Steam AppID sort in ctinfodialog

sonic2kk:
In the `ctinfodialog`, sorting by Steam AppID doesn't work properly. I guess it tries to sort them as strings instead of numbers. The fix for this is a small change to set the data properly as an integer. In my tests, a call to `setData` *is* actually required, as just `QTableWidgetItem(int(game.get_app_id_str()))` results in an empty AppID column :shrug: 

image

My bad for missing this initially. No change should be needed on the Lutris side as Lutris is sorted by slugs, which are strings anyway. I think Steam even stores AppIDs as integers so there shouldn't be any problem with setting the data type as integer.

Thanks :smile:


DavidoTek:
Thanks very much for also fixing those little issues! :tada:

Never tried sorting by AppID (I only have a few games installed...), but great it is fixed now :smile:

PR #181, opened 2023-02-04
Lutris: Add Vanilla Wine

sonic2kk:
Follow-up from discussion in #178, another component of #156 

Adds the ability to download Vanilla Wine as a runner for Lutris. It uses the build based on the [Ubuntu build system](https://github.com/Frogging-Family/wine-tkg-git/actions/workflows/wine-ubuntu.yml), as discussed in #178. This is gated behind advanced mode as builds of Wine based on vanilla Wine may be less compatible, and generally a user will *probably* want to use builds based on Valve Wine for gaming, likewise discussed in #178.

image

Not sure if we want anything further added to the description of the ctmod. I did consider adding a warning but that might be excessive, and a user in advanced mode should (ideally) know what they're doing :slightly_smiling_face: 

The process of creating the ctmod is actually exactly how I described it in the previous PR, it was able to download and extract perfectly fine without any changes to the base Tkg ctmod! :partying_face: 

If desirable, as part of this PR I could also look into adding the League of Legends build, gated behind advanced mode as well. Not sure how beneficial or useful this is though.

I did some tests with a couple of builds and it seemed to work fine, but if I missed anything please feel free to let me know :-) Thanks!

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_winetkg_vanilla_wine.py
+++ b/pupgui2/resources/ctmods/ctmod_winetkg_vanilla_wine.py
@@ -1,0 +1,25 @@
+# pupgui2 compatibility tools module
+# Proton-Tkg https://github.com/Frogging-Family/wine-tkg-git
+# Copyright (C) 2022 DavidoTek, partially based on AUNaseef's protonup
+
+from PySide6.QtCore import QCoreApplication, Signal
+
+from pupgui2.resources.ctmods.ctmod_protontkg import CtInstaller as TKGCtInstaller  # Use ProtonTKg Ctmod as base
+
+
+CT_NAME = 'Wine Tkg (Vanilla Wine)'
+CT_LAUNCHERS = ['lutris', 'advmode']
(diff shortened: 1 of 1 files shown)


DavidoTek:
Thanks :tada:

> Not sure if we want anything further added to the description of the ctmod

The title `Wine Tkg (Vanilla Wine)` probably will be enough for "advanced users". Do you think we should add a link to the GitHub page as well?

> The process of creating the ctmod is actually exactly how I described it in the previous PR, it was able to download and extract perfectly fine without any changes to the base Tkg ctmod! 🥳

That's great and will keep maintainance low!

> If desirable, as part of this PR I could also look into adding the League of Legends build, gated behind advanced mode as well. Not sure how beneficial or useful this is though.

It won't hurt anything, but I can't say how useful the build is to LoL players.

sonic2kk:
> Do you think we should add a link to the GitHub page as well?

The "Info" button should direct to the [Wine Ubuntu CI](https://github.com/Frogging-Family/wine-tkg-git/actions/runs/4095935362) for the current build, though we could always add a link to the repo more explicitly in the description (and we could mirror this for the Valve Wine ctmod).

> It won't hurt anything, but I can't say how useful the build is to LoL players.

Yeah, maybe it can wait until if/when it's requested :-)

PR #182, opened 2023-02-09
Game List Improvements

sonic2kk:
This PR fixes a couple of bugs introduced in #180, and adds a small feature that I figured out how to do in the process of fixing those bugs :smile: 

No screenshot this time as there isn't really much to "show" :sweat_smile: 

- **Critically**, fixes the ProtonDB rating text having the wrong flag, meaning it didn't display (needs `Qt.DispalyRole` instead of `Qt.UserRole`)
- Fixes items in the games list being able to be double-clicked and edited by setting `NoEditTriggers` in Qt Designer - I don't think this would cause any problems, but it's minor unwanted behaviour imo
- Adds double-click actions to cells in the Game List, similar to the "Show info" dialog
    - For Game Name, opens the Steam store page - Lutris does not have the "Show games list" button
    - For Anti-Cheat, open the Are We Anti-Cheat Yet web page - It doesn't appear that there is a way to open on a specific game yet
    - For ProtonDB, once a rating is revealed, open the given ProtonDB app page
- Adds a tooltip to the Game Name with `Game Name (AppID)`

The double-click actions were implemented by appending the relevant URLs to a `Qt.UserData` data entry for each `QTableWidgetItem`. Initially I thought of having type checks in the `item_doubleclick_action` method, but this got a bit unwieldy, and it was going to become messy to build things like the ProtonDB URL that way. So instead, we build pass the relevant URLs as data when we create the `QTableWidgetItem`.

The main fix here is the ProtonDB rating breaking, apologies for missing that initially.

Thanks! :-)

### DIFF
--- a/pupgui2/constants.py
+++ b/pupgui2/constants.py
@@ -65,8 +65,10 @@ PROTONDB_COLORS = {'platinum': '#b4c7dc', 'gold': '#4f492c', 'silver': '#a6a6a6'
 STEAM_API_GETAPPLIST_URL = 'https://api.steampowered.com/ISteamApps/GetAppList/v2/'
 STEAM_APP_PAGE_URL = 'https://store.steampowered.com/app/'
 AWACY_GAME_LIST_URL = 'https://raw.githubusercontent.com/Starz0r/AreWeAntiCheatYet/master/games.json'
+AWACY_WEB_URL = 'https://areweanticheatyet.com/'
 LOCAL_AWACY_GAME_LIST = os.path.join(TEMP_DIR, 'awacy_games.json')
 PROTONDB_API_URL = 'https://www.protondb.com/api/v1/reports/summaries/{game_id}.json'
+PROTONDB_APP_PAGE_URL = 'https://protondb.com/app/'
 
 STEAM_BOXTRON_FLATPAK_APPSTREAM = 'appstream://com.valvesoftware.Steam.CompatibilityTool.Boxtron'
 STEAM_PROTONGE_FLATPAK_APPSTREAM = 'appstream://com.valvesoftware.Steam.CompatibilityTool.Proton-GE'
--- a/pupgui2/pupgui2gamelistdialog.py
+++ b/pupgui2/pupgui2gamelistdialog.py
@@ -6,13 +6,13 @@ from PySide6.QtGui import QPixmap, QBrush, QColor
 from PySide6.QtWidgets import QLabel, QComboBox, QPushButton, QTableWidgetItem
 from PySide6.QtUiTools import QUiLoader
 
-from pupgui2.constants import PROTONDB_COLORS
+from pupgui2.constants import PROTONDB_COLORS, STEAM_APP_PAGE_URL, AWACY_WEB_URL, PROTONDB_APP_PAGE_URL
 from pupgui2.datastructures import AWACYStatus, SteamApp, SteamDeckCompatEnum
 from pupgui2.lutrisutil import get_lutris_game_list
 from pupgui2.steamutil import steam_update_ctools, get_steam_game_list
 from pupgui2.steamutil import is_steam_running, get_steam_ctool_list
 from pupgui2.steamutil import get_protondb_status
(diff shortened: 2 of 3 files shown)


sonic2kk:
Hang on, the first item in the games list seems to never show the ProtonDB rating. The click event doesn't seem to do anything. I will investigate.

DavidoTek:
> Critically, fixes the ProtonDB rating text having the wrong flag, meaning it didn't display

Actually I haven't noticed that... :smile: 

> Adds double-click actions to cells in the Game List, similar to the "Show info" dialog 

That's useful, great!

> When the index is 0, I guess Python sees this as falsy and doesn't run the block?
> Not sure what the best solution is here yet. Will think about it some more :-)

Yes. I don't think the `:=` makes a difference and Python just reads it as `if 0: ...`
Simplest solution if can think of is following (maybe there is a more Pythonic way): 
```python3
i = self.games.index(game)
if i != None:
    print('...')
```

> Makes the Games List dialog wider, from 650px to 800px
> make the Games List column 300px to allow more room for game titles
> In my humble opinion it looks nicer, and I really wanted to fix that scrollbar issue :sweat_smile:

Looks much cleaner now without the scrollbar.
I'm wondering, does the width impact use on Steam Deck? I don't think it uses Scaling on the desktop, right? Then 800px (<1280px) will be just fine.

---

Thanks!

sonic2kk:
>  Simplest solution if can think of is following (maybe there is a more Pythonic way):

Hmm, I implemented it and it works but I think there is a potential problem here. We already do the `None` check above, with `if not game`, but I guess there could be an instance where `game` is not `None`, but also is not in the `self.games` list. This will probably never happen I would assume but still.

Though having said that, I am not sure the `i := self.games.index` would help in this instance, either

```python
games = [ 'Cookie Clicker', 'PowerWash Simulator', 'Sonic Mania', 'Fallout: New Vegas' ]

# Should print an error
if i := games.index('STEINS;GATE'):
    print('What?')
```

Given that, the following code might be a solution:

```python
if i := self.games.index(game) != None:
```

But in this case, I am not sure how this differs from just:

```python
if self.games.index(game) != None:
```

Since `i` seems to go completely unused.

This could just be my inexperience showing, so please correct me :smile: 

> I'm wondering, does the width impact use on Steam Deck? I don't think it uses Scaling on the desktop, right? 

Good catch! There is indeed no scaling by default on Steam Deck. A user *could* enable it... but they could also enable a lot of things, so we may not have very much to worry about :sweat_smile: It should just "catch" on the corners of the screen even with scaling.

I would test this myself, but unfortunately I am not sure how to test on Steam Deck anymore as of SteamOS 3.4, the virtualenv stuff I was using before doesn't seem to work. It definitely used to, because I tested a bunch from source for the STL work in the past.

DavidoTek:
> This will probably never happen I would assume but still.

As we are relying on external data from other apps (Steam client), let's better be safe than sorry :smile:

> `if i := self.games.index(game) != None:`

I think what this does is:
```python3
i = (self.games.index(game) != None)
if i:
    ...
```

> Since i seems to go completely unused.

Seems to be a leftover from an older revision of the code: `self.ui.tableGames.setCellWidget(i, 4, lbl_protondb_compat)`

> There is indeed no scaling by default on Steam Deck. A user could enable it... but they could also enable a lot of things, so we may not have very much to worry about sweat_smile

All right, then it shouldn't be a problem. And even with 1.5x scaling it will be less than 1280px.

sonic2kk:
> As we are relying on external data from other apps (Steam client), let's better be safe than sorry :smile: 

I agree, it's easy to be burned by black boxes like that.

> I think what this does is:

Yup, looks to be correct. We can probably just do `self.games.index(game) != None` :smile: 

> Seems to be a leftover from an older revision of the code

Ahh I see, I thought perhaps it might've been but wasn't totally sure.

<hr>

Pushed a commit that implements the change to the `self.games.index` check. I did a quick test after changing and nothing seems to have broken :+1:

PR #185, opened 2023-02-10
feat: show message box for CTs failed to load

### LINKED ISSUE #184: Show a message box when a compatibility tool failed to load
Without reading the log, I didn't notice that there were some CTs failed to load.

Show a message box about CTs failed to load, and optionally the reason too

Read the logs, but since ProtonUp-Qt is designed for user-friendly experience I don't think we need to read the log for that thing.

teppyboy:
Fixes #184

### DIFF
--- a/pupgui2/util.py
+++ b/pupgui2/util.py
@@ -19,6 +19,54 @@ from pupgui2.datastructures import BasicCompatTool, CTType
 from pupgui2.steamutil import remove_steamtinkerlaunch
 
 
+def create_msgbox(
+    text: str,
+    info_text: str = None,
+    buttons: QMessageBox.StandardButton | tuple[QMessageBox.StandardButton] = QMessageBox.Ok,
+    default_button: QMessageBox.StandardButton = QMessageBox.Ok,
+    detailed_text: str = None,
+    icon: QMessageBox.Icon = QMessageBox.Information,
+    execute: bool = True,
+) -> int | QMessageBox:
+    """
+    Show a new message box
+    Args:
+        text: The text to show.
+        info_text: The informative text to show.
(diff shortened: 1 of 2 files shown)


pupgui2/util.py
```diff
@@ -19,6 +19,54 @@
...
 
+def create_msgbox(
+    text: str,
+    info_text: str = None,
+    buttons: QMessageBox.StandardButton | tuple[QMessageBox.StandardButton] = QMessageBox.Ok,
+    default_button: QMessageBox.StandardButton = QMessageBox.Ok,
+    detailed_text: str = None,
+    icon: QMessageBox.Icon = QMessageBox.Information,
+    execute: bool = True,
+) -> int | QMessageBox:
```
sonic2kk, line 30: I think this return type syntax is only available in Python 3.10, and ProtonUp-Qt may be limited to Python 3.8 (47d740c65a97c87acf3535a328a62cb02857a425).

I think in older versions of Python from 3.5 and onward (when type hinting as introduced iirc), we have to use `Union`:
```python3
from typing import Union
# ...
def create_msgbox(...) -> Union[int, QMessageBox]:
    # ...

sonic2kk:
Left a couple of comments about some functionality potentially being unavailable in Python 3.8 and tried to offer an alternative (even if it's not as pretty, sorry about that).

I had skim over the rest and didn't find anything else that might not work, but I also don't have Python 3.8 on my system to check :sweat_smile:

teppyboy:
Yeah, I forgot about Python 3.8 compatibility. Thank you for reviewing my PR :p

sonic2kk (approved):
Not sure if I can mark the comments/review as such now but my comments were addressed with the latest commits :+1:

PR #187, opened 2023-02-11
Add support for Proton-tkg (Wine Master)

sonic2kk:
Requested in a comment in https://github.com/DavidoTek/ProtonUp-Qt/issues/156#issuecomment-1426677052.

Adds support for Proton-tkg built against Wine Master when running in Advanced Mode. As far as I'm aware there is generally very little reason to want to use a build of Proton based on Wine Master, but there are *some* instances where it may have some fixes. The most recent that comes to mind is that supposedly builds of Wine 8.1 and newer fix issues with DOOM 3 BFG. Likely there were others that I am just unaware of too, but that's my justification for implementing this request :sweat_smile: 

I considered renaming the regular Proton-tkg tool we have but specifying may just be confusing for users. I'm not sure though, we could always mirror what we did for Wine-tkg and name it "Proton Tkg (Valve Wine)". I have no strong preference on this :-) 

image

I opted to go for the Arch Linux CI here instead of the Ubuntu one for two reasons:
1. Proton here seems to be built against the Arch CI anyway.
2. On the wine-tkg-git Readme, they link to the Arch CI for the Proton Wine Master builds.

This was implemented identically to how #178, #181 were implemented; by subclassing the Proton-tkg ctmod. Only a couple of variables needed to be changed here, the name, launchers, description, and `PROTON_PACKAGE_NAME` (so it knows which build system to fetch from).

The archive structure here for this Proton-tkg Wine Master is identical to regular Proton-tkg. It's still a zip file with a tar inside, named as `proton_tkg_<versioning info>`. I did notice that it took a considerably long time to extract. Maybe 30 seconds? I actually thought it wasn't going to succeed, because it stuck at 98%. I kept an eye on the `compatibilitytools.d` folder and eventually the extracted folder appeared and right away the Proton-tkg version appeared in ProtonUp-Qt :partying_face: 

Thanks!

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_protontkg_winemaster.py
+++ b/pupgui2/resources/ctmods/ctmod_protontkg_winemaster.py
@@ -1,0 +1,27 @@
+# pupgui2 compatibility tools module
+# Proton-Tkg https://github.com/Frogging-Family/wine-tkg-git
+# Copyright (C) 2022 DavidoTek, partially based on AUNaseef's protonup
+
+from PySide6.QtCore import QCoreApplication, Signal
+
+from pupgui2.resources.ctmods.ctmod_protontkg import CtInstaller as TKGCtInstaller  # Use ProtonTKg Ctmod as base
+
+
+CT_NAME = 'Proton Tkg (Wine Master)'
+CT_LAUNCHERS = ['steam', 'advmode']
+CT_DESCRIPTION = {'en': QCoreApplication.instance().translate('ctmod_protontkg_winemaster', '''Custom Proton build for running Windows games, built with the Wine-tkg build system.
+<br/>
+<br/>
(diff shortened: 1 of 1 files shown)


DavidoTek:
Thanks!

> I considered renaming the regular Proton-tkg tool we have but specifying may just be confusing for users. I'm not sure though, we could always mirror what we did for Wine-tkg and name it "Proton Tkg (Valve Wine)". I have no strong preference on this :-)

We could add `This build is based on Valve Wine (recommended)` to the description like you did with this one.

> I did notice that it took a considerably long time to extract. Maybe 30 seconds?

It's only a few MB larger than Value Wine but the zip size for both is about 1GB wow!
I think(hope) most users will have the patience to wait 30 seconds before reporting it as a bug :)

Issue #162, opened 2022-12-19
[Feature Request] Use a symlink to help auto update wine-ge


Etaash-mathamsetty:
essentially this is how the feature works, so whenever the user opens proton-up-qt and downloads a newer version of a runner, we will symlink this newer version to wine-ge-latest (or something similar), then all the user has to do to keep wine-ge updated is just download a newer version of wine-ge and heroic, and lutris which have their wine/proton paths pointing to the wime-ge-latest folder. 

steps to do it manually:  
1. you will have to move the proton-ge folder outside of the compatibility tools folder 
2. symlink it into the comatibility tools folder with name wine-ge-latest (use this name every time you perform the replacement)

an alternative method would be to rename the folders, but with that approach I can imagine things getting very messy, very fast

sonic2kk:
Possibly related / implementation detail for #117

DavidoTek:
That should be doable.

How would a user choose to use "wine-ge-stable" symlink option? I think we have 3 options:
1. Automatically create the symlink for every new version (no user option)
2. Add a checkbox like `Install option "Latest Wine-GE" to the launcher to automatically select the newest version` to the install dialog or the settings
3. Show the option `Latest Wine-GE` in the version drop down of the install dialog  
   (Not sure if that is confusing)

Internally, it should be quite easy to do:
1. Delete the old symlink if it exists
2. Create a symlink to the newly installed version folder but replace the version in the name with `latest`

Etaash-mathamsetty:
> That should be doable.
> 
> How would a user choose to use "wine-ge-stable" symlink option? I think we have 3 options:
> 1. Automatically create the symlink for every new version (no user option)
> 2. Add a checkbox like `Install option "Latest Wine-GE" to the launcher to automatically select the newest version` to the install dialog or the settings
> 3. Show the option `Latest Wine-GE` in the version drop down of the install dialog  
>    (Not sure if that is confusing)
> 
> Internally, it should be quite easy to do:
> 1. Delete the old symlink if it exists
> 2. Create a symlink to the newly installed version folder but replace the version in the name with `latest`
> 

Personally I like having choice and so I think option 2 is the best

Etaash-mathamsetty:
Any updates on this?

sonic2kk:
Been fiddling with this, if I make some more progress I will hopefully get a PR up soon for at least Wine-GE. I'm trying to keep in mind potentially doing this for other compatibility tools, as well as "cleaning up".

[...]

Issue #139, opened 2022-11-18
DXVK Updater


massatt212:
Can you add a feature to update Custom proton DXVK files

idk if its hard to do, but copy pasting dxvk everytime is becoming painfully tiering, i do a lot of testing so ill love a way and i sure others will also love that feature.

sonic2kk:
Related: #117

sonic2kk:
Actually, DXVK doesn't even get "installed" by ProtonUp-Qt. It's just downloaded into the Lutris runtime folder, and is selected from there. I didn't know this at the time of writing my above reply (shame on you, past sonic2kk)

I guess this is basically the same as #117 then?

sonic2kk:
Small note for this: DXVK has recently removed the (imo very useful) `setup_dxvk` script from git: https://github.com/doitsujin/dxvk/commit/64cb1ad208f80b11400eef779e99fec223754098

PR #192, opened 2023-02-18
Lutris: Add Basic Games List

sonic2kk:
This PR adds a Games List for Lutris, adjusting the data displayed to be relevant to Lutris games :smile: It displays games from all runners except Steam (as we already display Steam games in the Steam games list dialog), and also has full sorting support to match the Steam games list. Because I really wanted a change of pace from writing Bash :sweat_smile: 

I implemented this in `pupgui2gameslistdialog.py`. There was some logic there already in preparation for a Lutris games list it seems. I'm a bit undecided if this is "better" than having a separate dialog, but the ctinfo dialog is shared, and as mentioned there was already some prep work for a Lutris games list, so I figured it was fine at least for now :-)

image

## Overview
Instead of displaying the compatibility tool, Deck compatibility, Anti-Cheat status, and ProtonDB rating, we instead display the following:
- Game Name
  - Tooltip: `Game Name (Game Slug)`
  - Double click: Takes you to lutris.net page for that game, if an `installer_slug` is present (only games with an `installer_slug` have a page on lutris.net, games without are manually-added games)
- Runner (with first letter capitalized as a small design preference)
  - Tooltip: Wine version, DXVK version, and vkd3d version; whichever are available
- Install Location
  - Tooltip: Either `Double click to browse`, or a warning if the directory doesn't exist (Lutris has a separate section on its menu for games with invalid paths as well)
  - Double click: Opens that path with `xdg-open`, if it exists. Otherwise it does nothing.
- Installed Date (ISO formatted)
  - Tooltip: ISO formatted date with time in brackets, e.g. `2010-07-28 (12:51:03)`

The fifth column that the Steam games list has is hidden. The "Apply" button was also changed to "Close", since there is nothing to Apply in this menu.

I chose to put the Wine information in a tooltip because I figured it would be useful to show a "complete" Lutris games list, minus Steam games only because we already display that for the Steam runner. Since not all Lutris runners will have Wine information, I chose to put it in the tooltip when the runner is Wine.

There is the option of only showing Lutris games that have a Wine runner, and then we show the Wine version/DXVK version/vkd3d version in that column (in that fallback order, in case one is missing, and show the rest in the tooltip). In the case where there is no Wine version (yep, that can be a scenario...) we could display "Unknown". I am personally slightly more in favour of the approach I took but only because my personal preference is seeing a "full" Lutris game list.

Unlike the Steam dialog though, there is no way to "change" a compatibility tool, there is no dropdown in the Lutris list like its Steam counterpart, and there are a couple of reasons for this:
- Related to above, not all runners have a Wine version to change
- A game with a Wine runner can have its Wine, DXVK and vkd3d version changed independently. These *could* all be added to the dialog in separate columns, and there would have to be logic to skip/remove unset Wine/DXVK/vkd3d versions so that it uses the Lutris default, but users may inadvertently think they need to set the DXVK/vkd3d versions for games. This could potentially lead to confusion.
- Bulk-changing Lutris Wine versions may not be as desirable as it is for Steam games, though updating to the latest, for example, Wine-GE may be desirable in some cases- But instead of this being implemented in this Games List, it could be better left to the likes of #162 or #139.

## Other Changes
I made a couple of other changes along the way to implement this.
- Updated logic in `LutrisGame#get_game_config` to use `os.path.isfile` instead of `os.path.exists`. If `fn == ''`, then the result of `os.path.join(os.path.expanduser(lutris_config_dir), 'games', fn)` would point to `{lutris_config_dir}/games/` (because `fn` was blank, the end of the path was just not filled in). This config dir path should pretty much always exist if Lutris is installed.
- In `lutrisutil.get_lutris_game_list`, some extra logic was added to fetch the install path of a game and store it in `install_dir`. We don't bother checking if this directory actually exists as its still a valid Lutris entry regardless, and Lutris itself has a separate menu entry for games that are missing.
  - The Lutris query was updated to select `directory`, however this will only be set whn a game is installed via a Lutris installer, as during the install process it asks you for an install directory. Manually added games do not have this.
  - As a fallback if `directory` cannot be found, we check if the game has a Working Directory, which we can get from the Lutris game's config yml. This is an optional field as far as I know and Lutris will infer it probably from the exe base directory, but not all games have to have a working directory as far as I am aware (at least one of my games didn't :sweat_smile:).
  - As a final fallback, if we don't have `directory` (i.e. this is a manually added game) **and** if we don't have a working directory, we use the base directory of the exe as the install directory. This may not always be correct, imagine a scenario where an exe is buried in some folder structure like `/path/to/GameFolder/CompanyName/EngineName/Game.exe`. However as far as I know, this exe path can't ever really be blank, so it serves as a good final fallback in my opinion.
  - If we don't have a `directory` from the query, or a working directory or exe directroy, we default to an empty string.
- Remove the vertical header from the games list, as it doesn't appear to be used anywhere. I tested the Steam dialog as well and I was able to update compatibility tools, sort, etc without issue. If it is actually used I can re-enable it no problem though.

I have only tested this on my main PC, which only has a selection of Lutris games. Further testing is, of course, very welcome here to make sure this works for more than my setup :-)

<hr>

I decided to try and implement this on a whim, after I noticed the commented out lines to show the games list button for Lutris games. So I wanted to see if it would be feasible to implement a Lutris Games List. I don't use Lutris very much so it was just a bit of an experiment that could potentially end up being useful :-)

At the very least, I hope that this could serve as a "proof-of-concept" for a future Lutris games list implementation (perhaps similar to what I described with the Wine/etc comboboxes), if this simplistic "game information only" approach is not desirable.

Thanks!

### DIFF
--- a/pupgui2/pupgui2gamelistdialog.py
+++ b/pupgui2/pupgui2gamelistdialog.py
@@ -1,14 +1,17 @@
 import os
 import pkgutil
 
+from typing import List, Callable
+from datetime import datetime
+
 from PySide6.QtCore import QObject, Signal, Slot, QDataStream, QByteArray, Qt
 from PySide6.QtGui import QPixmap, QBrush, QColor
 from PySide6.QtWidgets import QLabel, QComboBox, QPushButton, QTableWidgetItem
 from PySide6.QtUiTools import QUiLoader
 
-from pupgui2.constants import PROTONDB_COLORS, STEAM_APP_PAGE_URL, AWACY_WEB_URL, PROTONDB_APP_PAGE_URL
+from pupgui2.constants import PROTONDB_COLORS, STEAM_APP_PAGE_URL, AWACY_WEB_URL, PROTONDB_APP_PAGE_URL, LUTRIS_WEB_URL
 from pupgui2.datastructures import AWACYStatus, SteamApp, SteamDeckCompatEnum
-from pupgui2.lutrisutil import get_lutris_game_list
+from pupgui2.lutrisutil import get_lutris_game_list, LutrisGame
 from pupgui2.steamutil import steam_update_ctools, get_steam_game_list
 from pupgui2.steamutil import is_steam_running, get_steam_ctool_list
 from pupgui2.steamutil import get_protondb_status
@@ -25,10 +28,10 @@ class PupguiGameListDialog(QObject):
         super(PupguiGameListDialog, self).__init__(parent)
         self.install_dir = install_dir
         self.parent = parent
-
         self.queued_changes = {}
 
         self.install_loc = get_install_location_from_directory_name(install_dir)
+        self.launcher = self.install_loc.get('launcher')
 
         self.load_ui()
         self.setup_ui()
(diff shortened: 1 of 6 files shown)


DavidoTek:
> I implemented this in pupgui2gameslistdialog.py. There was some logic there already in preparation for a Lutris games list it seems. I'm a bit undecided if this is "better" than having a separate dialog

Okay. If there are some fundamental changes to the Lutris game list in the future, we can always separate it from the Steam game list.

> we instead display the following: Game Name, Runner, Install Location, Install Date

Looks good to me

> Since not all Lutris runners will have Wine information, I chose to put it in the tooltip when the runner is Wine.

Yeah, I think that is fine. If we ever decide to add functionality for changing the Wine version, we may utilize the fifth column for that.
But as you pointed out, users may also want to change DXVK/vkd3d versions. That can be done easily in Lutris own settings.

> seeing a "full" Lutris game list

Agreed

> [Feature Request] Use a symlink to help auto update wine-ge #162 or DXVK Updater #139.

Agreed

> In lutrisutil.get_lutris_game_list, some extra logic was added to fetch the install path of a game and store it in install_dir

That's good. Even if the path may not be exactly correct, it should be good for users to get a rough idea where the game is stored and that's enough.

> Remove the vertical header from the games list, as it doesn't appear to be used anywhere

It was originally used for showing game ids, but that feature was eventually removed (can't remember when and why).

---

Thanks for the effort to implement this. I think it is much better than the current situation: no game list :smile:.
Adding this "game information only" list may encurrage users to think about more usecases for the dialog. Also, the "always non-empty install directory feature" is quite nice to have :smile:
I will do some testing and merge it in the next days.

pupgui2/pupgui2gamelistdialog.py
```diff
@@ -167,16 +185,68 @@ def update_game_list_steam(self):
...
+            # Some games may be in Lutris but not have a valid install path, though the yml should *usually* have some path
+            install_dir_text = game.install_dir or 'Unknown'
+            install_dir_item = QTableWidgetItem(install_dir_text)
+            if install_dir_text == 'Unknown':
+                install_dir_item.setForeground(QBrush(QColor(PROTONDB_COLORS.get('gold'))))
+            else:
+                if os.path.isdir(install_dir_text):
+                    # Set double click action to open valid install dir with xdg-open
+                    install_dir_item.setToolTip('Double-click to browse...')
+                    install_dir_item.setData(Qt.UserRole, lambda url: os.system(f'xdg-open "{url}"'))
```
DavidoTek, line 233: I don't think that will work with the Flatpak version of ProtonUp-Qt. If inside Flatpak, it requires the prefix `flatpak-spawn --host`:
`flatpak-spawn --host xdg-open /path/here/`

Maybe we should add a function like [`host_which`](https://github.com/DavidoTek/ProtonUp-Qt/blob/cb9b7168157743688cc5a8bc6a463e8d73b7976f/pupgui2/util.py#L384-L391) that runs commands on outside the sandbox.

pupgui2/pupgui2gamelistdialog.py
```diff
@@ -167,16 +185,68 @@ def update_game_list_steam(self):
...
 
-        for i, game in enumerate(games):
-            self.ui.tableGames.setItem(i, 0, QTableWidgetItem(game.name))
+        # Not sure if we can allow compat tool updating from here, as Lutris allows configuring more than just Wine version
+        # It lets you set Wine/DXVK/vkd3d/etc independently, so for now the dialog just displays game information
+        for i, game in enumerate(games): 
+            name_item = QTableWidgetItem(game.name)
+            name_item.setToolTip(f'{game.name} ({game.slug})')
+            if game.installer_slug:
+                # Only games with an installer_slug will have a Lutris web URL - Could be an edge case that runners get removed/updated from lutris.net?
```
DavidoTek, line 203: That's fine I guess. That shouldn't occur often.

pupgui2/pupgui2gamelistdialog.py
sonic2kk: I was not sure if it would work, I lifted this logic from [`Line 384-386 in pupgui2.py`](https://github.com/DavidoTek/ProtonUp-Qt/blob/main/pupgui2/pupgui2.py#L385).

If this won't work in Flatpak, we could see about implementing something like `host_xdgopen` or something along these lines specifically for this action :-)

There is also the possibility that this won't work outside of Flatpak, if `xdg-utils` is not installed. But I think *most* systems, especially those for gaming (including Steam Deck) will come with `xdg-utils`.

sonic2kk:
> users may also want to change DXVK/vkd3d versions. That can be done easily in Lutris own settings.

I think this is a good compromise, and using the fifth column for the Wine version is a good idea. The column is just hidden, so if we ever need to use it, it's just a case of removing `self.ui.tableGames.setColumnHidden(4, True)`.

Or another possibility is removing the install time data and replacing it with the Wine version. Install time information could be moved somewhere else, maybe the tooltip for the game name or install path.

> Adding this "game information only" list may encurrage users to think about more usecases for the dialog

Great point! I'm not a *huge* Lutris user so I'm sure Lutris users will have good ideas on improving it :-)

Issue #168, opened 2022-12-30
Add game list for Heroic Games Launcher


DavidoTek:
Add a game list for the [Heroic Games Launcher](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/).  
The game list could be implemented in a similar way as the Steam game list.

## Details
(As of December 2022)

Configuration for the Flatpak version of HGL can be found in `~/.var/app/com.heroicgameslauncher.hgl/config/heroic`  
Additional configuration for [Legendary](https://github.com/derrod/legendary) is stored in `~/.var/app/com.heroicgameslauncher.hgl/config/legendary`  

<details>
<summary><b>Games: Side-loaded apps</b></summary>

The file `~/.var/app/com.heroicgameslauncher.hgl/config/heroic/store/sideload_apps/library.json` contains a list of manually added games. Example:

```json
{
	"games": [
		{
			"runner": "sideload",
			"app_name": "sPZQ5kmzYj5KnZKdxE2bR1",
			"title": "Test Title",
			"install": {
				"executable": "test",
				"platform": "Windows"
			},
			"folder_name": ".",
			"art_cover": "file:///app/bin/heroic/resources/app.asar/build/assets/heroic_card.dae295a6.jpg",
			"is_installed": true,
			"art_square": "file:///app/bin/heroic/resources/app.asar/build/assets/heroic_card.dae295a6.jpg",
			"canRunOffline": true
		}
	]
}
```

</details>

<details>
<summary><b>Games: GOG games</b></summary>

The file `~/.var/app/com.heroicgameslauncher.hgl/config/heroic/store/gog_store/library.json` contains a list of [GOG](https://www.gog.com/) games. Example:

```json
{
	"games": [
		{
			"runner": "gog",
			"store_url": "https://gog.com/en/game/akalabeth_world_of_doom",
			"developer": "Richard Garriott",
			"app_name": "1207666073",
			"art_cover": "https://images-1.gog-statics.com/56262850671fd7232b8ff46deba530007ff7a437fab0ddd67147f8d41aa6bca0.jpg",
			"art_square": "https://images.gog.com/511cf474562bdaf2c8b9eec1a0c94e1eb5cf4e3900a618516adff2ebd89118e0.jpg?namespace=gamesdb",
			"cloud_save_enabled": false,
			"extra": {
				"about": {
[...]

PR #193, opened 2023-02-20
Heroic: Add Info Dialog (GOG, side-loaded, Legendary)

sonic2kk:
Some work is a prerequisite for #168.

Implements support for showing an Info Dialog for Heroic Games compatibility tools :partying_face: To faciliate this, this also implements a `HeroicGame` datastructure, some associated utility functions and some other misc work around this.

image

## Overview
This is heavily work-in-progress for the moment and has only seen light testing.
- No Epic Games integration - Not sure where the `library.json` file is (if it exists), what it looks like, etc
  - However, if the Epic Games data is similarly structured then it should be trivial to include support!
- No Games List integration - It could go in a separate PR, I didn't want to work on it until #192 was merged
  - I also have some other refactor work I want to submit for review for the games list dialog, and I don't want to submit that until the Lutris game list work is reviewed and merged
- Minimal testing of GOG and side-loaded apps - I had never used Heroic before a few hours ago :sweat_smile: 
- Minimal testing of compatibility tools - Only tested Wine-GE
- Only tested for Heroic Wine Flatpak - Did not test Proton Flatpak or Heroic AppImage/Heroic from repos
- Only parsed what I considered relevant from the `library.json` and `GamesConfig/<app_name>.json` - There is certainly room to expand to more fields 

### Performance
In my tests, performance seems pretty good, though I only have 176 games (175 from GOG, 1 side-loaded).

### Oddities
One other thing I would like to note is about the `is_installed` field in the `library.json`. This is set correctly for my side-loaded game, but not for my GOG game. It's installed, and Heroic can see that it's installed, and as well as this it has an entry in the `gog_store/installed.json` file, but it's still set to `false` for me.

For this reason I created a utility function `is_gog_game_installed` that cross-references with the `installed.json` file to check if the app name for the game is in that file.

## Future
When implementing this I tried to keep extensibility and maintenance in mind, for future runners that Heroic might add (if they add Amazon Games etc). It should be mostly straightforward to add support for this, assuming they keep the `library.json` formats the same, which I would *assume* they would :sweat_smile: 

Getting games from a new runner should really only be a case of adding its path to the `store_paths` list in `get_heroic_game_list`.

## Improvements
At the very least I would like to include support for showing games using the Epic/Legendary(?) runner in this PR, but I do not use the Epic Games Store so I don't know how Heroic stores data about it - And it doesn't appear to store that data unless you link an Epic Games account, which I don't have.

Also, any type hinting suggestions/improvements are welcome, as well as thoughts on implementing docstrings.

<hr>

Very open to feedback on this, especially around the information stored for `HeroicGame`s. I'm not sure if there is more we want to store, or if we're storing too much, so please feel free to give suggestion here :-) I tried to keep the rest of the code as concise as I could.

I don't want to submit something half-baked so if there is any further work that should be done here, or if you would prefer this PR to (eventually) capture all of #168 that's fine too. I'd like to fully implement #168 but I thought it better to spread it out across two PRs: One for the ctinfo dialog (as it would include creating the `HeroicGame` datastructure-related groundwork and getting a proof-of-concept of using that data in a table) and one for the Games List.

Further testing is welcome, as well as help on implementing Epic Games support.

Thanks! :-)

### DIFF
--- a/pupgui2/datastructures.py
+++ b/pupgui2/datastructures.py
@@ -166,10 +166,10 @@ class LutrisGame:
 
 
 # Information for games is stored in a per-storefront 'library.json' - This has most of the information we need
-# Information for installed Epic games is stored at '<heroic_dir>/legendary/installed.json' and has a separate JSON format but the same data we need -  
-# Information about game config (sideload, GOG, Epic) is stored in a 'GamesConfig/<app_name>.json' file, which is universal between GOG and sideloaded apps - This has Wine information
+# Information for installed Epic games is stored at '<heroic_dir>/legendary/installed.json' and has a separate JSON format but the same data we need
+# Information about game config (sideload, GOG, Epic) is stored in a 'GamesConfig/<app_name>.json' file, which is universal - This has Wine information
 class HeroicGame:
-    runner: str  # can be 'GOG', 'sideload' - all known for now
+    runner: str  # can be 'GOG', 'sideload' - Epic is hardcoded to 'legendary'
     app_name: str  # internal name encoded in some way, e.g. 'sPZQ5kmzYj5KnZKdxE2bR1'
     title: str  # Real name for game
     developer: str  # May be blank for side-loaded games
--- a/pupgui2/datastructures.py
+++ b/pupgui2/datastructures.py
@@ -175,7 +175,6 @@ class HeroicGame:
     developer: str  # May be blank for side-loaded games
     install: Dict[str, str]  # e.g. executable, platform details
     install_path: str  # e.g. '~/.config/heroic', '~/.var/app/com.heroicgameslauncher.hgl/config/heroic'
-    folder_name: str
     store_url: str  # May be blank for side-loaded games
     art_cover: str  # Optional?
     art_square: str  # Optional?
--- a/pupgui2/datastructures.py
+++ b/pupgui2/datastructures.py
@@ -6,6 +6,7 @@ import json
 from enum import Enum
 from typing import Dict
 
+
 class SteamDeckCompatEnum(Enum):
     UNKNOWN = 0
     UNSUPPORTED = 1
@@ -82,7 +83,8 @@ class SteamApp:
 
     def get_shortcut_id_str(self) -> str:
         return str(self.shortcut_id)
-    
+
+
 class BasicCompatTool:
     displayname = ''
     version = ''
--- a/pupgui2/datastructures.py
+++ b/pupgui2/datastructures.py
@@ -165,9 +165,8 @@ class LutrisGame:
 
 
 # Information for games is stored in a per-storefront 'library.json' - This has most of the information we need
-# Information about game config is stored in a 'GamesConfig/<app_name>.json' file, which is universal between GOG and sideloaded apps - This has Wine information
-#
-# TODO Epic support
+# Information for installed Epic games is stored at '<heroic_dir>/legendary/installed.json' and has a separate JSON format but the same data we need -  
+# Information about game config (sideload, GOG, Epic) is stored in a 'GamesConfig/<app_name>.json' file, which is universal between GOG and sideloaded apps - This has Wine information
 class HeroicGame:
     runner: str  # can be 'GOG', 'sideload' - all known for now
     app_name: str  # internal name encoded in some way, e.g. 'sPZQ5kmzYj5KnZKdxE2bR1'
--- a/pupgui2/datastructures.py
+++ b/pupgui2/datastructures.py
@@ -1,8 +1,10 @@
 import os
-from enum import Enum
 import vdf
 import yaml
+import json
 
+from enum import Enum
+from typing import Dict
 
 class SteamDeckCompatEnum(Enum):
     UNKNOWN = 0
@@ -160,3 +162,29 @@ class LutrisGame:
             return {}
         with open(lutris_game_cfg, 'r') as f:
             return yaml.safe_load(f)
+
+
+# Information for games is stored in a per-storefront 'library.json' - This has most of the information we need
+# Information about game config is stored in a 'GamesConfig/<app_name>.json' file, which is universal between GOG and sideloaded apps - This has Wine information
+#
+# TODO Epic support
+class HeroicGame:
+    runner: str  # can be 'GOG', 'sideload' - all known for now
+    app_name: str  # internal name encoded in some way, e.g. 'sPZQ5kmzYj5KnZKdxE2bR1'
+    title: str  # Real name for game
+    developer: str  # May be blank for side-loaded games
+    install: Dict[str, str]  # e.g. executable, platform details
+    folder_name: str
+    store_url: str  # May be blank for side-loaded games
+    art_cover: str  # Optional?
+    art_square: str  # Optional?
+    is_installed: bool  # Not always set properly by Heroic for GOG?
+    wine_info: Dict[str, str]  # can store bin, name, and type - Has to be fetched from GamesConfig/app_name.json
+
+    def get_game_config(self, install_path: str):
+        game_config = os.path.join(install_path, 'GamesConfig', f'{self.app_name}.json')
+        if not os.path.isfile(game_config):
+            return {}
+        
+        with open(game_config, 'r') as gcf:
+            return json.load(gcf).get(self.app_name, {})
--- a/pupgui2/datastructures.py
+++ b/pupgui2/datastructures.py
@@ -174,6 +174,7 @@ class HeroicGame:
     title: str  # Real name for game
     developer: str  # May be blank for side-loaded games
     install: Dict[str, str]  # e.g. executable, platform details
+    install_path: str  # e.g. '~/.config/heroic', '~/.var/app/com.heroicgameslauncher.hgl/config/heroic'
     folder_name: str
     store_url: str  # May be blank for side-loaded games
     art_cover: str  # Optional?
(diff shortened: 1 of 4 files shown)


sonic2kk:
I tested one build of GE-Proton from the Heroic Proton Flatpak list and it worked. Of course more testing is welcome here too :-)

As well as this, I added some logic to set the `ct.no_games` for Heroic. This isn't present for Lutris but this could be looked into for a separate PR, I just thought it would be a nice addition for Heroic to have. Full parity can come to all install dirs with time 😄

DavidoTek:
> No Epic Games integration - Not sure where the library.json file is (if it exists), what it looks like, etc 

As far as I can tell, Heroic doesn't keep track of installed games, but rather uses legendary for that. It stores information about installed games with a different structure in `~/.var/app/com.heroicgameslauncher.hgl/config/legendary/installed.json`. The app id correspondents with the name of the json file in `GamesConfig`.

> No Games List integration - It could go in a separate PR,

Yes, no rushing. Adding the basic functionality is already great!

> It's installed, and Heroic can see that it's installed, and as well as this it has an entry in the gog_store/installed.json file, but it's still set to false for me.

Well that is strange. Maybe a problem with Heroic?
Still good you added the workaround.

> When implementing this I tried to keep extensibility and maintenance in mind [...] Getting games from a new runner should really only be a case of adding its path to the store_paths list in get_heroic_game_list.

That's great! Yeah I guess they will use the same format for all stores, at least I hope so.

> type hinting suggestions/improvements are welcome, as well as thoughts on implementing docstrings.

Looks good so far. There are some places where no parameter type is specified, they are mostly strings.
If the parameter isn't self explanatory (e.g. it's not clear whether a path, name or object is expected), a short description what is expected and what the function does can be added.

---

Thanks! :tada:

pupgui2/datastructures.py
```diff
@@ -160,3 +162,29 @@ def get_game_config(self):
...
+    developer: str  # May be blank for side-loaded games
+    install: Dict[str, str]  # e.g. executable, platform details
+    folder_name: str
+    store_url: str  # May be blank for side-loaded games
+    art_cover: str  # Optional?
+    art_square: str  # Optional?
+    is_installed: bool  # Not always set properly by Heroic for GOG?
+    wine_info: Dict[str, str]  # can store bin, name, and type - Has to be fetched from GamesConfig/app_name.json
+
+    def get_game_config(self, install_path: str):
```
DavidoTek, line 184: Can `install_path` be stored as a attribute in `HeroicGame`?

pupgui2/heroicutil.py
```diff
@@ -0,0 +1,67 @@
...
+        hg.art_square: str = game.get('art_square', '')
+        hg.is_installed: str = game.get('is_installed', False) or is_gog_game_installed(hg, install_path)  # Some installed gog games may not be marked properly in library.json, so cross-reference with installed.json
+        hg.wine_info: Dict[str, str] = hg.get_game_config(install_path).get('wineVersion', {})
+
+        hgs.append(hg)
+    
+    return hgs
+
+
+def is_heroic_launcher(launcher) -> bool:
```
DavidoTek, line 47: `launcher: str`
Maybe add a docstring, something like `Returns True if the supplied launcher name is a valid name for Heroic Games Launcher, e.g. "heroicwine"`

pupgui2/datastructures.py
sonic2kk: Yup, no problem! I just saw that the `lutrisutil` function does this too. As a nice bonus we should be able to get rid of passing `install_path` to HeroicGame#get_game_config`, since each `HeroicGame` should be aware of its install path, and likewise for `is_gog_game_installed`

For clarity, this path refers to the "base" Heroic install path, which can vary between Flatpak (uses `~/.var/app/com.heroicgameslauncher.hgl/config/heroic`) and AppImage/repo/etc (uses `~/.config/heroic`).

pupgui2/heroicutil.py
sonic2kk: Thanks! Updated with 853c82a.

sonic2kk:
> As far as I can tell, Heroic doesn't keep track of installed games, but rather uses legendary for that. It stores information about installed games with a different structure in `~/.var/app/com.heroicgameslauncher.hgl/config/legendary/installed.json`. The app id correspondents with the name of the json file in `GamesConfig`

Ah okay. I'm not sure if we want a separate datastructure for `LegendaryGame` or anything, or if we should have a separate function (probably called from `get_heroic_game_list`) to build a list of `HeroicGame`s from the Legendary json structure. I'll take a look at how to implement this, hopefully it won't be too complex :-) Thanks!

> Maybe a problem with Heroic?

I think so, I haven't tested on other machines so I'm not sure if this is something specific to this PC somehow. I tried a couple of other GOG games ("Dungeon Keeper Gold" and "DOOM I Enhanced") and they were not marked as installed either. So all 3 I have tested are not marked properly.

<details>
<summary>Another Heroic oddity that might mildly impact usage:</summary> 
I noticed another Heroic "oddity" that applies to GOG and sideloaded games. During install, selecting tools installed to `~/.var/app/<heroic>/config/heroic/tools/wine` or `tools/proton` (both manually and with ProtonUp-Qt) will cause Heroic to show this:

image

Games won't start when this text shows up, and as a side-effect these games won't show up in the `ctinfodialog` in ProtonUp-Qt. I have to select a different Wine and then re-select this tool to get the warning goes away and for the the game to run. After this the `ctinfodialog` always works. This has to be done with *every game* set to use *any* tool in these folders, installed with or without ProtonUp-Qt.

My guess is that the Wine path is not properly written out to the Heroic config files and that's why _ProtonUp-Qt_ doesn't see it, as the implementation I did should be correct as it works when Heroic does not erroneously show this warning text. I don't think there's a good way to work around this one, and I think this is a Heroic bug as it happens when installing the tools manually.
</details>

<hr>

The next step now is implementing Legendary support. I'll see if I can "mimic" a Legendary `installed.json` based on the example in #168, and use that to test out reading from. But as I don't have any "real" games to test it's definitely possible there could be some teething bugs :-)

sonic2kk:
Did an initial attempt at implementing Legendary game fetching based on the example `installed.json`. It's possible that "real" installed games may have other quirks but it works for now :-)

I also just assume that Legendary game configs are stored alongside GOG and sideloaded configs in `Heroic/GamesConfig/<app_name>.json`, as the base "Legendary" folder didn't have any config folder. If it doesn't we may need to adjust `get_game_config` to fetch (and potentially) parse the config json separately. Otherwise if the Legendary config structure and location are the same, the Wine information should also be fetched properly now :partying_face: 

<hr>

In that same commit I also removed the `folder_name` attribute. It is not set at all for Legendary games and does not seem to be set for GOG games. It is only set for side-loaded apps. For GOG and Heroic, they store a `install_path` in the `install.json`, which points to a folder which would be equivalent to a `folder_name`.

The `install_path` value would also conflict with *our* `install_path`, which points to the Heroic install path.

I have an idea on how to improve this and will push a commit with a refactor :-)

sonic2kk:
So in my research it seems like some but not all GOG games set `folder_name`. Installed games will _usually but not always_ have it, and it is a bit inconsistent.

- `folder_name` for sideloaded games refers to the full path and is stored in `library.json`.
- `folder_name` for GOG games refers to only the install folder, but is not always set, and is stored in the `library.json`
- `folder_name` is not set at all for Legendary games
- `install_path` is used as the equivalent to a sideloaded games' `folder_name`, whereby `install_path` for GOG and Legendary points to the full install folder path as opposed to just the folder name, and this is stored in the `installed.json` (which afaict there is no `installed.json` for sideloaded apps)

I did a bit of a refactor as follows:
- Rename our old `install_path`, which referred to the Heroic install path (e.g. `~/.config/heroic`) to `heroic_path`
- Add a `install_path` attribute which points to the game install path - We then fetch this from the `installed.json` for GOG and Legendary games, and from the `library.json` for sideloaded games
- Create a new helper function called `get_gog_installed_game_entry` which returns the installed GOG game's JSON object entry in `installed.json` or `{}` if the game was not found
- Refactor `is_gog_game_installed` to use this new function. Since we only care in this case if the entry exists we can cast the return dict to a `bool`, where `{}` will be `False` and a dict with any values will be `True`
    - The logic to return the GOG game's JSON object mirrors what was already in the old `is_gog_game_installed`, but that function used to return `True` or `False` instead of a dict. Therefore we can just cast the returned Dict as a bool

Hope the approach I used makes sense :-)

pupgui2/heroicutil.py
```diff
@@ -0,0 +1,95 @@
...
+
+    return bool(get_gog_installed_game_entry(game))
+
+
+def get_gog_installed_game_entry(game: HeroicGame) -> Dict:
+    """ Return JSON entry as dict for an installed GOG game from heroic/gog_store/installed.json """
+
+    gog_installed_json_path = os.path.join(game.heroic_path, 'gog_store', 'installed.json')
+    if not os.path.isfile(gog_installed_json_path) or not game.runner == 'gog':
+        return ''
```
DavidoTek, line 88: Maybe replace `return ''` with `return {}`

sonic2kk:
The Legendary integration may need further review and testing before this is "fully" ready, I'm unable to test it beyond mimicking files for Legendary games. If you find issues and want to directly push a fix please feel free! Otherwise I'm happy to take on board further feedback - I appreciate all the time you put into reviewing my PRs :-)

sonic2kk:
Just noticed a couple more comments I should probably update that were written before we figured out the Epic part

DavidoTek:
I made a few little changes to the code (optimization, whitespaces (this PR/general), changed data type hints).
I think we can merge it now. Is there anything else to be added/changed?

sonic2kk:
Changes look good to me, I don't have anything further to do on this that comes to mind 😄 From my testing it should be ready to go!

PR #197, opened 2023-02-24
Fix Steam games list regression from #183

sonic2kk:
I introduced a small regression in #183 that I somehow did not catch. When an appmanifest didn't exist for a SteamApp, the entire app list loading would trigger the exception in `get_steam_app_list`, meaning a lot of tools were erroneously listed as unused.

I added a check before attempting to load the vdf file to ensure it is actually a valid file.

### DIFF
--- a/pupgui2/steamutil.py
+++ b/pupgui2/steamutil.py
@@ -52,8 +52,8 @@ def get_steam_app_list(steam_config_folder: str, cached=False, no_shortcuts=Fals
                 fid_steamapps_path = os.path.join(fid_libraryfolder_path, 'steamapps')  # e.g. /home/gaben/Games/steamapps
                 appmanifest_path = os.path.join(fid_steamapps_path, f'appmanifest_{appid}.acf')
                 if os.path.isfile(appmanifest_path):
-                    appmanifest_install_path = vdf.load(open(appmanifest_path)).get('AppState', {}).get('installdir')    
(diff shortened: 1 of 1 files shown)


sonic2kk:
Actually, it wasn't introduced in that PR... Almost certainly from #183 ? Though I didn't notice this bug back then...

I have been downloading, moving, and uninstalling a bunch of games these last couple of days (new ISP, wanted to test speeds) so I wonder if perhaps this was some semi-edge case that I didn't discover back then that has only reared its head now.

I also noticed an issue with the `.get` where I forgot to add a fallback if it can't get `installdir`. It falls back to `None`, then we do a check to ensure that the appmanifest path was found to begin with, and then check to ensure the found path exists. This truthy check should also cover blank paths, ensuring that `isdir` will not fire True in the case of `/path/to/steamapps/common` existing (which it always should).

sonic2kk:
Huh, after opening Steam, the unused issue goes away even on main...

Well, I guess this PR may not be needed in the end. Maybe the checks make sense to have anyway, just in case?

sonic2kk:
> Maybe the .acf files get created after they are downloaded to the library, so ProtonUp-Qt already detected the game but the .acf file didn't exist yet

I was considering too that this could be the case, it definitely seems like some kind of edge case that shouldn't normally happen. Or maybe it was a hiccup on the Steam side, either way having more robust checking could only be a good thing I think :-)

PR #198, opened 2023-02-24
Games List: Move Steam-Specific Logic to steamutil 

sonic2kk:
Initially the Games List really only managed Steam game stuff. Lutris recently got support for showing a Games List, and in future hopefully there will also be a Heroic games list.

No screenshot here as there should be no user-facing change, just "backend" changes :-)

This PR splits the ProtonDB fetching logic and the Anti-Cheat status logic out into separate functions in `steamutil`.  This is not really faster or anything as far as I can tell (nor does it appear to be any slower), but the hope is to make the games list a little bit cleaner. Instead of having all of this logic in the games list, it's separated out into utility functions.

This also introduces the possibility of fetching ProtonDB and Anti-Cheat status outside of the games list dialog if it's ever needed, though I don't know when this would be used.

There are also some other small tweaks in this file that I noticed and thought would fit in this PR too, but the main goal of this was to break down the logic in the games list to make it a bit more clean and hopefully easier to maintain as this dialog could grow to accommodate more game lists.

Hope this makes sense and please let me know if there are any further changes/improvements that should be made, or if there are any changes made here that should be undone.

Thanks!

### DIFF
--- a/pupgui2/pupgui2gamelistdialog.py
+++ b/pupgui2/pupgui2gamelistdialog.py
@@ -38,8 +38,7 @@ class PupguiGameListDialog(QObject):
     def load_ui(self):
         data = pkgutil.get_data(__name__, 'resources/ui/pupgui2_gamelistdialog.ui')
         ui_file = QDataStream(QByteArray(data))
-        loader = QUiLoader()
-        self.ui = loader.load(ui_file.device())
+        self.ui = QUiLoader().load(ui_file.device())
 
     def setup_ui(self):
         if self.install_loc.get('launcher') == 'steam':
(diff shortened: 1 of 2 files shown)


DavidoTek:
I agree it is a good idea to those into separate function. That makes the code a lot cleaner :)

> into separate functions in `steamutil`.

You pointed out this makes the AWACY strings available to other parts of ProtonUp-Qt.
This is a valid argument though I'm not sure whether we should put those functions in `steamutil` as it contains more backend stuff at the moment. Also, they are only used by the game list at the moment.

I also played around a bit with the i18n part.
`self.tr(some_dynamic_variable)` sadly doesn't work (or at least is not intended to work without work-arounds :smile:). Every string needs own `self.tr('')`. If we move `get_steamapp_awacystatus` and `get_steamdeck_compatibility` to the game list class, nothing needs to be changed in the translation files, though the functions are only ment to be accessed by the game list then.
Otherwise we would need to change the "context" of the translations in the translations files and use `QApplication.instance().translate('<context>', '<text to translate>')` in steamutil.

So I personally would put `get_steamapp_awacystatus` and `get_steamdeck_compatibility` as methods in the game list class for now. We can change it later if needed.
What do you think?


Thanks!

DavidoTek:
Okay.

>  If there comes a time when we absolutely need the Steam Deck compatibility/anti-cheat status somewhere else, that can be resolved separately :-)

sonic2kk:
Ah, then yup please go ahead and push if you already have it :smile:

PR #199, opened 2023-02-27
Heroic: Add Basic Games List

sonic2kk:
This PR implements #168, at least in a basic capacity :-) There is also some slight refactoring and additions to the fields stored in `HeroicGame`.

### Layout
Here's the layout I came up with:

image

Note: There are no Legendary games in the screenshot as my test `installed.json` games don't have corresponding `GamesConfig/<app_name>.json` files, so they were showing up as native games, and so I decided to exclude them as this would not normally happen in practice.

### Implementation
The data for this list is fetched from a list of `HeroicGame`s. The Heroic Wine and Heroic Proton install locations share a gamelist, I felt that from a *user* perspective there is not much reason to split them up. In my mind I think it would be more frustrating to have to switch between each install directory just to see the games used. It makes sense for individual compatibility tools but in my mind it makes a bit less sense for viewing a list of games. But that's also just *my* opinion :-)

Plus, if they were split, native games would be shown on both, leading to inconsistencies there as well.

However if it is preferred to split them, it shouldn't be too hard of a change. We could just set another filter condition for the heroic games list fetching to only show those with the relevant `type` in their `wine_info` based on the `self.launcher` value.

Like with the Steam and Lutris game lists, I have implemented some tooltip information and double-click functionality where I thought it made sense to do so.
- Game Name
    - Tooltip shows `Game Name`, and appends ` (executable_name)` if the game executable can be found. Game name is shown in tooltip in case the game name is too long.
    - Double click opens the `store_url` using `xdg-open`, if the `store_url` is valid (not set for sideloaded games, _should_ be set for GOG games, unsure about legendary?)
- Compatibility Tool - Shows the name of the compatibility tool with `Wine - ` or `Proton - ` stripped off (e.g. `Proton 7.0` instead of `Proton - Proton 7.0`), or `Native` for native Linux games
    - Tooltip shows the name, path, and type (`Wine`/`Proton`) in that order if each can be found, and simply shows `Type: Native` for native Linux games
    - No double-click action, though perhaps we could bind this to opening the compatibility tool `path` (if present)?
-  Install location - Path to the game's base folder as per `HeroicGame.install_path`
    - Tooltip shows "Double click to browse..." if path is valid, otherwise shows "Install location does not exist!" -- This mirrors Lutris games list
    - Double click action will open the path in the user's default file manager, if it is valid
- Runner
    - No tooltip or double-click action

### Considerations
The data I chose to include here is purely based on my own initial thoughts of what would be good to include. Please let me know if there are better choices or more preferable choices. Of course, users will also be a good way to gauge what should be shown, so I didn't spend *too* long deliberating and just tried to show what I felt would be the most useful and what matched the other game lists best.

I don't own any Legendary games to test this with, so please do test with some Legendary games installed!

Speaking of Legendary, I couldn't find out if those games have a `store_url`. Heroic has a button in its hamburger menu for GOG games which stores lets you view the store page. I would be a bit surprised if it didn't have this for Legendary. In the example in #168 I did see a `base_urls`, but this seems to be for downloading and not for a store page. Perhaps this is stored elsewhere, and if so we can store this on legendary `HeroicGame`s, and then the double-click functionality will work for Legendary games too :smile: 

### `HeroicGame` data
I updated the `HeroicGame` datastructure a little bit. The main change was that I removed the `install` dict, which is based on the `install` dict for Heroic games. The issue is that this can store varying information. Instead, I decided to remove that and store the fields we would be mostly interested in as separate attributes. Not all `install` sections of a Heroic Game JSON will store the same data, so it didn't make much sense looking back to try and shoehorn our `HeroicGame` class to do this.

Now, we have three new fields on the `HeroicGame`:
- `platform` - The *current platform* version of a game e.g., even if a game supports Windows/macOS/Linux, this only stores the platform for the current installed version, e.g. "Linux"
- `executable` - The basename for the executable which is what Legendary stores and what GOG stores *generally* - Sometimes GOG stores paths like `DOSBox\game.exe`
    - `executable` for GOG games seems to be a blank entry in `installed.json`, but in my research I discovered that Wine/Proton games store the executable in a separate way I'll describe later, and native games default to `start.sh`.
    - Sideloaded  games store the full path
    - Legendary games seem to just store the exe basename, but I only have one example to work off of.
- `is_dlc` - Denotes whether a game is DLC, and we filter this out of the Games List -- This field is not set for sideloaded games but should be set for GOG and legendary, we default it to `False` to be safe though

Figuring out how to get the executable for GOG games was a bit of a (heroic) adventure. In the end I found out that the executable is stored in a rather roundabout way for Wine/Proton games. In a Wine/Proton game's folder, there will be a file named `goggame-<app_name>.info`, which is a file with a JSON file structure. Here is an example of this file:

<details>
    <summary>DOOM I Enhanced (goggame-2015545325.info)</summary>

```json
{
    "buildId": "56055485420912305",
    "clientId": "55704669042708932",
    "gameId": "2015545325",
    "language": "English",
    "languages": [
        "en-US"
    ],
    "name": "DOOM I Enhanced",
    "osBitness": [
        "64"
    ],
    "playTasks": [
        {
            "arguments": "--bnet-language en",
            "category": "game",
            "isPrimary": true,
            "languages": [
                "en-US"
            ],
            "name": "DOOM I Enhanced",
            "osBitness": [
                "64"
            ],
            "path": "DOOM.exe",
            "type": "FileTask"
        }
    ],
    "rootGameId": "2015545325",
    "version": 1
}
```
</details> 

The executable path (which afaict is the relative path from the install folder) is always stored in the `playTasks` list of objects, and it's always stored on the object whose `name` attribute matches the root `name` attribute -- I checked this with 5 different GOG games and it seemed to be the case. From reading through a few of these files, I would guess that `playTasks` is used for things like alternative game executables that you can use, such as multiplayer servers. Some files like Dungeon Keeper for example had several of the `playTasks`, and we couldn't guarantee that the first entry would be the correct one, so we had to resort to searching the file for the correct `path` to use.

I created a separate utility function in heroicutil to get this value: `get_gog_game_executable` which manages doing this. It defaults to `start.sh` for native GOG games (instead of just an empty string). For Wine/Proton, it searches for this `goggame` info file and attempts to extract the relevant `path` attribute from the correct object.

For the game I tested this seemed to work but as always, the more testing, the better! :smile: 

### Games List Dialog changes
I created a utility method in the games dialog, `set_item_data_directory`, which manages attaching a path to a `QTableWidgetItem`'s data. This logic was duplicated between Lutris and Heroic for the Install Path double-click functionality, so I created a helper method for this.

If we do want to add a double-click action for compat tools which opens their folder, this method could be used.

<hr>

The changes here turned out to be a little bigger than anticipated in the end, with the HeroicGame refactor. Hopefully I justified the changes I made there well enough. Other fields could potentially be removed and that is certainly up for discussion, maybe storing the art paths is not really that useful. Some of the fields stored are not explicitly used at least not yet, but I thought they were good to store as it's good general information to store about a game. Likewise if there is some data that was missed we can always expand the `HeroicGame`.

Thanks! :-)

### DIFF
--- a/pupgui2/constants.py
+++ b/pupgui2/constants.py
@@ -81,4 +81,5 @@ STEAM_STL_DATA_PATH = os.path.join(os.path.expanduser('~'), '.local', 'share', '
 STEAM_STL_SHELL_FILES = [ '.bashrc', '.zshrc', '.kshrc' ]
 STEAM_STL_FISH_VARIABLES = os.path.join(os.path.expanduser('~'), '.config/fish/fish_variables')
 
-LUTRIS_WEB_URL='https://lutris.net/games/'
+LUTRIS_WEB_URL = 'https://lutris.net/games/'
+EPIC_STORE_URL = 'https://store.epicgames.com/p/'
--- a/pupgui2/datastructures.py
+++ b/pupgui2/datastructures.py
@@ -183,7 +183,7 @@ class HeroicGame:
     is_installed: bool  # Not always set properly by Heroic for GOG?
     wine_info: Dict[str, str]  # can store bin, name, and type - Has to be fetched from GamesConfig/app_name.json
     platform: str  # Game platform, stored differently for sideload, GOG and legendary
-    executable: str  # Path to game executable
+    executable: str  # Path to game executable, always stored at 'start.sh' for native Linux GOG games 
     is_dlc: bool  # Stored for GOG and legendary, defaults to False for sideloaded
 
     def get_game_config(self):
--- a/pupgui2/datastructures.py
+++ b/pupgui2/datastructures.py
@@ -175,7 +175,6 @@ class HeroicGame:
     app_name: str  # internal name encoded in some way, e.g. 'sPZQ5kmzYj5KnZKdxE2bR1'
     title: str  # Real name for game
     developer: str  # May be blank for side-loaded games
-    install: Dict[str, str]  # e.g. executable, platform details
     heroic_path: str  # e.g. '~/.config/heroic', '~/.var/app/com.heroicgameslauncher.hgl/config/heroic'
     install_path: str  # Path to game folder, e.g. '/home/Gaben/Games/Half-Life 3'
     store_url: str  # May be blank for side-loaded games
@@ -183,6 +182,9 @@ class HeroicGame:
     art_square: str  # Optional?
     is_installed: bool  # Not always set properly by Heroic for GOG?
     wine_info: Dict[str, str]  # can store bin, name, and type - Has to be fetched from GamesConfig/app_name.json
+    platform: str  # Game platform, stored differently for sideload, GOG and legendary
+    executable: str  # Path to game executable
+    is_dlc: bool  # Stored for GOG and legendary, defaults to False for sideloaded
 
     def get_game_config(self):
         game_config = os.path.join(self.heroic_path, 'GamesConfig', f'{self.app_name}.json')
--- a/pupgui2/heroicutil.py
+++ b/pupgui2/heroicutil.py
@@ -1,9 +1,11 @@
 import os
 import json
+import re
 
 from typing import List, Dict
 
 from pupgui2.datastructures import HeroicGame
+from pupgui2.constants import EPIC_STORE_URL
 
 
 def get_heroic_game_list(heroic_path: str) -> List[HeroicGame]:
@@ -63,6 +65,7 @@ def get_heroic_game_list(heroic_path: str) -> List[HeroicGame]:
             lg.developer: str = ''  # Not stored or stored elsewhere?
             lg.heroic_path: str = heroic_path
             lg.install_path: str = game_data.get('install_path', '') 
+            lg.store_url: str = f'{EPIC_STORE_URL}{re.sub("[^a-zA-Z0-9]", "-", lg.title.lower())}'
             lg.art_cover: str = ''  # Not stored or stored elsewhere?
             lg.art_square: str = ''  # Not stored or stored elsewhere?
             lg.is_installed: str = True  # Games in Legendary `installed.json` should always be installed
--- a/pupgui2/heroicutil.py
+++ b/pupgui2/heroicutil.py
@@ -35,12 +35,22 @@ def get_heroic_game_list(heroic_path: str) -> List[HeroicGame]:
         hg.heroic_path: str = heroic_path
         # Sideloaded games uses folder_name as their full install path, GOG games store a folder_name but this is *just* their install folder name
         # Prioritise getting install_path for GOG games as this is the GOG game equivalent to 'folder_name'
-        hg.install_path: str = get_gog_installed_game_entry(hg).get('install_path', '') if hg.runner == 'gog' else game.get('folder_name', '')
+        hg.install_path: str = get_gog_installed_game_entry(hg).get('install_path', '') if hg.runner.lower() == 'gog' else game.get('folder_name', '')
         hg.store_url: str = game.get('store_url', '')
         hg.art_cover: str = game.get('art_cover', '')  # May need to replace path if it has 'file:///app/blah in name - See example in #168
         hg.art_square: str = game.get('art_square', '')
         hg.is_installed: bool = game.get('is_installed', False) or is_gog_game_installed(hg)  # Some installed gog games may not be marked properly in library.json, so cross-reference with installed.json
         hg.wine_info: Dict[str, str] = hg.get_game_config().get('wineVersion', {})
+        # Sideloaded games store platform in its library.json (it has no installed.json) under the 'install' object
+        # GOG games store the platform for the version of the installed game in `installed.json` (as GOG games can target multiple platforms, installed will show if the user has the Windows or Linux version)
+        hg.platform: str = game.get('install', {}).get('platform', '').capitalize() if hg.runner.lower() == 'sideload' else get_gog_installed_game_entry(hg).get('platform', '').capitalize()  # Capitalize ensures consistency
+        # TODO get executable for GOG games? Seems to be stored in some game dir files with inconsistent structure?
+        # Wine games seem to have `goggame-<app_name>.info` file in their installed folder (JSON format), inside is a a "playTasks" list of objects -- The one with the "name" which matches the `folder_name` (can get from dirname(install_path)) has the `path` to the executable
+        # How do native games do it though?
+        #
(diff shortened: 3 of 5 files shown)


DavidoTek:
> The Heroic Wine and Heroic Proton install locations share a gamelist, I felt that from a user perspective there is not much reason to split them up

Agreed.
The only reasons they are split up in the first place are that it is easier to handle internally and most users probably only want to use Wine anyway and Proton is more of a bonus.

> Tooltip shows Game Name, and appends  (executable_name) if the game executable can be found. Game name is shown in tooltip in case the game name is too long.

That's useful.

> Double click opens the store_url using xdg-open, if the store_url is valid (not set for sideloaded games, should be set for GOG games, unsure about legendary?)

I think Epic Games URLs are using the name, but lower-case and spaces and special characters replaced with `-`. Though I'm not sure if that is always the case.
Example: `Destiny 2: Legacy Collection` -> https://store.epicgames.com/p/destiny-2--legacy-collection

> though perhaps we could bind this to opening the compatibility tool path (if present)?

Why not.

> Of course, users will also be a good way to gauge what should be shown, so I didn't spend too long deliberating and just tried to show what I felt would be the most useful and what matched the other game lists best.

Looks good to me. If users want anything else, they will probably open an issue.
I also don't think we need something like the AreWeAntiCheatYet.com status as Heroic should already include this (at least it includes an `areweanticheatyet.json` file).

> I decided to remove that and store the fields we would be mostly interested in as separate attributes.

Looks very clean now.


---

In my testing, all columns populated correctly for gog, sideloaded and legendary games. Double-clicking also opened the URL/folder just fine (tested 7 games).
Detection of the executable also worked fine even when they are in a subdirectory.

---

Thanks!

sonic2kk:
> I think Epic Games URLs are using the name, but lower-case and spaces and special characters replaced with -. Though I'm not sure if that is always the case.

We could add build `store_url` for legendary games like this then, storing the Epic Games base store URL as a constant: `lg.store_url = f'{EPIC_GAMES_STORE_URL}{re.sub("[^a-zA-Z0-9]", "-", lg.title.lower())}'` where the regex matches any character that isn't alphanumeric (case insensitive) with a `-`.

Indeed this may not always be the case, I'm not sure if there's a good way that we could sanity check a URL. I haven't checked the code to see how Heroic gets the Epic store URL, if at all. Based on the URL you gave though it *seems* like it could be a safe assumption, though I wonder what happens in cases where a name changes. This is likely a hardcore edge-case but: Say "Destiny 2" gets renamed to "Destiny Two" and the Epic store URL updates, but it's already been downloaded from Heroic by a user. Maybe Heroic refreshes the downloaded names but if not, ProtonUp-Qt may build the wrong URL (using `destiny-2` instead of `destiny-two`, if the Epic URL updates at all that is).

Again that's very likely an extreme edge-case, and Heroic may refresh the next time it's opened, but it's something that came into my mind. Maybe having something that's *generally* correct (or maybe even very often correct) is better than not having anything :sweat_smile: 

> Looks good to me. If users want anything else, they will probably open an issue.
I also don't think we need something like the AreWeAntiCheatYet.com status as Heroic should already include this (at least it includes an areweanticheatyet.json file).

Yeah Heroic lists this along with some info from `howlongtobeat.com`, and possibly other info I haven't seen. It also lets you view a game's compatibility I think in the hamburger menu, which redirects to ProtonDB - Though this may not always be totally accurate as there are cases where builds of games from GOG/Epic don't work the same as Steam builds and vice versa (can't find the issue now, but I remember the GOG build of Cyberpunk 2077 didn't work initially at launch, but the Steam version did). It may not be worth showing this either as, once again, Heroic itself can show this information to users :-)

> In my testing, all columns populated correctly for gog, sideloaded and legendary games. Double-clicking also opened the URL/folder just fine (tested 7 games).
> Detection of the executable also worked fine even when they are in a subdirectory.

Woohoo :partying_face: Thank you for testing.

<hr>

I'll push up a couple more changes :-)

sonic2kk:
Pushed a change that:
* attempts to build a base Epic Games URL, may not always be valid but could be better than no URL. We can remove it if preferred
* adds a double-click action for Wine/Proton compatibility tool cells which opens their base directory. If we have a path given in the `wine_info` of a heroic game, we split the path here based on the name of the compatibility tool (which we split earlier on for the cell text), and then use `os.path.join` to join the path.
    * For example, say the tool name (which we use to build the cell name) is `Proton - GE-Proton7-42`. We set the cell text currently by splitting at the first `-` and then `strip`ping the trailing whitespaces from the name. This is effectively the folder name for the tool
    * If we have a `bin` in the `wine_info` for a Heroic Game, it will be a valid path (otherwise Heroic won't write it out to the GamesConfig JSON afaik, when it shows me the invalid Heroic text then it won't display the path in ProtonUp-Qt), this path is to the actual Wine binary (e.g. `/path/to/GE-Proton7-42/bin/wine`). So we split this path based on the name we got earlier since this is the folder name. So `/path/to/GE-Proton7-42/bin/wine` gets split at `GE-Proton7-42`, and we take the first split (which is `/path/to/`) and use `os.path.join` to then essentially do `os.path.join('/path/to', 'GE-Proton7-42')`, and now we have the path to the compat tool base folder!
    * Since different compatibility tools (e.g. GE-Proton vs lutris-GE) have varying paths (outlined a little in a comment in #168), this means we don't have to do any checks and we can just build the base directory in a (imo) relatively straightforward way

Not sure how many people will use the double-click functionality but I think it's nice to have :-)

DavidoTek:
Thanks!

> attempts to build a base Epic Games URL, may not always be valid but could be better than no URL. We can remove it if preferred

Good. I cannot remember a game where they changed its name, but that's really an edge case and Heroic should deal with, i.e. they probably want the name to change in Heroic as well so they would add a mechanism for that if it ever happens.

> adds a double-click action for Wine/Proton compatibility tool cells which opens their base directory

Great, even works fine when the display name doesn't match.

> Not sure how many people will use the double-click functionality but I think it's nice to have :-)

Yeah, I'm wondering if we should document the "hidden functionality" somewhere or leave it as an easter egg :)

sonic2kk:
> Yeah, I'm wondering if we should document the "hidden functionality" somewhere or leave it as an easter egg :)

Hehe, some of it is """documented""" in that the game install path cells for Heroic and Lutris both have a tooltip that says you can double click to browse. Though then there's also the case of how many users would actually read docs :wink: No objections from me either way!

> PS: Regarding the translations, there need to be some changes. I will create a commit...

Thanks for always being able to help with that :smile:

PR #200, opened 2023-03-09
Show Tool Count on Main Menu, Show Total Installed Games on Games List DIalog

sonic2kk:
This PR does a few things:
- Show total installed tools for a given launcher beside its "Installed compatibility tools:" label.
- To the right of this, if there are any unused tools, display a label that says "Unused: {num_unused_tools}" (this is hidden if there are no unused tools).
    - This currently only applies to Steam and Heroic, as there is no logic for checking unused Lutris tools
- On the Games List dialog, display the total found installed games.
    - I originally thought to show this count in the dialog title, but in future this may not always be visible (i.e. if running in Steam Deck Game Mode), so I went with adding a label despite it taking up slightly more real-estate in the dialog than I'd like

It can be useful at a glance to see how many compatibility tools you have installed and, if you have a lot installed, how many are unused. It can help highlight if, for example, you end up in a scenario where you have a bunch of old GE-Proton versions laying around.

## Screenshots
image

image

## Overview
I've actually been sitting on this one for a while, and was very back and forth on how the UI should look. I think re-using the "Installed compatibility tools" label on the main menu is very natural. I experimented with a lot of different positions and such for the unused text, putting it in the bottom-right, putting it in brackets, different wordings, and in the end I landed on the "Unused: X" in the top-right to the side of the installed tools count. The positioning and wording mirrors the active downloads text above it. However I am absolutely open to any kind of improvements here. I'm no UI designer but I still like to make interfaces better where I can :-)

Finally with the games list dialog, as mentioned the label takes up slightly more real-estate than I'd like, but I wasn't sure where else to put it. The titlebar is kind of out imo. There could be semi-hacky conditional checks to display it in the titlebar under normal circumstances, and only append a label programmatically if we're running in Game Mode, but window decorations can be hid normally anyway, so I just don't feel like it's a solid approach to take. The bottom-left section of the dialog is out too as that's where the Steam warning label goes, and they could end up conflicting. If the Steam warning text is present it would really throw off the "look" of the label. So the top-left position seemed to be the only logical place left to put it.

One last note is that, at least to my eye, it feels like there is a noticeable amount of spacing between the "Installed compatibility tools" label and the actual count, and the same goes for the total installed games count on the games list. I couldn't figure out a way to resolve this, I would guess that this is because they are two separate labels and so they have some padding and such. If there's a way to fix it that's worth the effort I'm definitely open to improving that :-)

<hr>

Thanks!

### DIFF
--- a/pupgui2/pupgui2.py
+++ b/pupgui2/pupgui2.py
@@ -185,6 +185,7 @@ class MainWindow(QObject):
     def update_ui(self):
         """ update ui contents """
         install_loc = get_install_location_from_directory_name(install_directory())
+        unused_ctools = 0
 
         self.ui.listInstalledVersions.clear()
         self.compat_tool_index_map = get_installed_ctools(install_directory())
@@ -206,8 +207,12 @@ class MainWindow(QObject):
                 ct.no_games = len(map.get(ct, []))
 
         for ct in self.compat_tool_index_map:
-            self.ui.listInstalledVersions.addItem(ct.get_displayname(unused_tr=self.tr('unused')))
+            ct_displayname = ct.get_displayname(unused_tr=self.tr('unused'))
+            if self.tr('unused') in ct_displayname:
+                unused_ctools += 1
 
+            self.ui.listInstalledVersions.addItem(ct_displayname)
+            
         self.ui.txtActiveDownloads.setText(str(len(self.pending_downloads)))
         if len(self.pending_downloads) == 0:
             self.set_default_statusbar()
@@ -223,7 +228,9 @@ class MainWindow(QObject):
         else:
             self.ui.btnShowGameList.setVisible(False)
 
-        self.ui.txtInstalledVersions.setText(str(len(self.compat_tool_index_map)))
+        # installed_versions_txt = f'{len(self.compat_tool_index_map)}'
+        self.ui.txtUnusedVersions.setText(f'{self.tr("Unused")}: {unused_ctools}' if unused_ctools > 0 else '')
+        self.ui.txtInstalledVersions.setText(f'{len(self.compat_tool_index_map)}')
 
     def get_installed_versions(self, ctool_name, ctool_dir):
         for ct in get_installed_ctools(ctool_dir):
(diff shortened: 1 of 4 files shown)


sonic2kk:
> Doesn't look like anything is missing anymore :)

Heh, I actually feel the same. I've had this for a while and going back to main, it has stood out a lot more to me that it's missing, glad the feedback on this UI tweak PR is positive :-)

> Not sure how well that would look and if that makes sense

Yeah, it might be a bit strange to have it there, like it might stand out a bit too much. A tooltip on the game column would work but I'm not sure if this shows in things like Game Mode either. Though a tooltip would be more in line with other parts of the game menu.

> It makes the actual number protrude.

Ah! You're right, I didn't even think about that. Good point!

sonic2kk:
So I tested out the idea of putting the game count in the tooltip, and it works no problem. I tested in GameScope to emulate Steam Deck Game Mode and tooltips seem to work in GameScope, so it's a viable option:

image

The only time tooltips didn't work was when I used the `--expose-wayland` flag, which afaik is still only available in GameScope-git and still in early stages (most SDL games with the Wayland backend enabled don't work in my tests), so it's probably not a scenario we have to worry about too much though I wanted to mention it nonetheless :-)

This approach I went with assumes that the first column will always be the games column. I can't foresee a case where this wouldn't be true when we have a *games* list, but something to be aware of I suppose.

I would be fine with either approach, or with including both (or neither if it comes to it :sweat_smile:)

DavidoTek:
Looks good.

> I would be fine with either approach, or with including both (or neither if it comes to it sweat_smile)

Personally, I would only keep the tooltip and remove the label as it is consuming quite a lot of space.
Though I wonder if there is any future use for the space next to the label. If we find any, we may add it back then.

> ```diff
> + <height>325</height>
> ```

I think we should keep the height increase anyway :)

sonic2kk:
Removed the label, think I removed any reference to it in the code too (the only one should've been the line to set the text in `setup_ui` I think). And I think the height increase is nice too, just gives that little bit of extra space :-)

PR #207, opened 2023-03-18
Add Tooltip to Custom Install Directory button

sonic2kk:
It isn't immediately obvious at first glance what this button does, so this PR adds a tooltip that describes what the button does.

The rest of the UI imo is fairly self-explanatory and intuitive, and the user only has to click on this tool button to figure out what it does, but a tooltip is a nice quality-of-life improvement I think :-)


DavidoTek:
Thanks!

Issue #206, opened 2023-03-18
[Feature Request] Add More Keyboard Shortcuts


sonic2kk:
**Is your feature request related to a problem? Please describe.**
Nope, this is a *very minor* quality-of-life idea.

**Describe the solution you'd like**
It dawned on me recently that ProtonUp-Qt does not have many keyboard shortcuts. I wanted to open this issue to gauge whether or not it would be desirable to have more keyboard shortcuts and, if so, what those shortcuts should be. Here's a rough list of what I was thinking:

- Quit: Ctrl+Q (#205)
- Show Info dialog: Alt+Enter
    - This shows the Properties dialog in KDE's Dolphin file explorer, but a more universal and intuitive shortcut is very welcome!
- Remove Compat Tool: Del button, or some modifier key + backspace for keyboards which don't have a delete key
- Add Compat Tool: Ctrl+N?
    - Not fully sold on this one to be honest. There is also the possibility of having different keyboard shortcuts as shorthands for opening the new dialog with a default compat tool selected. Maybe something like Ctrl+Alt+G for GE-Proton and so on. Not totally sure on this either, just an idea I had :sweat_smile: 
    - I'm also cautious that maybe this should be used for the Custom Install Directory option. Though I'm not sure how widely used this is, and this option could instead be set to Shift+Ctrl+N
- About dialog: F1
    - This is a somewhat-standard "Help" button, it's even built-in with `QKeySequence.HelpContents`

I should state that I am more than happy to implement these shortcuts myself, but I would like to get some feedback on what the shortcuts should be beforehand.

**Describe alternatives you've considered**
Using the "Tab" key to navigate the UI, which works, but intuitive keyboard shortcuts would be nicer I think. Emphasis on Intuitive, which is a big part of what I want feedback on.

Adding a shortcut to open a dialog is as straightforward as a one-line call to `QShortcut(QKeySequence('shortcut'), self.ui).activated.connect(func)` (or a lambda to here to pass data to an existing method if it requires it). The important part is getting the shortcuts right.

**Additional context**
I don't think it would be fitting necessarily to add these as tooltips, but documenting the shortcuts somewhere might be a good idea. Or they could be left as "easter eggs" heheh ;-)

DavidoTek:
Great idea!
I think users who regularly use ProtonUp-Qt will benefit a lot.

> I should state that I am more than happy to implement these shortcuts myself, but I would like to get some feedback on what the shortcuts should be beforehand.

[...]

PR #209, opened 2023-03-19
pupgui2: Add More Keyboard Shortcuts

sonic2kk:
Implement #206.

This is currently missing the shortcuts for the compatibility tools, but should have the rest as described in that issue:
- Add Compat Tool: `Ctrl+N`
- Add Install Location: `Ctrl+Shift+N`
- Remove Compat Tool: `Delete`, `Ctrl+Backspace` (for keyboards without delete key)
- About: `F1`, `Ctrl+,`
- Compat Tool Info: `Alt+Enter`
- Games List: `Ctrl+G`

For the compatibility tools, the `PupguiInstallDialog` could have a method to set the selected compatibility tool. We could then pass an optional parameter to `MainWindow#btn_add_version_clicked` which would specify the tool. A lambda something like this would  then be attached to the shortcut:

```python
# MainWindow
def setup_ui(self):
    # ...
    QShortcut(QKeySequence('Ctrl+Alt+T'), self.ui).activated.connect(lambda: self.btn_add_version_clicked(compat_tool='Proton-tkg'))
    # ...
```

<hr>

Aside from having to modify some of the install dialog logic, the cost to implement the majority of these shortcuts seems quite low (one line each!), which is always a good thing :smile:

### DIFF
--- a/pupgui2/pupgui2.py
+++ b/pupgui2/pupgui2.py
@@ -152,6 +152,12 @@ class MainWindow(QObject):
         QShortcut(QKeySequence('Alt+Return'), self.ui).activated.connect(self.btn_show_ct_info_clicked)  # Uses 'Return' even though docs mention 'Enter' - https://doc.qt.io/qt-6/qkeysequence.html
         QShortcut(QKeySequence('Ctrl+G'), self.ui).activated.connect(self.btn_show_game_list_clicked)
 
+        # TODO These are currently Steam-specific, need to be able to handle other launchers
+        QShortcut(QKeySequence('Ctrl+Shift+T'), self.ui).activated.connect(lambda: self.btn_add_version_clicked(compat_tool='Proton Tkg'))
+        QShortcut(QKeySequence('Ctrl+Shift+B'), self.ui).activated.connect(lambda: self.btn_add_version_clicked(compat_tool='Boxtron'))
+        QShortcut(QKeySequence('Ctrl+Shift+L'), self.ui).activated.connect(lambda: self.btn_add_version_clicked(compat_tool='Luxtorpeda'))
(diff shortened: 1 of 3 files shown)


sonic2kk:
Added an initial implementation of having shortcuts to open the dialog for a given compatibility tool.

Currently they are all Steam compat tools, and if you have say Lutris selected, these shortcuts will still open the versions dialog. This isn't ideal of course. Probably these shortcuts could be moved into a `setup_shortcuts` method for `MainWindow`, which will check the launcher before binding the shortcuts. Though I'm not sure how this would work, since once the shortcuts are created, I don't know if they're ever "unbound". Dynamic shortcuts like these may not really be an "intended" use-case.

It's something I'll keep looking into :slightly_smiling_face:

sonic2kk:
Thought about this a bit, got a coffee, and came up with something, but it still isn't perfect.

I added a boolean check to the `PupguiInstallDialog#set_selected_compat_tool` dialog method, and then we just won't show the dialog if the check returns False. However this is not great because it still instantiates the dialog, and as a side effect, counts towards the rate limit. So you could spam the shortcut and still hit the rate limit.

Probably it's better to have a helper/utility method that can check the ctobjs that the dialog would have, and have that method outside of the dialog. Then we can have a separate method that checks this before we try to instantiate the dialog.

sonic2kk:
Created the helper method as described and put it in `util.py`. Now we only create the dialog if we can find the name passed in among the list of available ctmods in the ct_loader for the given launcher install location (if the given compat_tool name is truthy).

I think this is a cleaner approach than the one I had in https://github.com/DavidoTek/ProtonUp-Qt/pull/209/commits/649a51a30692fcd7856e22110c95541803c6aeae, but there may yet still be room for improvement :-)

PR #211, opened 2023-03-21
Games List: Use cached list for Steam

sonic2kk:
This PR enables caching when opening the Steam Games List to give a major speed boost, reducing the time it takes to open from a couple of seconds to a fraction of a second. 

## Overview
In the games list, the `update_game_list_steam` method makes two calls: One to `get_steam_game_list`, and then another to `get_steam_ctool_list`. These both internally make calls to `get_steam_app_list`. This has been massively sped up since #171 (seriously, 2.7.7 is painfully slow in comparison, that fix rocks!), but it can still take around 2 seconds. Making the call twice does add a noticeable delay for opening the Steam Games List.

The `get_steam_app_list` method has an optional `cached` parameter, which returns the cached games list. From what I can understand this cache is generated based on the last call to the method with `caching=False`. The `get_steam_game_list` and `get_steam_ctool_list` methods can also take a `cached` parameter which gets passed to their calls to `get_steam_app_list`. Enforcing caching on these calls *massively* speeds up the Games List opening, to the point of being pretty much instantaneous!

On main, the `update_game_list_steam` method takes roughly 3.1s to execute on my PC with my 308 installed games. With this PR that goes down to 0.0002s - It might as well be instantaneous, since all it's doing is returning the global list that was cached previously.

A small concern here is that if a user downloads a game while ProtonUp-Qt is open, and then opens the games list, then it won't update to show that new game. However, I believe this to be an edge case, and the massive speed boost of enabling caching to far outweigh this case for a couple of reasons:
1. This is probably not going to happen very often, probably it's *slightly* more likely to happen on Steam Deck Desktop Mode as Steam opens automatically when switching to Desktop Mode, and so downloads resume. However, I don't think many users would set a game to download, then open ProtonUp-Qt and then open the games list to tinker with the game they *just* finished downloading while ProtonUp-Qt is opened.
2. When Steam is opened, the user is warned that they will need to restart Steam for compatibility tool changes to take effect, therefore it would follow that they should have Steam closed before opening the games list to begin with 

This PR also does a minor refactor for `self.games` in the Games List, setting the type to be a list which can have either `SteamApp`,  `LutrisGame` or `HeroicGame`, because I happened to notice it while tinkering around here.

## Alternative Solution
I couldn't see any disastrous consequences from enabling this. Updating compatibility tools still worked fine, there were no missing games, and the Steam games list opens so much faster. However if there was a reason caching was never set here, this PR could be closed and a potential alternative solution would be to only call out to `get_steam_app_list` once.

We could make a single call to `get_steam_app_list` in the Games Dialog method, then `get_steam_game_list` and `get_steam_ctool_list` methods could take an optional app list parameter. This kind of refactor was not implemented here as the speed boost already reduces the time so much that it I didn't think it would be necessary. But if enabling caching is not an option, this is another potential approach which could probably around half the time it takes to open the games list.

<hr>

Thanks :-)

### DIFF
--- a/pupgui2/pupgui2gamelistdialog.py
+++ b/pupgui2/pupgui2gamelistdialog.py
@@ -1,7 +1,7 @@
 import os
 import pkgutil
 
-from typing import List, Callable, Tuple
+from typing import List, Callable, Tuple, Union
 from datetime import datetime
 
 from PySide6.QtCore import QObject, Signal, Slot, QDataStream, QByteArray, Qt
@@ -30,7 +30,7 @@ class PupguiGameListDialog(QObject):
         self.install_dir = install_dir
         self.parent = parent
         self.queued_changes = {}
-        self.games = []
+        self.games: List[Union[SteamApp, LutrisGame, HeroicGame]] = []
 
         self.install_loc = get_install_location_from_directory_name(install_dir)
         self.launcher = self.install_loc.get('launcher', '')
@@ -103,9 +103,9 @@ class PupguiGameListDialog(QObject):
 
     def update_game_list_steam(self):
         """ update the game list for the Steam launcher """
-        self.games = get_steam_game_list(steam_config_folder=self.install_loc.get('vdf_dir'))
+        self.games = get_steam_game_list(steam_config_folder=self.install_loc.get('vdf_dir'), cached=True)
         ctools = [c if c != 'SteamTinkerLaunch' else 'Proton-stl' for c in sort_compatibility_tool_names(list_installed_ctools(self.install_dir, without_version=True), reverse=True)]
-        ctools.extend(t.ctool_name for t in get_steam_ctool_list(steam_config_folder=self.install_loc.get('vdf_dir')))
+        ctools.extend(t.ctool_name for t in get_steam_ctool_list(steam_config_folder=self.install_loc.get('vdf_dir'), cached=True))
 
         self.ui.tableGames.setRowCount(len(self.games))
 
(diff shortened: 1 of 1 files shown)


DavidoTek:
> In the games list, the update_game_list_steam method makes two calls: One to get_steam_game_list, and then another to get_steam_ctool_list. 

Yeah, somehow I missed that.
`get_steam_app_list` with `caching=False` is called in `pupgui2.py#update_ui` anyway before opening the game list and when clicking the apply button.

> seriously, 2.7.7 is painfully slow in comparison, that fix rocks!

I think it's time for a new release. Probably at least `2.8` with that many additions!!

> roughly 3.1s to execute on my PC with my 308 installed games. With this PR that goes down to 0.0002s 

Wow, that's great.

> Steam Deck Desktop Mode as Steam opens automatically when switching to Desktop Mode, and so downloads resume
> Alternative Solution

Calling `get_steam_app_list` twice is definitively overkill, though I'm not sure if it would make sense to update the cache when opening the game list.
I think it's the obvious thing to do, to restart ProtonUp-Qt if a game isn't displayed properly.
It also doesn't make sense to change something in the game list dialog while Steam is running, as the changed won't be saved properly anyway. I'm wondering if this has caused any issues on Steam Deck as, at least on my desktop machine, I had to close Steam before applying any changes.
I'm open on this and think we can give it a try as reducing the load times seems more useful to most users than some weird edge case that may never even occur. If it does, we can revert later (or add a refresh button :smile:).

PR #212, opened 2023-03-21
Use UI file for Install Dialog

sonic2kk:
## Overview
This PR converts the install dialog to use a Qt Designer `.ui` file. This is more consistent with the other dialogs that ProtonUp-Qt uses, and allows for some cleanup of the install dialog class.

This also has an indirect benefit: it works around a PySide6 bug with Plasma Wayland, where dialogs on scaled displays don't render correctly until moved to a non-scaled display and back again.

image

I attempted to keep the UI as close as possible to the existing install dialog but I did have to increase the size of the dialog just a small amount. On main, the dialog and its elements extend it to a size of `314x279`. However in Qt Designer, with the way it aligns elements, this meant that everything was slightly off-center; elements on the right had slightly more spacing than those on the left.

To allow for even spacing, the dialog dimensions were increased to `320x280`, as round figures that would divide evenly. I felt that a width of `310` was a bit too narrow and I felt giving the description box more room rather than less was a better compromise, so I rounded that size upward.

I hope the size change isn't too noticeable. I thought increasing the size was the better tradeoff than having the UI be slightly off-center. I'll provide a screenshot of main to compare with.

<details>
<summary>CtInstall dialog on Main</summary>

image
</details>

<hr>

Another change I made was to disable the install dialog comboboxes and install button if no tags were found for the given tool, i.e. if GitHub is down, or the user is rate-limited.

I tried my best to rate-limit myself to get a screenshot of this, but *of course* I can't do it when I need to showcase something :sweat_smile: So please be sure to test this as well. I implemented this by changing the logic to not emitting `is_fetching_releases(False)` if the tags list is empty. This is a little bit hacky I suppose but I guess it also still follows logically, as if the user is rate limited or if GitHub is down, it never really *fetches* the releases.

<hr>

There were also some minor code cleanups done along the way that I noticed, like removing some unused variables, and replacing for loops to add items with `addItems` and a list comprehension to help streamline the code a little more. The logic for loading, setting up and showing the UI for the installer dialog was also moved into its `__init__` instead of in the `pupgui2` file where its called, which is again more consistent with other parts of the code. 

I felt this would be a good change for consistency. It could be done for other dialogs to but the CtInstaller dialog was the last "major" one as far as I could see that didn't use a UI file.

Thanks :-)

### DIFF
--- a/pupgui2/pupgui2.py
+++ b/pupgui2/pupgui2.py
@@ -317,11 +317,7 @@ class MainWindow(QObject):
             dialog = PupguiInstallDialog(install_loc, self.ct_loader, parent=self.ui)
             dialog.compat_tool_selected.connect(self.install_compat_tool)
             dialog.is_fetching_releases.connect(self.set_fetching_releases)
-            dialog.setup_ui()
             dialog.set_selected_compat_tool(compat_tool)
-            dialog.show()
-            dialog.setFixedSize(dialog.size())
-
 
     def btn_remove_selcted_clicked(self):
         ctools_to_remove = []
--- a/pupgui2/pupgui2.py
+++ b/pupgui2/pupgui2.py
@@ -237,7 +237,6 @@ class MainWindow(QObject):
             if ct.no_games == 0:
                 unused_ctools += 1
 
-            
         self.ui.txtActiveDownloads.setText(str(len(self.pending_downloads)))
         if len(self.pending_downloads) == 0:
             self.set_default_statusbar()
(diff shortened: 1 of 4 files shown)


DavidoTek:
> This also has an indirect benefit: it works around a PySide6 bug with Plasma Wayland, where dialogs on scaled displays don't render correctly until moved to a non-scaled display and back again.

Interesting... why would that be different with ui files?

> I attempted to keep the UI as close as possible to the existing install dialog but I did have to increase the size of the dialog just a small amount
> I hope the size change isn't too noticeable

That should mean that they are now consistent with the other dialogs as they are using the default spacing of Qt Designer :)
I don't think it will break anything.

> Another change I made was to disable the install dialog comboboxes and install button if no tags were found for the given tool, i.e. if GitHub is down, or the user is rate-limited

Good.

> This is a little bit hacky I suppose but I guess it also still follows logically, as if the user is rate limited or if GitHub is down, it never really fetches the releases.

Makes sense. I will do some further testing, but code looks good.

---

Thanks a lot!

PR #213, opened 2023-03-21
Heroic: Type Hint cleanup

sonic2kk:
I made a pretty bad mistake when implementing the Heroic util stuff. For some reason I tried to type hint assigning the values to the HeroicGame object. I recently re-configured my vscodium install to get my development environment in order and this was something it yelled at me about this :sweat_smile: 

This also removes what appears to be an unused element, `hg.install`, which was not part of the HeroicGame data structure - I think this was discussed and removed from the data structure, but it seems I missed heroicutil.

As far as I know this shouldn't have caused any problems, but it's invalid usage of type hints, so my bad there.

### DIFF
--- a/pupgui2/heroicutil.py
+++ b/pupgui2/heroicutil.py
@@ -29,27 +29,26 @@ def get_heroic_game_list(heroic_path: str) -> List[HeroicGame]:
     for game in games_json:
         hg = HeroicGame()
 
-        hg.runner: str = game.get('runner', '')
-        hg.app_name: str = game.get('app_name', '')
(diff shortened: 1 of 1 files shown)


DavidoTek:
Yeah, I noticed that all the types were hinted.
My first thought was that it is for better readability though vscode/ium/... should show the types when hovering over them.

> As far as I know this shouldn't have caused any problems, but it's invalid usage of type hints, so my bad there.

I don't think that's how they are supposed to be used but Python doesn't hinder anyone from using them like that :)

Thanks.

PR #215, opened 2023-03-22
Add a Refresh Button to the Game List and Compat Tool Info dialog

DavidoTek:
With https://github.com/DavidoTek/ProtonUp-Qt/pull/211 and https://github.com/DavidoTek/ProtonUp-Qt/pull/214, caching of games has been enabled to increase the load times of the dialogs. This comes with the limitation that external changes are not registered, even when re-opening the game list or ctinfo dialogs.

This PR adds a refresh button to both the game list and the ctinfo dialog that updates the cached games and rebuilds the list of displayed games.

### DIFF
--- a/pupgui2/pupgui2ctinfodialog.py
+++ b/pupgui2/pupgui2ctinfodialog.py
@@ -48,9 +48,22 @@ class PupguiCtInfoDialog(QObject):
         self.ui.btnSearch.setVisible(False)
         self.ui.searchBox.setVisible(False)
 
+        self.update_game_list()
+
+        self.ui.btnSearch.clicked.connect(self.btn_search_clicked)
+        self.ui.btnRefreshGames.clicked.connect(self.btn_refresh_games_clicked)
+        self.ui.btnClose.clicked.connect(lambda: self.ui.close())
+        self.ui.listGames.cellDoubleClicked.connect(self.list_games_cell_double_clicked)
+        self.ui.searchBox.textChanged.connect(self.search_ctinfo_games)
+
+        if self.ui.listGames.rowCount() > 0:
+            self.ui.btnSearch.setVisible(True)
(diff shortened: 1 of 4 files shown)


sonic2kk:
Just being nosy here: This uses a QToolButton on the Games List dialog but a QPushButton on the Info dialog (the search button is also a QPushButton on the info dialog). Is that intentional? The QToolButton has a slightly different size than the QPushButton, and is more square, which might be intentional. Plus, "tool button" sounds more logical for this sort of action, I probably should've used that in #210 but it didn't even cross my mind :sweat_smile: 

QToolButton:
image

QPushButton:
image

For the QPushButton change I had to set the focusPolicy in QtDesigner to `NoFocus`, otherwise the refresh button was the default selected widget.

Absolutely *not* a big deal, I am just curious :-)

DavidoTek:
> Just being nosy 

Quite the reverse, good hint!

I noticed the different but wanted to use a Tool Button because as far as I know it is used for this kind of stuff (it has some different implementation with QAction, but more important, is designed for an icon instead of text).
On the other hand, it makes the refresh button to look the same as the search button in the ct info dialog and I didn't think of changing the other button too :)

I probably should change the type, so all icon buttons use QToolButton.

DavidoTek:
> For the QPushButton change I had to set the focusPolicy in QtDesigner to NoFocus

One thing I haven't considered when merging the last PRs is that setting the focusPolicy to NoFocus will prevent controller support from working properly as the [D-Pad simulates tab presses](https://github.com/DavidoTek/ProtonUp-Qt/blob/c51e3552ee9f009789eda2fea62719a7d6cc3d6f/pupgui2/gamepadinputworker.py#L21)...

> otherwise the refresh button was the default selected widget.

I think that can be prevented by changing the tab order.

---

I will look into that and provide a PR.

Issue #196, opened 2023-02-23
Add support for CryoByte's Steam Deck Utilities


DavidoTek:
**Is your feature request related to a problem? Please describe.**
@CryoByte33's [Steam Deck Utilities/CryoUtilities](https://github.com/CryoByte33/steam-deck-utilities) is "a utility to improve performance and help manage storage on Steam Deck".
Currently, it can be installed by downloading a `.desktop` file and double-clicking it.

As many Steam Deck users also use ProtonUp-Qt, it might be worth adding an installer for the CryoUtilities.

**Describe the solution you'd like**
A new dialog "Steam Deck Tools" could be added to ProtonUp-Qt that allows installing different tools for Steam Deck that are not directly compatibility tools.
The dialog could offer tools like CryoUtilities, Boot Video Changer, maybe glibc-eac for Arch (https://github.com/DavidoTek/ProtonUp-Qt/issues/195) etc.

**Describe alternatives you've considered**
Not adding an installer. Is that a real alternative? :smile:

**Additional context**
- Official CryoUtilities install script: https://github.com/CryoByte33/steam-deck-utilities/blob/main/install.sh
- Discussion on CryoUtilities install location: https://github.com/CryoByte33/steam-deck-utilities/issues/45
- Ideally this would be kept a bit separate of the other ProtonUp-Qt stuff, so an independent sub-module "pupgui2.decktools" should be created.

CryoByte33:
Let me know if you need anything from my side, thank you for considering this!

PR #225, opened 2023-04-08
Custom Install Directory Dialog Overhaul

sonic2kk:
I had some time so I decided to take a look at overhauling the custom install directory dialog, making several small tweaks that turned out to be quite an overhaul. This is not yet complete but I made enough progress that I'm ready to open it and get some feedback :-)

image

## Overview
This PR overhauls the Custom Install Directory dialog in a number of ways:
- Convert the dialog to use a `.ui` file - This doesn't reduce the size of the file in the end but allows for consistency with the other dialogs.
- Add a short description on the purpose of this dialog - Phrasing improvements welcome! 
- Add a "Close" button, which allows for the dialog to be closed when running in GameScope (i.e. on Steam Deck).
- Add a "Default" button, which takes the place of the "Reset" functionality previously coupled to the "Save" button.
    - This emits `custom_id_set` when clicked, so the launcher dropdown on the main menu will update.
- Add some logic to the "Save" button which only enables it if a valid directory path is selected *and* if we can write to this directory.
    - This is only checked when the path is being entered, unless there is an existing check for custom install directories, there is no logic added in this PR (currently anyway) which validates that a given custom install directory still exists on boot of ProtonUp-Qt. So if a path disappeared I am not sure what would happen.
- Rework the launcher names in the dropdown, so instead of using something like `heroicwine`, it now says `Heroic (Wine)` - Very minor change I wanted to make :sweat_smile: 
- Select the current launcher in the combobox - So if the dialog is opened when we're on Lutris, it will select "Lutris".

## Implementation
I have a couple of reservations about the way I went about some things in this PR specifically that I wanted to comment on.
- `set_selected_launcher` is identical to the logic in the install dialog for checking the exact same thing -- Finding the index of a dropdown element based on the text and selecting that item. Maybe this could be broken out into a utility function?
- I created a method `get_dict_key_from_value`, which is used to get the "internal" name key (i.e. `heroicwine`) from the display text (i.e. `Heroic (Wine)`). This feels... not-pythonic at best. If there is a better way to go about this please do correct me :-) 
- There's an explicit call to emit the index changed signal when removing the selected custom install directory. This is because when the custom dir is selected in the main menu and then defaulted in the dialog, the launcher combobox value on the main menu updates (defaults to the first element) but it doesn't emit the signal. So when we send an empty string to the custom install directory signal (i.e. when the default button is pressed) we check for that and manually emit the index changed signal. I'd be happy to take any feedback to improve this! 

And of course feedback is welcome on the rest of the code, those are just items I specifically wanted to call out.

## Considerations
I don't have Bottles installed so I haven't tested any of these changes against it. However, for Steam (`steam` package from Arch repos), Lutris (`lutris-git` from AUR), and Heroic (Flatpak), things seem to work as expected.

When selecting "Bottles" and adding a custom install directory through, I noticed it just adds the path and does not set a display name. This happens in main as well. For example adding `/home/gaben/PressurizedBottles/`, the dropdown in the main menu only displays the path and not the launcher display name, nor any icon. This isn't something that broke in this PR but it is something that could be fixed here or in future.

<hr>

I'm not sure how often this feature is used (to be honest I haven't actually used it...), but I wanted to try my hand at improving it, so all feedback to better reach that end goal is welcome :-)

Thanks!

P.S. - Looks like the combined downloads and Flathub installs has surpassed 1 million, congrats! :-)

<hr>

## TODO
- [ ] Potentially improve how the custom install directory is written out (icon, display name)
- [x] Select the new directory in the main menu on "Save" 
- [x] Refresh compat tool list when removing currently selected custom dialog (dropdown value updates but compat tool list does not)

### DIFF
--- a/pupgui2/pupgui2.py
+++ b/pupgui2/pupgui2.py
@@ -204,9 +204,8 @@ class MainWindow(QObject):
                 self.ui.comboInstallLocation.setCurrentIndex(i)
 
         self.updating_combo_install_location = False
-        # Update compat list when custom install directoy is removed
+        # Update compat list when custom install directoy is removed -- Not called because of `self.updating_combo_install_location = False` -- Could be improved?
         if custom_install_dir is not None and len(custom_install_dir) <= 0:
-            # currentIndexChanged is not emitted automatically?
             self.ui.comboInstallLocation.currentIndexChanged.emit(self.ui.comboInstallLocation.currentIndex())
 
     def update_ui(self):
--- a/pupgui2/pupgui2.py
+++ b/pupgui2/pupgui2.py
@@ -183,13 +183,13 @@ class MainWindow(QObject):
         t = threading.Thread(target=_set_default_statusbar_thread, args=[self.update_statusbar_message])
         t.start()
 
-    def update_combo_install_location(self):
+    def update_combo_install_location(self, custom_install_dir = None):
         self.updating_combo_install_location = True
 
         self.ui.comboInstallLocation.clear()
         self.combo_install_location_index_map = []
 
-        current_install_dir = install_directory()
+        current_install_dir = custom_install_dir or install_directory()
         for i, install_dir in enumerate(available_install_directories()):
             icon_name = get_install_location_from_directory_name(install_dir).get('icon')
             display_name = get_install_location_from_directory_name(install_dir).get('display_name')
@@ -199,6 +199,8 @@ class MainWindow(QObject):
                 self.ui.comboInstallLocation.addItem(install_dir)
             self.combo_install_location_index_map.append(install_dir)
             if current_install_dir == install_dir:
+                if custom_install_dir is not None:
+                    self.updating_combo_install_location = False
                 self.ui.comboInstallLocation.setCurrentIndex(i)
 
         self.updating_combo_install_location = False
--- a/pupgui2/pupgui2.py
+++ b/pupgui2/pupgui2.py
@@ -204,6 +204,10 @@ class MainWindow(QObject):
                 self.ui.comboInstallLocation.setCurrentIndex(i)
 
         self.updating_combo_install_location = False
+        # Update compat list when custom install directoy is removed
+        if custom_install_dir is not None and len(custom_install_dir) <= 0:
+            # currentIndexChanged is not emitted automatically?
+            self.ui.comboInstallLocation.currentIndexChanged.emit(self.ui.comboInstallLocation.currentIndex())
 
     def update_ui(self):
         """ update ui contents """
--- a/pupgui2/pupgui2.py
+++ b/pupgui2/pupgui2.py
@@ -361,7 +361,7 @@ class MainWindow(QObject):
             self.update_ui()
 
     def btn_manage_install_locations_clicked(self):
-        customid_dialog = PupguiCustomInstallDirectoryDialog(parent=self.ui)
+        customid_dialog = PupguiCustomInstallDirectoryDialog(install_directory(), parent=self.ui)
         customid_dialog.custom_id_set.connect(self.update_combo_install_location)
 
     def show_launcher_specific_information(self):
--- a/pupgui2/pupgui2customiddialog.py
+++ b/pupgui2/pupgui2customiddialog.py
@@ -6,15 +6,26 @@ from PySide6.QtGui import QIcon
 from PySide6.QtWidgets import QFileDialog, QLabel, QPushButton, QLineEdit, QComboBox, QFormLayout
 from PySide6.QtUiTools import QUiLoader
 
-from pupgui2.util import config_custom_install_location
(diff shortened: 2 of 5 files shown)


sonic2kk:
There are a couple of areas that I noticed could be improved while working on this PR, but which _may_ not really fit this PR:
- The remove logic in `install_directory` could be split out into a separate function. It appears this logic is only used by the custom install directory dialog
- A base class dialog could be created, since each dialog now uses a UI file. This would take the path to the UI file and automatically call `load_ui` based on the given path to the UI file and such. This would reduce the duplicated code between dialog and streamline creating new dialogs, which could end up being useful for #196 if any additional dialogs are required for this :-)

sonic2kk:
I broke the logic for finding the combobox text in `set_selected_launcher` out into a util function, and applied it for this custom install dialog as well as the install dialog. I felt it would be cleaner to go ahead and do so :-)

DavidoTek:
Great!

Having three buttons, instead of just one seems like a good idea :)

> which validates that a given custom install directory still exists on boot of ProtonUp-Qt. So if a path disappeared I am not sure what would happen.

I think it will just not display anything. Not much of a problem.

> I created a method get_dict_key_from_value, which is used to get the "internal" name key (i.e. heroicwine) from the display text (i.e. Heroic (Wine)). This feels... not-pythonic at best. If there is a better way to go about this please do correct me :-)

There are more-pythonic ways, see e.g. https://www.geeksforgeeks.org/python-get-key-from-value-in-dictionary/
But I think the `get_dict_key_from_value` method works fine and is easy to comprehend while at least some pythonic methods are quite messy in my opinion.

> There's an explicit call to emit the index changed signal when removing the selected custom install directory

Intersting. Might be caused by https://github.com/DavidoTek/ProtonUp-Qt/blob/3c81fc754604619f7e7a759ba3da2f5a5941e488/pupgui2/pupgui2.py#L363-L364

> When selecting "Bottles" and adding a custom install directory through, I noticed it just adds the path and does not set a display name

You're right. Currently, it just ignores the display name and launcher. See https://github.com/DavidoTek/ProtonUp-Qt/blob/3c81fc754604619f7e7a759ba3da2f5a5941e488/pupgui2/util.py#L200

> The remove logic in install_directory could be split out into a separate function. It appears this logic is only used by the custom install directory dialog

Not sure, do you mean calling `config_custom_install_location(install_dir='remove')`?

> A base class dialog could be created, since each dialog now uses a UI file [...] which could end up being useful for Add support for `CryoByte's Steam Deck Utilities #196`

Yeah, that could be a useful addition to make development easier.

--------------

Thanks!

> P.S. - Looks like the combined downloads and Flathub installs has surpassed 1 million, congrats! :-)

Nice, I have seen that too. I think we're doing a great job of improving user experience for Linux and Steam Deck gaming! :tada: :tada: :tada:

sonic2kk:
>[...] some pythonic methods are quite messy in my opinion.

Yeah, some of the ways in that link are shorter for sure but not quite as clean.

> Intersting. Might be caused by <code>

I was messing around here but I just tried removing the check in `combo_install_location_current_index_changed` and yup, now the method is called properly.

Just out of curiosity, what does the `updating_combo_install_location` do? Just so I can get a better idea of how to potentially fix the logic so that we don't have to make an explicit call to the combobox signal :slightly_smiling_face: 

I wonder if it might be worth creating another signal for custom install directory removed. Not sure for now.

> You're right. Currently, it just ignores the display name and launcher

Ah good point, and also here (sorry, I don't know how to do the actual line embed :sweat:)

https://github.com/DavidoTek/ProtonUp-Qt/blob/main/pupgui2/util.py#L271

`config_custom_install_location` also notes that `display_name` is always `''`. That could be changed in this PR, or at a later date, or not at all if it's going to be a breaking change :sweat_smile: 

> Not sure, do you mean calling `config_custom_install_location(install_dir='remove')`?

Yeah, I was just thinking that it might be slightly cleaner to have it as a separate function, or perhaps now that I think about it, it could be a separate argument. Of course it isn't a massive issue, it was just a small observation :smile: 

<hr>

I pushed a couple of small changes related to fixing some incorrect behaviour with and when using `get_dict_key_from_value`, and some small comment cleanup.

DavidoTek:
> Just out of curiosity, what does the updating_combo_install_location do?

It will clear the install location combo box in the main dialog and re-populate it with all available install locations. It is called from custom install dialog so the new launcher gets added.

> I wonder if it might be worth creating another signal for custom install directory removed.

`updating_combo_install_location` should handle that fine.

> I don't know how to do the actual line embed sweat

Pasting a permalink to a line should do that automatically. Probably it is not working here because it's a normal link, and not a permalink.

> it might be slightly cleaner to have it as a separate function, or perhaps now that I think about it, it could be a separate argument

Yeah, a separate argument would be a clean option.

---

One thing I noticed is that the dialog doesn't show the current custom path when opened (but neither did the old version of the dialog). I'll take a look and add it quick.

---

Thanks

Issue #226, opened 2023-04-10
Protonup-qt crashes immediately (2.8.0 regression)


cylgalad:
Please fill out the following when reporting a new bug:

**Describe the bug**  
Immediate crash at launch.

**To Reproduce**  
Launch protonup-qt

**Expected behavior**  
Protonup-qt window.

**Desktop (please complete the following information):**  
 - **Platform**: PC
 - **System**: Archlinux
 - **Version**: ProtonUp-Qt 2.8.0
 - **How did you install ProtonUp-Qt?**: yay -S protonup-qt

**Additional context**  
Maybe tied to https://github.com/DavidoTek/ProtonUp-Qt/issues/202#issue-1620272605
Commenting out lines 73-74 in steamutil.py fixes the issue, but I still don't have any Steam games in the list.

**Terminal output**  
```
ProtonUp-Qt 2.8.0 by DavidoTek. Build Info: built from source.
Python 3.10.10 (main, Mar  5 2023, 22:26:53) [GCC 12.2.1 20230201], PySide 6.4.3
Platform: Arch rolling Linux-6.2.10-arch1-1-x86_64-with-glibc2.37
Loading locale fr / fr_FR
Loaded ctmod GE-Proton
Loaded ctmod Wine-GE
Loaded ctmod Boxtron
Loaded ctmod D8VK (nightly)
Loaded ctmod Kron4ek Wine-Builds Vanilla
Loaded ctmod Lutris-Wine
Loaded ctmod Luxtorpeda
Loaded ctmod NorthStar Proton (TitanFall 2)
Loaded ctmod Proton Tkg
Loaded ctmod Proton Tkg (Wine Master)
Loaded ctmod Roberta
Loaded ctmod Steam-Play-None
Loaded ctmod SteamTinkerLaunch
Loaded ctmod SteamTinkerLaunch-git
Loaded ctmod vkd3d-lutris
Loaded ctmod vkd3d-proton
Loaded ctmod Wine Tkg (Valve Wine)
Loaded ctmod Wine Tkg (Vanilla Wine)
Loaded ctmod DXVK
Loaded ctmod DXVK Async
Loaded ctmod DXVK (nightly)
qt.pysideplugin: Environment variable PYSIDE_DESIGNER_PLUGINS is not set, bailing out.
qt.pysideplugin: No instance of QPyDesignerCustomWidgetCollection was found.
[...]

PR #227, opened 2023-04-10
steamutil: Move shortcuts check inside try block, add logging for VDF file loading

sonic2kk:
Related to #226.

This moves the `if not no_shortcuts` check inside the `try` block to prevent a crash if `c` is not assigned. If something inside the `try` block fails early on, the libraryfolders VDF (`v`) or config VDF (`c`) variables may not be set appropriately and so a crash will occur when trying to reference `c` before it is assigned a value - If `v` throws an exception then `c` will never be set, or `c` is not loaded properly it could also be `None`. If the VDF library can read it though it should always return an empty dict, but if any of the `get`s fail it could end up being `None` I think (not sure, never had an exception get thrown with chained `get`s, not sure if it will just be `None`, take on the last return value, or if it will be unassigned thus triggering the crash).

This also adds some logging for the VDF file loading. Since it's come up a couple of times now that these files aren't parsed correctly it would be useful to see which file can't be loaded. This log message will get triggered each time the user switches to the Steam utility but for most users this should be invisible, it might just end up being "noisy" for those of us that run from the command line often. So perhaps the logging is less desirable but I can remove it. If it is desired I figured it would be better to catch that in this PR too.

<hr>

The root cause of #226 is still likely that a Steam file is corrupted or otherwise invalid, but this should prevent the crash in such a scenario (the Steam games list will probably still be blank though).

### DIFF
--- a/pupgui2/steamutil.py
+++ b/pupgui2/steamutil.py
@@ -40,8 +40,10 @@ def get_steam_app_list(steam_config_folder: str, cached=False, no_shortcuts=Fals
     
     try:
         v = vdf.load(open(libraryfolders_vdf_file))
+        print(f'Successfully loaded {libraryfolders_vdf_file}.')
         c = vdf.load(open(config_vdf_file)).get('InstallConfigStore').get('Software').get('Valve').get('Steam').get('CompatToolMapping')
-        
+        print(f'Successfully loaded {config_vdf_file}.')
+
         for fid in v.get('libraryfolders'):
             if 'apps' not in v.get('libraryfolders').get(fid):
                 continue
@@ -67,11 +69,11 @@ def get_steam_app_list(steam_config_folder: str, cached=False, no_shortcuts=Fals
                 apps.append(app)
         apps = update_steamapp_info(steam_config_folder, apps)
         apps = update_steamapp_awacystatus(apps)
+
+        if not no_shortcuts:
+            apps.extend(get_steam_shortcuts_list(steam_config_folder, c))
     except Exception as e:
         print('Error: Could not get a list of all Steam apps:', e)
-
-    if not no_shortcuts:
-        apps.extend(get_steam_shortcuts_list(steam_config_folder, c))
 
     _cached_app_list = apps
     return apps
--- a/pupgui2/steamutil.py
+++ b/pupgui2/steamutil.py
@@ -337,7 +337,8 @@ def steam_update_ctool(game: SteamApp, new_ctool=None, steam_config_folder='') -
     game_id = game.app_id
 
     try:
-        c = get_steam_vdf_compat_tool_mapping(vdf.load(open(config_vdf_file)))
+        d = vdf.load(open(config_vdf_file))
+        c = get_steam_vdf_compat_tool_mapping(d)
 
         if str(game_id) in c:
             if new_ctool is None:
@@ -365,7 +366,8 @@ def steam_update_ctools(games: Dict[SteamApp, str], steam_config_folder='') -> b
         return False
 
     try:
-        c = get_steam_vdf_compat_tool_mapping(vdf.load(open(config_vdf_file)))
+        d = vdf.load(open(config_vdf_file))
+        c = get_steam_vdf_compat_tool_mapping(d)
 
         for game, new_ctool in games.items():
             game_id = game.app_id
--- a/pupgui2/steamutil.py
+++ b/pupgui2/steamutil.py
@@ -103,7 +103,7 @@ def get_steam_shortcuts_list(steam_config_folder: str, compat_tools: dict=None) 
 
     try:
         if not compat_tools:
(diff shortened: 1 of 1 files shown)


DavidoTek:
Thanks and also thanks for doing the investigation for # 226! :tada:

> This moves the if not no_shortcuts check inside the try block

I wonder if we should put the `if not no_shortcuts: ...` inside the `else:` part of the try-except. I think it is the correct way of doing it in Python:

```python
try:
    c = ...
except:
    ...
else:
    if not no_shortcuts:
        ...
```

(PS: I noticed I've accidentally used `finally` instead of `else` in some places. I will correct that :) )

> Since it's come up a couple of times now that these files aren't parsed correctly it would be useful to see which file can't be loaded

Can we change it to print a message only if loading fails?

sonic2kk:
Moved the shortcuts check into the `else` block - Just looked it up and I agree, didn't even realise Python *had* this syntax :sweat_smile: 

> Can we change it to print a message only if loading fails?

So I thought about this and realised a couple of things:
- To do this we'd probably have to initialize `v` and `c` to `None` so we can check it in the else/exception block, but then that itself would fix the referenced-before-assignment problem, though it still doesn't hint at when `v` or `c` might be `None`. I am not sure how best to go about this, I would rather a solution that avoids initialising them both to `None` 
- `open()` will already log if the file couldn't be opened, see #228 (` No such file or directory: '$HOME/.local/share/Steam/config/libraryfolders.vdf`)

So for now I removed it altogether, but I am open to improvements on where/how to display this :-) (A Python debugger could do it I guess...)

DavidoTek:
> So for now I removed it altogether, but I am open to improvements on where/how to display this :-) 

Okay. If there is a key error, it should print the exception anyway. With the approach below, we should see in which half of the get-chain the error is.

> it may be worth adding .get('Valve', 'valve') 

I don't think that would work as the second parameter is the fallback value. What we could do is:

````python
c1 = vdf.load(open(config_vdf_file)).get('InstallConfigStore').get('Software')
if 'Valve' in c1:
    c = c1.get('Valve').get('Steam').get('CompatToolMapping')
elif 'value' in c1:
    c = c1.get('valve').get('Steam').get('CompatToolMapping')
else:
    raise Exception('Error! config.vdf/InstallConfigStore/Software neither contains key "Valve", nor "valve"')
```

sonic2kk:
> I don't think that would work as the second parameter is the fallback value.

You're right, that was silly of me 😅 

The snippet you posted would work, I wonder if there's a better way just in general, as mentioned in another issue perhaps there's a better way to use `get`. Maybe creating some dedicated class to manage VDF data and giving it a `get` method but that's probably way overkill for simply managing a dict.

We may even just be able to call the variables `c` instead of using `c1` but I'm not sure, didn't test yet :-)

sonic2kk:
Hmm, when I was looking into adding the block mentioned, I noticed that there are a few places where the chained `get`s are used to get the `CompatToolMapping`, which is under the `Valve` key. So I tried to write a generic function to do what we discussed:

```python
def get_steam_vdf_compat_tool_mapping(vdf_file: dict):

    c = vdf_file.get('InstallConfigStore').get('Software')

    # Sometimes the key is 'Valve', sometimes 'valve', see #226
    c = c.get('Valve') or c.get('valve')
    if not c:
        raise KeyError('Error! config.vdf InstallConfigStore.Software neither contains key "Valve" nor "valve" - config.vdf file may be invalid!')

    return c.get('Steam').get('CompatToolMapping')
```

Since `get` does not raise an exception, we can use an `or` here to check for `Valve` or `valve` as a fallback. After this `or`, if `c` is `None`, we can raise the error. If the key does exist, we go through the rest of the `get`s.

I broke the `get`s down in this way because I thought it would be easier to manage, should the format of this file ever change (which I would *guess* it won't drastically change, but I also don't work for the big V :sweat_smile:). It would be possible to change it to `vdf_file.get('InstallConfigStore').get('Software').get('Valve') or vdf_file.get('InstallConfigStore').get('Software').get('valve')` though.

It should be possible to use this function like so:

```
v = vdf.load(open(libraryfolders_vdf_file))
c = get_steam_vdf_compat_tool_mapping(vdf.load(open(config_vdf_file)))
```

I have done some light testing on my PC and behaviour appears to be the same, though further testing is a good idea as this is a fairly significant change to how the `config.vdf` is loaded.

DavidoTek:
> I thought it would be easier to manage, should the format of this file ever change

Yes, that makes it more flexible indeed.

> It would be possible to change it to vdf_file.get('InstallConfigStore').get('Software').get('Valve') or 

I think it is good as you did. That should also make it more clear what the problem is if they should ever change the format again.

> I have done some light testing on my PC and behaviour appears to be the same, though further testing is a good idea as this is a fairly significant change to how the config.vdf is loaded.

I will test it and then merge.

Thanks! :tada:

PR #231, opened 2023-04-15
Core: Check for VDF files before selecting Steam install directory

sonic2kk:
Following discussion in https://github.com/DavidoTek/ProtonUp-Qt/issues/226#issuecomment-1509838832, it appears Steam install folders can exist and have the valid directory structure, but not the actual files that we need.

This PR addresses this by changing the `_STEAM_ROOT` selection check. Now we check that `config.vdf` and `libraryfolders.vdf` exist in the given Steam root dir, as if they do, that implies `_STEAM_ROOT/config` exists as well, so no need to check that as well.


DavidoTek:
> Following discussion in https://github.com/DavidoTek/ProtonUp-Qt/issues/226#issuecomment-1509838832, it appears Steam install folders can exist and have the valid directory structure, but not the actual files that we need.

Sorry I couldn't respond earlier. Definitively good findings!

Thanks!

PR #234, opened 2023-04-28
D8VK: Update branch name

sonic2kk:
[D8VK apparently uses `main` as its default branch now](https://github.com/AlpyneDreams/d8vk/branches), which breaks the download check. Since it skips any workflows that aren't based on the `head_branch` being `master` it was skipping all of them and not fetching any releases.

I'm not sure when the branch name changed but when investigating an implementation of D8VK for Heroic I came across this issue (not in this PR). D8VK seems to work fine now with this change.


DavidoTek:
Thanks, well observed!

PR #263, opened 2023-07-07
Heroic: Add DXVK

sonic2kk:
This adds DXVK for `heroicwine` and `heroicproton`, mirroring the implementation for #245. I tested that this didn't break Lutris, and tested a couple of different DXVK versions (v2.2 and v2.0) for each.

image

DXVK versions are displayed on the main menu without any changes, because #245 added the logic to check for and display DXVK and vkd3d tools. The only change required here was changing the path that DXVK is extracted to.

An equivalent change can be made for VKD3D as well, though that can go in a separate PR that will depend on whether #262 is accepted or not ;-) 

Thanks!


DavidoTek:
Thanks! :tada:

PR #264, opened 2023-07-08
Proton-tkg: Remove unnecessary variables from subclasses

sonic2kk:
Removes some duplicated variables from the Proton-tkg subclasses, like the buffer size and signal. These are already set by the main Proton-tkg class and the values didn't change for the subclasses, so we don't need to set them.

Did a quick spot check and this didn't break anything, Tkgs still downloaded successfully for Lutris, Heroic and Steam.

This is just minor code maintenance, and has no impact on functionality. So no screenshots for this one :-)

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_protontkg_winemaster.py
+++ b/pupgui2/resources/ctmods/ctmod_protontkg_winemaster.py
@@ -2,7 +2,7 @@
 # Proton-Tkg https://github.com/Frogging-Family/wine-tkg-git
 # Copyright (C) 2022 DavidoTek, partially based on AUNaseef's protonup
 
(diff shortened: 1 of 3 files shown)


sonic2kk:
Hm, just had a quick glance on my phone, I think there are some unused imports now (mainly for `Signal`). I can clean those up.

sonic2kk:
Removed unused imports. I left the comments as-is, not sure if we want to update the names to distinguish Proton-tkg/Wine-tkg, or even things like the copyright years.

DavidoTek:
Thanks, I appreciate. That keeps the code clean :)

---

Sorry for the delayed response, there's quite a lot of other stuff to do lately..

sonic2kk:
No worries, real life comes first!

Issue #235, opened 2023-05-01
No game list in Nobara


Martinligabue:
Please fill out following when reporting a new bug:

**Describe the bug**  

**To Reproduce**  
1. Open ProtonU-Qt
2. Click game list on Steam

**Expected behavior**  
you should see the list of games installed in steam

**Screenshots**  

**Desktop (please complete the following information):**  
 - **Platform**: PC
 - **System**: Nobara Linux 37
 - **Version**: ProtonUp-Qt 2.8.0
 - **How did you install ProtonUp-Qt?**: Preinstalled (same as Steam, I'm using the preinstalled version)

**Additional context**  
Appimage works flawlessly

**Terminal output**  
```
ProtonUp-Qt 2.8.0 by DavidoTek. Build Info: built from source.
Python 3.11.3 (main, Apr  5 2023, 00:00:00) [GCC 12.2.1 20221121 (Red Hat 12.2.1-4)], PySide 6.5.0
Platform: Nobara Linux 37 (Thirty Seven) Linux-6.2.12-200.fsync.fc37.x86_64-x86_64-with-glibc2.36
qt.dbus.integration: Could not connect "org.freedesktop.IBus" to globalEngineChanged(QString)
Loading locale it / it_IT
Loaded ctmod GE-Proton
Loaded ctmod Wine-GE
Loaded ctmod Boxtron
Loaded ctmod D8VK (nightly)
Loaded ctmod Kron4ek Wine-Builds Vanilla
Loaded ctmod Lutris-Wine
Loaded ctmod Luxtorpeda
Loaded ctmod NorthStar Proton (TitanFall 2)
Loaded ctmod Proton Tkg
Loaded ctmod Proton Tkg (Wine Master)
Loaded ctmod Roberta
Loaded ctmod Steam-Play-None
cat: /etc/lsb-release: File o directory non esistente
Loaded ctmod SteamTinkerLaunch
cat: /etc/lsb-release: File o directory non esistente
Loaded ctmod SteamTinkerLaunch-git
Loaded ctmod vkd3d-lutris
Loaded ctmod vkd3d-proton
Loaded ctmod Wine Tkg (Valve Wine)
Loaded ctmod Wine Tkg (Vanilla Wine)
Loaded ctmod DXVK
Loaded ctmod DXVK Async
Loaded ctmod DXVK (nightly)
qt.pysideplugin: Environment variable PYSIDE_DESIGNER_PLUGINS is not set, bailing out.
[...]

PR #265, opened 2023-07-09
Bump steam dependency version to match requirements.txt

sonic2kk:
The `install_requires` dependencies for `setup.cfg` mostly match that of `requirements.txt`. The only exception is the `steam` library which is still at `1.1.0`, when really it should be `1.4.4` (as this has the updated magic for VDF parsing, see #157). This PR bumps that requirement version.

This hasn't affected anything and `>= 1.1.0` should probably always use the most up-to-date `steam` dependency, I can't envision a reasonable case when it wouldn't without explicit user intervention. I don't think there's any harm in keeping these in sync, though.

I believe `setup.cfg` is used by setuptools, though I am not exactly sure how this file applies in that context to ProtonUp-Qt.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,7 +25,7 @@ install_requires =
     vdf>=3.4
     inputs==0.5
     pyxdg>=0.27
-    steam>=1.1.0
+    steam>=1.4.4
     PyYAML==6.0
     zstandard>=0.19.0
 


DavidoTek:
That seems like a good idea.

I just looked up, Fedora repos and AUR are also up to 1.4.4, so that wouldn't break anything.
- https://github.com/DavidoTek/ProtonUp-Qt/issues/235#issuecomment-1529706737
- https://aur.archlinux.org/packages/python-steam

PR #268, opened 2023-07-22
Heroic: Add vkd3d

DavidoTek:
Adds vkd3d-lutris and vkd3d-proton for heroicwine/heroicproton, similar to https://github.com/DavidoTek/ProtonUp-Qt/pull/263

> sonic2kk: An equivalent change can be made for VKD3D as well, though that can go in a separate PR that will depend on whether # 262 is accepted or not ;-)

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_vkd3dlutris.py
+++ b/pupgui2/resources/ctmods/ctmod_vkd3dlutris.py
@@ -8,7 +8,7 @@ from pupgui2.resources.ctmods.ctmod_vkd3dproton import CtInstaller as VKD3DInsta
 
 
 CT_NAME = 'vkd3d-lutris'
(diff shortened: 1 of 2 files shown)


sonic2kk:
Tested a couple versions of each locally, they install as expected:

image

Also confirmed the folder structure matches what Heroic downloads for vkd3d-proton. I expected it to, since it mirrors what DXVK/D8VK use folder (top level folder with `x86` and `x64` folders inside). Checking and unchecking the "Auto Install/Update VKD3D on Prefix" checkbox in the Heroic game settings forces it to download/fetch VKD3D, which is how I verified the folder structure.

image

DavidoTek:
Thanks for also testing this. Great! I will merge this now.

Issue #243, opened 2023-05-23
Add dxvk-gplasync as a compatibility tool


NaiveInvestigator:
**Describe the solution you'd like**
To add this project [`dxvk-gplasync`](https://gitlab.com/Ph42oN/dxvk-gplasync) as an compatibility tool option. 

It is an dxvk async fork which increases performance well beyond what the 2.0 version does and brings a signficant boost in performance on low end PCs. Way more than the async dxvk 2.0. Well atleast on my machine lol.

I hope you consider this feature request! 

Also not related to this request lol but TYSM protonup-qt!! It helped a lot and was one of the main factors for me to try switching to linux full time.

sonic2kk:
There is no existing logic that can fetch from GitLab, so it would take a bit of extra effort to implement, but it should be feasible to add this for Lutris. Once we figure out DXVK etc for Heroic it would also be possible for them.

It is probably possible to use this tool with Proton but it would involve modifying Proton builds' `lib`s and their prefixes, so that part is probably out-of-scope. It would be possible to add for Lutris though provided it has the same folder structure as DXVK (which I assume it would).

The main issue I see here is that the repository doesn't appear to distribute "packages" but rather stories archives on the repo itself, which is an unusual way of going about things. This would probably need to be a separate ctmod and wouldn't be easily derived from an existing one since it uses GitLab and uses VCS for storing the packages, so we'd need to clone the entire repo and extract the selected version, which brings me to the next problem: How to go about fetching the versions as well as then selecting them. Maybe this is something to report upstream first, to ask them to use packages, and then we'd investigate using a GitLab API to fetch the releases that way.

Side note: I thought GPL was just an option that was enabled for RADV (and that it was on by default with Mess 23.1) but I'm not a graphics expert, so it's very possible I'm wrong here :-)

sonic2kk:
As an aside, I am wondering what the likelihood of this being "upstreamed" into dxvkasync is. The same argument could be made and has been made for D8VK being merged upstream into DXVK, and we still include that.

sonic2kk:
I wanted to take a look at this a while back, but work+other projects got on top of me a little bit.

Might take a look at it later this week if I can spare some time, the main thing is as mention allowing downloads from both GitHub and GitLab, since they are two different APIs. The extraction and general makeup of the ctmod should be similar, it just has to download from GitLab using that API.

[...]

PR #269, opened 2023-07-23
DXVK Async: Make DXVK Async inherit from DXVK

sonic2kk:
Another code refactor like #262, but this time for DXVK and DXVK Async.

This PR allows DXVK Async to inherit from the DXVK ctmod, removing a lot of the repeated code between the two ctmods. The ctmods were so similar that no changes at all were needed to the DXVK ctmod, which is why there's only a diff for DXVK Async.

If you compare the code for DXVK and DXVK Async, they are almost identical. Even the archive formats for releases are identical and have been for several releases at least ([DXVK releases](https://github.com/doitsujin/dxvk/releases/) and [DXVK Async](https://github.com/Sporif/dxvk-async/releases/) releases), they both use `.tar.gz`.

In fact, the only difference between the ctmods (aside from the ctmod URLs being different) is that DXVK checks where it should be extracted to because it now supports Heroic. Before that they were very similar.

Another benefit of this change is that it would make it trivial to add Heroic support for DXVK Async. We won't need to copy the extraction directory check or change any other logic, as that is all taken care off by the parent DXVK ctmod. All we would need to do is add `'heroicwine', 'heroicproton'` to the `CT_LAUNCHERS` list for DXVK Async. The only reason I didn't do it in this PR is because it may be a good idea to test this separately, though I'd be happy to add it for this PR as well.

So the main benefits to this PR are:
1. Less duplicated code, always nice to see :smile: 
2. Makes it easy to extend support to Heroic (or other launchers in future) as the logic only has to be added to the parent class

<hr>

There may be other reasons why this is not desired, such as a preference to have explicitly different ctmods for projects from different repos. But if it is desired, it would probably be possible to extend this kind of inheritance to other DXVK ctmods as well, such as DXVK Nightly and D8VK. That might require a little bit more work more akin to the recent changes for vkd3d (extraction logic changes, etc). This pattern may also make it easier to add support for other DXVK variations, such as DXVK Async-gpl (#243) and D8VK stable. This PR could be considered a "proof-of-concept" for applying this inheritance pattern to DXVK.

Hope my explanation and rationale made sense.

Thanks! :smile:

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_z1dxvkasync.py
+++ b/pupgui2/resources/ctmods/ctmod_z1dxvkasync.py
@@ -2,12 +2,9 @@
 # DXVK with async patch for Lutris: https://github.com/Sporif/dxvk-async/
 # Copyright (C) 2022 DavidoTek, partially based on AUNaseef's protonup
 
-import os
-import requests
+from PySide6.QtCore import QObject, QCoreApplication
 
-from PySide6.QtCore import QObject, QCoreApplication, Signal, Property
-
-from pupgui2.util import ghapi_rlcheck, extract_tar
+from pupgui2.resources.ctmods.ctmod_z0dxvk import CtInstaller as DXVKInstaller
 
 
 CT_NAME = 'DXVK Async'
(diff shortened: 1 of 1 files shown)


DavidoTek:
That makes sense, they are in fact the same :)

> such as DXVK Nightly and D8VK

Yes, they are a bit different as they require artifacts from nightly.link. It should be possible to implement a "universal nightly.link downloader" though.

> such as DXVK Async-gpl (https://github.com/DavidoTek/ProtonUp-Qt/issues/243)

Having base classes that handle similar tools seems to be a good idea, that makes adding tools much easier.
I wonder how well we could just override the methods for use with Gitlab (maybe add a `__fetch_gitlab_data` to the child class and then override `get_tool`, just an idea).

---

Thanks.

sonic2kk:
> I wonder how well we could just override the methods for use with Gitlab (maybe add a `__fetch_gitlab_data` to the child class and then override `get_tool`, just an idea).

I was wondering something similar :-) In regards to this and and a universal nightly.link downloader, I had actually looked into something along these lines: Creating some "unified" helper methods for downloading files with requests. The main "issue" I came across with implementing a feature like this is that the download methods in ctmods do things like update the progress bar or check if the download has been cancelled.

I think having more methods for repeated code in ctmods that we can call/override is a great idea. Maybe for GitHub/GitLab we could have a single `__fetch_data` method, and then in that method we can check if a URL is GitHub or GitLab. It's something to think about down the road :-)

DavidoTek:
> I think having more methods for repeated code in ctmods that we can call/override is a great idea.

> ```python
> class CtInstaller(QObject):
> ```

I noticed that most super classes still inherit from QObject. We could create a "real" super class SuperCtInstaller from which all ctmods inherit from.

That could implement the download functionallity, fetch_data for GitHub and GitLab and maybe even `get_extract_dir` (though that would need an additional parameter so it knows whether it's dealing with dxvk, vkd3d, proton or wine)

PR #270, opened 2023-07-23
Gamelist: Search game on AWACY when double-clicking status

DavidoTek:
When double-clicking the AreWeAntiCheatYet.com status in the gamelist dialog and the awacy status is known, the awacy site is opened with only the selected game displayed (game name in search).
If the status of the game is not known, https://areweanticheatyet.com/ will open normally.

Search URL: https://areweanticheatyet.com/?search={GAMENAME}&sortOrder=&sortBy=


sonic2kk:
This is pretty dang useful, I didn't know AWACY had a search feature.

Works in my quick test. Caught me off guard slightly that the URL always has the `search=&sortOrder=&sortBy=` parameters until I checked the code, but the site is smart enough to open the normal page even with these parameters (I double-checked that there was no difference in the list across Chromium and Firefox).

DavidoTek:
> Caught me off guard slightly that the URL always has the search=&sortOrder=&sortBy= parameters until I checked the code

That's actually the behavior of the site itself, if you clear the search box it will just set search to an empty string in the URL :smile:

PR #271, opened 2023-07-26
Heroic: Add Amazon Games to Games List

sonic2kk:
Heroic recently introduced support for Amazon Games, in the v2.9.0 release a few days ago. This PR add support for listing installed Amazon Games in the Heroic Games List.

image

All games marked as "nile" are the Amazon Games. Internally, Heroic refers to Amazon Games as "Nile" games, pretty clever!

## Existing Heroic Implementation
As a quick recap on how the existing GOG, sideload and Legendary games work (#168 has a more in-depth breakdown):
- Every Heroic game has a unique identifier string called `app_name`. GOG, sideload and Legendary have different formats (GOG typically uses numbers, sideload and Legendary typically use an alphanumeric string).
- Sideload and GOG games store information in `/path/to/heroic/sideload_apps/library.json` and `/path/to/heroic/gog_store/library.json` - These store basic game information such as title and `app_name` (the internal identifier for games managed by Heroic). It also stores an indicator for whether or not a game is installed (`is_installed`), but no further information about the installation.
- Sideloaded apps are only in the library if they are "Installed" (manually added), so it doesn't differentiate between a sideload library and list of installed sideloaded games (if a sideloaded game is in the library, it has to be installed). So all information we need (apart from Wine information) is stored in the sideload `library.json`
- GOG tracks installed games separately in `/path/to/heroic/gog_store/installed.json`, and its here that we can get the installation location and so on. We relate the entries in `library.json` and `installed.json` by their `app_name`.
    - Some GOG games are not marked as installed in `library.json` even if they are installed, so we have a utility function to cross-reference the `installed.json` and verify this.
- Legendary games are stored differently again, in `/path/to/heroic/legendary/installed.json`, and this file gives us almost all the information we need (again, apart from Wine information). There is no library JSON for Legendary games, so this file has game install information too.
- Finally, each game type has a JSON configuration file named based on its `app_name` at `/path/to/heroic/GamesConfig/<app_name>.json` - This file gives us our Wine information, and we can find the right file because we store `app_name`.

To generate the list of `HeroicGame`s, we have two loops:
- One loop generates the list of GOG and sideloaded games, because the file structures are so similar save for a handful of discrepancies which we allow for with utility functions. Each `HeroicGame` can also find its own config file at `GamesConfig/<app_name>.json`. The keys for the most part in the library.json are the same, and as far as I can tell, identical for *all* game types in `GamesConfig/<app_name>.json`.
- A separate loop is used for Legendary games, since the file structure here is slightly different.

[...]

PR #272, opened 2023-07-28
Heroic: Correctly display Browser games in Games List

sonic2kk:
## Overview
Heroic v2.9.0 added a feature for adding Browser games to your library. This works by giving the game a name, path, and a URL. Heroic will then run this game in its Browser, which is Chromium (with a spoofed browser agent to act like Google Chrome afaik).

ProtonUp-Qt can display these, but this PR fixes up their entries to make them a bit more useful.

Main branch:
image

This PR:
image

## Background
Browser games are treated internally like sideloaded apps, with their information stored in the `sideload_apps.json` file. They also still have an `app_name`. So there were no code changes required to get Browser games to display. Right now, ProtonUp-Qt _will_ display these browser games in the Heroic Games List, but with a couple of problems:
1. Compatibility Tool is listed as "Not Found", because Browser games actually have a Wine field but the Wine name in the Heroic JSON file is "Not Found" - Not sure why they store it like this, because there is no option (currently at least) to run a game in a browser with Wine or anything like that.
2. Install Location is listed as "`.`", because sideload apps have a `folder_name` key to store their installation folder name, that _is actually_ what the `folder_name` is in the JSON.

So the problems here are actually that ProtonUp-Qt is reading the information properly for Browser games, but that the information is not really useful for Browser games.

## Implementation
To make Browser game listings a bit more useful, a few changes were made:

### Properly Display Browser "Compatibility Tool" Name
Check on the `HeroicGame.platform` string when setting the "Compatibility Tool" name. This was not done initially out of concern that it may not be as reliable as checking for Wine information to differentiate between a Windows and native title. But now that we have Browser titles, I think it's safe to check on the platform string.

When we're using Windows we still build the compatibility tool string the same as normal, but in the else block, instead of hardcoding to "Native" we now set the `game.platform` string. This does mean that native titles will say "Linux" instead of "Native" now, but I don't think that's a big problem, and it saves on conditional logic to check "if platform is linux then set string and tooltip to Native". Doing it with the platform string is much more straightforward.

### Use `browserUrl` for the Install Location
The Install Location column has a bit of logic to try its best to find the Heroic installation path if we can't find it. But for Browser games we know this won't be set at all, so instead, when we're building the list of Heroic games with `get_heroic_game_list`, we check for `browserUrl` (set properly for Browser games and non-existent for all other sideload apps) before we check for `folder_name` (set to "`.`" for browser games and because it has a value, it would be truthy, so the check for `browserUrl` has to go before it, since that will be falsey for non-Browser sideload games).

### Double-Click action for Browser Games
When double clicking the "Install Location" for sideload/gog/legendary games, if the path exists, it will be opened in the file browser with `xdg-open`. But since the Browser URL set above doesn't fails at the `os.path.isdir` check, double-clicking would do nothing because we don't set the data on the list item.

To fix this, I added an optional parameter `ignore_invalid_path` to `set_item_data_directory` which is used to call `setData` on the list item passed to it even if the path doesn't exist. The default value for this parameter is `False`.

### Amending Tooltips
For the Compatibility Tool, as I hinted at earlier, the tooltip has been adjusted from saying `Platform: Native` to `Platform: {game.platform}`. It uses the platform string instead of being hardcoded.

For the Install Location, the tooltip was updated from saying `Double-click to browse` to `Double-click to open in browser`, to more accurately reflect what it does.

<hr>

That sums up the changes and rationale here. All feedback is welcome on this :-) 

Thanks!

### DIFF
--- a/pupgui2/heroicutil.py
+++ b/pupgui2/heroicutil.py
@@ -36,7 +36,7 @@ def get_heroic_game_list(heroic_path: str) -> List[HeroicGame]:
         hg.heroic_path = heroic_path
         # Sideloaded games uses folder_name as their full install path, GOG games store a folder_name but this is *just* their install folder name
         # Prioritise getting install_path for GOG games as this is the GOG game equivalent to 'folder_name'
-        hg.install_path = get_gog_installed_game_entry(hg).get('install_path', '') if hg.runner.lower() == 'gog' else game.get('folder_name', '')
+        hg.install_path = get_gog_installed_game_entry(hg).get('install_path', '') or game.get('install', []).get('install_path', '') or game.get('browserUrl', '') or game.get('folder_name', '')
         hg.store_url = game.get('store_url', '')
         hg.art_cover = game.get('art_cover', '')  # May need to replace path if it has 'file:///app/blah in name - See example in #168
         hg.art_square = game.get('art_square', '')
--- a/pupgui2/heroicutil.py
+++ b/pupgui2/heroicutil.py
@@ -36,7 +36,7 @@ def get_heroic_game_list(heroic_path: str) -> List[HeroicGame]:
         hg.heroic_path = heroic_path
         # Sideloaded games uses folder_name as their full install path, GOG games store a folder_name but this is *just* their install folder name
         # Prioritise getting install_path for GOG games as this is the GOG game equivalent to 'folder_name'
-        hg.install_path = get_gog_installed_game_entry(hg).get('install_path', '') or game.get('install', []).get('install_path', '') or game.get('browserUrl', '') or game.get('folder_name', '')
+        hg.install_path = get_gog_installed_game_entry(hg).get('install_path', '') or game.get('install', {}).get('install_path', '') or game.get('browserUrl', '') or game.get('folder_name', '')
         hg.store_url = game.get('store_url', '')
         hg.art_cover = game.get('art_cover', '')  # May need to replace path if it has 'file:///app/blah in name - See example in #168
         hg.art_square = game.get('art_square', '')
--- a/pupgui2/pupgui2gamelistdialog.py
+++ b/pupgui2/pupgui2gamelistdialog.py
@@ -252,7 +252,7 @@ class PupguiGameListDialog(QObject):
                 compat_tool_tooltip += self.tr('\nType: {wine_type}').format(wine_type=game.wine_info.get("type", "").capitalize()) if game.wine_info.get('type', '') else ''
             else:
                 # Linux/Browser games
-                compat_item_text = self.tr(game.platform)
+                compat_item_text = self.tr('Browser') if game.platform.lower() == 'browser' else self.tr('Native')
                 compat_tool_tooltip = self.tr('Type: {PLATFORM}').format(PLATFORM=game.platform)
 
             compat_item.setText(compat_item_text)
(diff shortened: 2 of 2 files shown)


pupgui2/heroicutil.py
```diff
@@ -36,7 +36,7 @@ def get_heroic_game_list(heroic_path: str) -> List[HeroicGame]:
         hg.heroic_path = heroic_path
         # Sideloaded games uses folder_name as their full install path, GOG games store a folder_name but this is *just* their install folder name
         # Prioritise getting install_path for GOG games as this is the GOG game equivalent to 'folder_name'
-        hg.install_path = get_gog_installed_game_entry(hg).get('install_path', '') if hg.runner.lower() == 'gog' else game.get('folder_name', '')
+        hg.install_path = get_gog_installed_game_entry(hg).get('install_path', '') or game.get('install', []).get('install_path', '') or game.get('browserUrl', '') or game.get('folder_name', '')
```
DavidoTek, line 39: I think that line needs to be:

```
hg.install_path = get_gog_installed_game_entry(hg).get('install_path', '') or game.get('install', {}).get('install_path', '') or game.get('browserUrl', '') or game.get('folder_name', '')
```

(replaced game.get('install', []).get with game.get('install', {}).get)

---

I guess https://github.com/DavidoTek/ProtonUp-Qt/pull/271 should be merged first for simplicity.

pupgui2/pupgui2gamelistdialog.py
```diff
@@ -251,15 +251,20 @@ def update_game_list_heroic(self):
                     compat_item.setData(Qt.UserRole, lambda path: os.system(f'xdg-open "{compat_tool_folder}"'))
                 compat_tool_tooltip += self.tr('\nType: {wine_type}').format(wine_type=game.wine_info.get("type", "").capitalize()) if game.wine_info.get('type', '') else ''
             else:
-                compat_item_text = self.tr('Native')
-                compat_tool_tooltip = self.tr('Type: Native')
+                # Linux/Browser games
+                compat_item_text = self.tr(game.platform)
```
DavidoTek, line 255: Translations sadly won't work like this. We could do something like:

```python
compat_item_text = self.tr('Browser') if game.platform == 'Browser' else self.tr('Native')
```

DavidoTek:
> Compatibility Tool is listed as "Not Found",

They probably didn't want to break existing implementations by removing the wine field

> set properly for Browser games and non-existent for all other sideload apps

Interesting, here the field `browserUrl` won't exist for some entries. Luckly, if they should ever add it to all entries for consistency, the logic will still work as Python's `or` handles `''` that same as `None` ;)

---

Thanks! :tada:

pupgui2/heroicutil.py
sonic2kk: Merging #271 first is a good idea, pushed a fix to this branch anyway though :-) Good catch on both PRs!

pupgui2/pupgui2gamelistdialog.py
sonic2kk: Fixed :-) Translations always catch me out, thanks for staying on top of this in my PRs.

sonic2kk:
Something that occurred to me when addressing the translation comment, I wonder if we should have another enum for Heroic platform types? Something like:

```python
class HeroicPlatform(enum):
    WINDOWS = 'Windows'
    LINUX = 'Native'  # Or 'Linux', no preference here :-)
    BROWSER =  'Browser'
```

Then we can check on `game.platform == HeroicPlatform.WINDOWS` and so on instead of a "raw" string comparison. Probably doesn't make much of a difference, just a thought I had.

DavidoTek:
> I wonder if we should have another enum for Heroic platform types?

That seems like a good idea. Feel free to add it if you want.

Otherwise I think this is ready to merge. Thanks.

sonic2kk:
Did a bit more testing and it would be more involved than I thought, so it can be merged as-is and refactored later :-)

PR #275, opened 2023-08-04
Fixed non steam game names not shown

GreatApo:
For some reason, my steam deck has AppName as lowercase in the .steam/steam/userdata/xxxxx/config/shortcuts.vdf
This is a quick fix that supports both uppercase and lowercase variations.

### DIFF
--- a/pupgui2/steamutil.py
+++ b/pupgui2/steamutil.py
@@ -127,8 +127,10 @@ def get_steam_shortcuts_list(steam_config_folder: str, compat_tools: dict=None) 
                 app.app_id = appid
                 app.shortcut_id = sid
                 app.shortcut_path = svalue.get('StartDir')
-                app.app_type='game'
+                app.app_type = 'game'
                 app.game_name = svalue.get('AppName')
+                if app.game_name == None or app.game_name == '':
+                    app.game_name = svalue.get('appname')
                 if ct := compat_tools.get(str(appid)):
                     app.compat_tool = ct.get('name')
                 apps.append(app)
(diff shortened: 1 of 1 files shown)


pupgui2/steamutil.py
```diff
@@ -127,8 +127,10 @@ def get_steam_shortcuts_list(steam_config_folder: str, compat_tools: dict=None)
                 app.app_id = appid
                 app.shortcut_id = sid
                 app.shortcut_path = svalue.get('StartDir')
-                app.app_type='game'
+                app.app_type = 'game'
                 app.game_name = svalue.get('AppName')
+                if app.game_name == None or app.game_name == '':
+                    app.game_name = svalue.get('appname')
```
sonic2kk, line 133: `app.game_name` = svalue.get('AppName') or svalue.get('appname') should also work here and might be slightly cleaner. Python should treat `None` and `''` (empty string) both as falsey.

No problem at all either way, it just might save a few lines without sacrificing readability too much.

GreatApo: Neat, tested and seems to work nicely

sonic2kk:
Looks great and good catch!

Tested locally and this doesn't cause any regressions with my `shortcuts.vdf` file on my PC, though it is not a very extensive list of shortcuts. Since the `shortcuts.vdf` file uses the binary VDF format, I wasn't able to test editing it to use a lowercase key name, but I assume it will work from looking at the code.

This reminds me of an issue for #226 where one of the VDF files keys had a casing issue - `valve` instead of `Valve` (fixed in #227). It seems VDF file casing can be rather inconsistent. On my Steam Deck, shortcuts display fine fwiw, so I'd go out on a limb and say these are Steam bugs. Regardless it's good to fix them in ProtonUp-Qt as well.

pupgui2/steamutil.py
GreatApo: Neat, tested and seems to work nicely

DavidoTek:
Thanks!

> Tested locally and this doesn't cause any regressions with my shortcuts.vdf file on my PC which has a CamelCase key name, though it is not a very extensive list of shortcuts

Seems good, I don't think it can break anything.

> It seems VDF file casing can be rather inconsistent.

Definitively.

PR #282, opened 2023-08-28
Add a Steam Shortcut Editor

DavidoTek:
This PR adds a Shortcut Editor for Steam (both Native and Flatpak supported).

Currently, editing the App Name, Executable Path, Starting Directory and App Icon is supported.

grafik

### TODO/DONE:

- [x] Add a Delete Button
   - Mark shortcut (red text) and delete it after pressing `Save`
- [x] Allow adding new shortcuts
   - Determine a user for new shortcuts like this: Check if there are already other shortcuts
      - If there are: Use the most common user across the other shortcuts
      - Otherwise: Determine which Steam user was most recently logged in
      - If there isn't any Steam user (e.g. fresh install): Return and don't do anything - currently no user feedback
- [x] Maybe resize the window a bit

### Later TODOs (for later PRs):

- Add a tool button that lets the user choose an executable/directory using a file dialog
- Modify the available columns / let the user choose
- Fix the Start In field automatically: https://github.com/DavidoTek/ProtonUp-Qt/pull/282#discussion_r1331177772

### DIFF
--- a/pupgui2/pupgui2gamelistdialog.py
+++ b/pupgui2/pupgui2gamelistdialog.py
@@ -54,6 +54,8 @@ class PupguiGameListDialog(QObject):
         elif is_heroic_launcher(self.launcher):
             self.setup_heroic_list_ui()
 
+        self.ui.btnShortcutEditor.setVisible(self.launcher == 'steam')
+
         self.ui.btnSearch.setVisible(False)
         self.ui.searchBox.setVisible(False)  # Hide searchbox by default
 
@@ -86,8 +88,6 @@ class PupguiGameListDialog(QObject):
         self.ui.tableGames.setColumnWidth(0, 300)
         self.ui.tableGames.setColumnWidth(3, 70)
         self.ui.tableGames.setColumnWidth(4, 70)
-
-        self.ui.btnShortcutEditor.setVisible(True)
     
     def setup_lutris_list_ui(self):
         self.ui.tableGames.setHorizontalHeaderLabels([self.tr('Game'), self.tr('Runner'), self.tr('Install Location'), self.tr('Installed Date'), ''])
@@ -99,8 +99,6 @@ class PupguiGameListDialog(QObject):
         self.ui.tableGames.setColumnWidth(3, 30)
         self.ui.tableGames.setColumnHidden(4, True)
 
-        self.ui.btnShortcutEditor.setVisible(False)
-
     def setup_heroic_list_ui(self):
         self.ui.tableGames.setHorizontalHeaderLabels([self.tr('Game'), self.tr('Compatibility Tool'), self.tr('Install Location'), self.tr('Runner'), ''])
         self.update_game_list_heroic()
@@ -110,8 +108,6 @@ class PupguiGameListDialog(QObject):
         self.ui.tableGames.setColumnWidth(2, 250)
         self.ui.tableGames.setColumnWidth(3, 40)
         self.ui.tableGames.setColumnHidden(4, True)
-
-        self.ui.btnShortcutEditor.setVisible(False)
 
     def update_game_list_steam(self, cached=True):
         """ update the game list for the Steam launcher """
--- a/pupgui2/pupgui2gamelistdialog.py
+++ b/pupgui2/pupgui2gamelistdialog.py
@@ -307,7 +307,7 @@ class PupguiGameListDialog(QObject):
 
     def btn_shortcut_editor_clicked(self):
         self.ui.close()
-        PupguiShortcutDialog(self.install_loc.get('vdf_dir'), self.ui)
+        PupguiShortcutDialog(self.install_loc.get('vdf_dir'), self.game_property_changed, self.ui)
 
     def search_gamelist_games(self, text):
         for row in range(self.ui.tableGames.rowCount()):
--- a/pupgui2/pupgui2gamelistdialog.py
+++ b/pupgui2/pupgui2gamelistdialog.py
@@ -12,6 +12,7 @@ from PySide6.QtUiTools import QUiLoader
 from pupgui2.constants import PROTONDB_COLORS, STEAM_APP_PAGE_URL, AWACY_WEB_URL, PROTONDB_APP_PAGE_URL, LUTRIS_WEB_URL
 from pupgui2.datastructures import AWACYStatus, SteamApp, SteamDeckCompatEnum, LutrisGame, HeroicGame
 from pupgui2.lutrisutil import get_lutris_game_list
+from pupgui2.pupgui2shortcutdialog import PupguiShortcutDialog
 from pupgui2.steamutil import steam_update_ctools, get_steam_game_list
 from pupgui2.steamutil import is_steam_running, get_steam_ctool_list
 from pupgui2.steamutil import get_protondb_status
@@ -66,6 +67,7 @@ class PupguiGameListDialog(QObject):
         self.ui.btnApply.clicked.connect(self.btn_apply_clicked)
         self.ui.btnSearch.clicked.connect(self.btn_search_clicked)
         self.ui.btnRefreshGames.clicked.connect(self.btn_refresh_games_clicked)
+        self.ui.btnShortcutEditor.clicked.connect(self.btn_shortcut_editor_clicked)
         self.ui.searchBox.textChanged.connect(self.search_gamelist_games)
 
         # Hide Search button and disable shortcut if no games
@@ -84,6 +86,8 @@ class PupguiGameListDialog(QObject):
         self.ui.tableGames.setColumnWidth(0, 300)
         self.ui.tableGames.setColumnWidth(3, 70)
         self.ui.tableGames.setColumnWidth(4, 70)
+
+        self.ui.btnShortcutEditor.setVisible(True)
     
     def setup_lutris_list_ui(self):
         self.ui.tableGames.setHorizontalHeaderLabels([self.tr('Game'), self.tr('Runner'), self.tr('Install Location'), self.tr('Installed Date'), ''])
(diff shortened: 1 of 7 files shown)


sigboe:
Launch options is a generally a nice thing to be able to edit

pupgui2/pupgui2gamelistdialog.py
```diff
@@ -84,6 +86,8 @@ def setup_steam_list_ui(self):
         self.ui.tableGames.setColumnWidth(0, 300)
         self.ui.tableGames.setColumnWidth(3, 70)
         self.ui.tableGames.setColumnWidth(4, 70)
+
+        self.ui.btnShortcutEditor.setVisible(True)
```
sonic2kk, line 90: Since this will only apply to Steam, maybe in `self.setup_ui`, alongside the `btnSearch` and `searchBox` calls, we could add `self.ui.btnShortcutEditor.setVisible(self.launcher == 'steam')`. That'll save a few calls :-)

Alternatively, we could just set it to `False` by default at the top of `setup_ui`, and then set it to `True` inside of `setup_steam_list_ui` (or in the `if self.launcher == 'steam'` check).

pupgui2/pupgui2shortcutdialog.py
```diff
@@ -0,0 +1,105 @@
...
+        self.ui.tableShortcuts.setRowCount(len(self.shortcuts))
+
+        for i, shortcut in enumerate(self.shortcuts):
+            txt_name = QLineEdit(shortcut.game_name)
+            txt_exe = QLineEdit(shortcut.shortcut_exe)
+            txt_path = QLineEdit(shortcut.shortcut_startdir)
+            txt_icon = QLineEdit(shortcut.shortcut_icon)
+
+            txt_name.editingFinished.connect(lambda i=i: self.txt_changed(i, 0))
+            txt_exe.editingFinished.connect(lambda i=i: self.txt_changed(i, 1))
```
sonic2kk, line 57: This is *way* into minor usability enhancement territory here, but I had a thought related to this. Since the `shortcuts.vdf` surrounds the paths in quotes, it would be handy if we could conditionally surround the path in quotes. For example:
- User enters path without quotes -> we fill in the quotes before writing out
- User enters path with quotes -> no change, happy path :-)
- User enters path and accidentally removes one quote -> not sure if the Steam Client can handle this and fill in the gap, but we could append a quote to the start/end of the text if it's missing

This might take some work so we could always do this another time.

pupgui2/steamutil.py
```diff
@@ -586,3 +589,76 @@ def install_steam_library_shortcut(steam_config_folder: str, remove_shortcut=Fal
...
+        # read shortcuts.vdf
+        shortcuts_vdf = {}
+        with open(shortcuts_file, 'rb') as f:
+            shortcuts_vdf = vdf.binary_load(f)
+
+        # update existing shortcuts or create new ones
+        for sid in list(shortcuts_vdf.get('shortcuts', {}).keys()):
+            if not sid in shortcuts_by_user.get(userf, {}):
+                continue
+            svalue = shortcuts_vdf.get('shortcuts', {}).get(sid)
```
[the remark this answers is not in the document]
sonic2kk: Maybe this needs to go outside the `for` loop? Since this is for updating existing entries.

It could go in an `else` block for the `for` loop, which should mean new shortcuts get added to the end of the dictionary and thus onto the bottom of the file. I don't think Steam is fussy, but appending new shortcuts to the bottom of the file and "preserving" the order probably won't harm anything.

sonic2kk:
This is really awesome. There are some other fields which may only be useful in some cases (such as the Steam Overlay conditional). There's a "hidden" flag but in my own testing I couldn't get it to work, so I'm not sure if it's worth adding.

Another thing I'm thinking is how difficult it would be to add some kind of way to browse for executables/icons. I couldn't think of an easy way to do this myself so I don't have any suggestions 😅 

I love this though. As someone who has also worked a lot recently with the damn `shortcuts.vdf` I can say any software that makes it less of a pain to work with, the better!

pupgui2/pupgui2shortcutdialog.py
sigboe: There is actually a bug in the Steam client related to this, would be funny if ProtonUp-Qt handled it properly first.
The "Start In" field in the steam client doesn't have this problem, the string in the field is taken literally,  as if its quoted. 
The Target field requires quotes, especially if there are spaces in the path, or it wont work. As you need to quote a path with spaces if it is to be taken as a literal path with spaces in the path. When you make a new shortcut, the Target field is automatically quoted. But if you use the browse button to select a new target it is not then quoted or handled as if it is quoted. If there is spaces in the path then the shortcut doesn't work. Add the quotes manually and it works.

I've seen videos on youtube where people encounter this issue, and then make a new shortcut because they don't know it needs quotes. But this does give a new appID and a new compatdata prefix, so it may sometimes not work as well, especially if the program is not portable. (changing the target from the setup.exe to the installed exe file)

I have looked for this issue on the issue tracker on the steam for linux github page, and this is not talked about there. I haven't made an issue yet

sigboe:
> Another thing I'm thinking is how difficult it would be to add some kind of way to browse for executables/icons. I couldn't think of an easy way to do this myself so I don't have any suggestions 😅

@sonic2kk  I don't know really about the Qt framework widgets that well, but the (I don't know the actual name) list/table view widget in OP's picture. Is it possible to add a browse button in the relevant fields that is only visible when the relevant field is selected? And then just open the Qt file/folder chooser

pupgui2/steamutil.py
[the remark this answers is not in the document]
DavidoTek: > Maybe this needs to go outside the for loop? Since this is for updating existing

Yeah, I changed the loop variable. It now loops over the given shortcuts list and check whether they are already in the shortcuts.vdf.

pupgui2/pupgui2gamelistdialog.py
DavidoTek: Good suggestion.

pupgui2/pupgui2shortcutdialog.py
DavidoTek: > conditionally surround the path in quotes. For example:

The following logic adds quotes at the start and end of the string. It will also fail, if there are more than two quotes (I don't think paths on Linux are allowed to contain quotes so this should be fine).

```python
if os.path.exists(text.replace('"', '', 2)):
    if not text.startswith('"'):
        text = '"' + text
    if not text.endswith('"'):
        text = text + '"'
    self.ui.tableShortcuts.cellWidget(index, col).setText(text)
```

> There is actually a bug in the Steam client related to this, would be funny if ProtonUp-Qt handled it properly first

What we could do is to automatically do the above check for all shortcuts when starting ProtonUp-Qt. That would fit a separate PR I guess.

By "Start In" you probably mean the `StartDir` attribute. That should currently be fixed at least when manually editing using ProtonUp-Qt.

DavidoTek:
> There are some other fields which may only be useful in some cases (such as the Steam Overlay conditional). There's a "hidden" flag but in my own testing I couldn't get it to work, so I'm not sure if it's worth adding.

Adding new column is mostly straight forward. We can discuss which are important, I just chose ones that seemed relevant to me.

> browse for executables/icons [..] Is it possible to add a browse button in the relevant fields that is only visible when the relevant field is selected? And then just open the Qt file/folder chooser

It is possible to add Qt Widgets to the QTableWidget. By adding QLineEdit and a QToolButton inside of a QHBoxLayout this should work. The logic for updating text in `txt_changed` would need to be revised though.

pupgui2/pupgui2shortcutdialog.py
```diff
@@ -0,0 +1,161 @@
...
+        # new id should be higher than last one
+        highest_id = 0
+        for shortcut in self.shortcuts:
+            sid = int(shortcut.shortcut_id)
+            if sid > highest_id:
+                highest_id = sid
+
+        # assume that the most common user of other shortcuts is the correct one
+        # TODO: get this to work when there are no other shortcuts. Remember to change the tooltip.
+        most_common_user = Counter([s.shortcut_user for s in self.shortcuts]).most_common(1)[0][0]
```
DavidoTek, line 145: I've also added a `Add new shortcut` button. It currently only works if there already are other shortcuts as it gets the user folder id from them. Not sure if there is a better solution for figuring out the current user in case there are multiple (maybe let the user choose?).

sonic2kk: If there's a way to get the current user's username you can get their User ID from `config.vdf` and do `user_id & 0xFFFFFFFF` to convert it to the short ID that the folder name uses. Example structure:

```
"Accounts"
{
    "sonic2kk""
    {
        "UserID"        "123123123123123"
    }
}
```

I am not sure how other software does this, such as Steam-ROM-Manager. SteamTinkerLaunch assumes only one user.

One potential solution is to have a dropdown populated with usernames from the `config.vdf` file (so if the two users were `sonic2kk` and `davidotek`, it would display those), maybe something similar to how the launcher dropdown works on the main screen. This would probably mean refreshing the shortcuts list each time the user is changed to show that user's shortcuts, again like the main screen.

We could then just store the parsed `Accounts` block of `config.vdf`, since Python will store it as a dict. If there are some awkward elements there, or if we want to store some custom elements ahead-of-time (like the short UserID or the shortcuts VDF path) we could have a function parse the contents of the `Accounts` block from `config.vdf` and return a list of dictionaries representing each user something like this:

```python3
# We'd have a method to return a dict something like this
steam_users = {
    "sonic2kk": {
        "long_user_id": 12312323123,  # UserID raw from config.vdf
        "short_user_id": 444412,  # long_user_id & 0xFFFFFFFF
        "shortcuts_path": '/home/gaben/.local/share/Steam/config/short_user_id/config/shortcuts.vdf'
    },
    "davidotek": {
        "long_user_id": 45645656456,  # UserID raw from config.vdf
        "short_user_id": 555523,  # long_user_id & 0xFFFFFFFF
        "shortcuts_path": '/home/gaben/.local/share/Steam/config/short_user_id/config/shortcuts.vdf'
    }
}

dropdown_steam_users = [user for user in steam_users.keys()])
print(steam_users['davidotek'])
```

It would also mean being able to see other users' shortcuts, but I don't think that's a huge deal and is possible either way :-)

DavidoTek: > I just noticed that the file .../config/loginusers.vdf contains the flag "MostRecent". We could use this (though I haven't tested if it actually does what I think it does) as the user probably wants to add the shortcut to the currently used/most recent account:

I implemented following logic:

- Check if there are already other shortcuts
   - If there are: Use the most common user across the other shortcuts
   - Otherwise: Determine which Steam user was most recently logged in
   - If there isn't any Steam user (e.g. fresh install): Return and don't do anything

PR #285, opened 2023-09-01
SteamTinkerLaunch: minor description fix

sonic2kk:
Remove double '..' from the end of the tool description.

Pretty sure I noticed this a while ago and just forgot about it, until I was copypasting something from this description into an STL issue :sweat_smile:


DavidoTek:
Thanks.

I will update the translation files accordingly (I'll probably add a commit to this PR).
It may be time to also merge https://github.com/DavidoTek/ProtonUp-Qt/pull/252 then, though I wasn't able to check all translations for correctness ;)

PR #307, opened 2023-11-16
Move requests session to ctmods

DavidoTek:
Currently, all ctmods share a single requests session which may contain the authorization header for GitHub. With #302, fetching compatibility tools from GitLab is now possible too. This requires ctmods to be able to choose between the GitHub and GitLab authorization key.

This PR does following changes:
- Each ctmod has its own requests session
- The ctmods will read the authorization keys from `main_window.web_access_tokens` and build a header using `util.py#build_headers_with_authorization` which is added to the local requests session

See https://github.com/DavidoTek/ProtonUp-Qt/pull/302#issuecomment-1808909601 for details

### DIFF
--- a/pupgui2/pupgui2.py
+++ b/pupgui2/pupgui2.py
@@ -80,7 +80,6 @@ class MainWindow(QObject):
     def __init__(self):
         super(MainWindow, self).__init__()
 
-        self.rs = requests.Session()
         self.web_access_tokens = {
             'github': os.getenv('PUPGUI_GHA_TOKEN'),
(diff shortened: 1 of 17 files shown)


sonic2kk (approved):

sonic2kk:
Glad you also caught removing the header update in the DXVK ctmod's `__download`, I meant to bring that up :smile: 

This should also make it more convenient for other ctmods to adopt the download methods we use here. My main concern with this approach in #302 was to do with maybe a higher memory cost or something, but after discussion in that PR, it seems that is a non-issue and there is no concern with each ctmod having its own `requests.Session`.

Looks great :-)

PR #308, opened 2023-11-17
Config: Use read_update_config_value for advancedmode and theme

sonic2kk:
This PR replaces the logic in `util#config_theme` and `util#config_advanced_mode` to use the generic function `util#read_update_config_value` from #306, a change which was discussed in passing in that PR.

The logic in these functions was used as a template to create `read_update_config_value`, so there should be no compatibility breaking changes. The option names are also have not been changed. This change just makes the code slightly cleaner :-)

This PR does not touch `install_directory` or `config_custom_install_location`, as discussed we could implement this in future if we feel we need such a change.

Thanks!

### DIFF
--- a/pupgui2/util.py
+++ b/pupgui2/util.py
@@ -150,21 +150,7 @@ def config_theme(theme=None) -> str:
     Return Type: str
     """
 
-    config = ConfigParser()
-
-    if theme:
-        config.read(CONFIG_FILE)
-        if not config.has_section('pupgui2'):
-            config.add_section('pupgui2')
-        config['pupgui2']['theme'] = theme
-        os.makedirs(os.path.dirname(CONFIG_FILE), exist_ok=True)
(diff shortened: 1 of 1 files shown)


DavidoTek:
Thanks!

---

I made a small change to `read_update_config_value`. The default behaviour of `config_theme` is to return `None` if no parameter is given and the config doesn't exist yet, while the behaviour of `read_update_config_value` was to return `''`.

This would break the combobox index matching in the about dialog when no config file exists:

https://github.com/DavidoTek/ProtonUp-Qt/blob/4ac3b5f9aaa289b3aa68123137ecb5cd4cdf2a55/pupgui2/pupgui2aboutdialog.py#L63C41-L63C41

I adjusted it to return the "inherited" `value`. That shouldn't break the api token configs either. If we need the default return value to be a string, we can define this in the "parent" function.

PR #310, opened 2023-11-18
Roberta: Use fetch_project_releases and fetch_project_release_data

sonic2kk:
Updates the Roberta ctmod to use `util#fetch_project_releases` and `util#fetch_project_release_data` from #302.

I chose Roberta as a straightforward and lower-cost way to proof this method out for other ctmods :-)


sonic2kk:
I have noticed that some of them (Luxtorpeda, Roberta, Boxtron) seem pretty similar minus their dependencies. I was planning to go ahead and make these changes in separate PRs for these three tools, however because they're so similar, I may try to refactor these to use a common ctmod first. They're sister projects of one another I believe anyway so I think this makes sense.

Instead of making three separate PRs to make the fetching changes, and then exploring a refactor to make these ctmods common, I think I'll just make them common. Then, if that fails (i.e if it becomes difficult to set overrides to how we fetch assets for some reason) I can go back to this approach.

Issue #254, opened 2023-06-04
Tool marked as unused even if it is the "global" compatibility tool.


MithicSpirit:
**Describe the bug**  
Compatibility tools are marked as "(unused)" even if the tool is set as the global compatibility tool (under steam settings, steam play, run other titles with)

**To Reproduce**  
1. Install a compatibility tool like Proton GE
2. Set it as the global compatibility tool. This requires working directly with the configuration file due to https://github.com/ValveSoftware/steam-for-linux/issues/6043. Steps for doing so are available [here](https://github.com/sonic2kk/steamtinkerlaunch/issues/278#issuecomment-932826852), but instead of `Proton-stl` set it to the Proton GE version that was installed instead.
3. Re-open protonup-qt to check
4. Still marked as "(unused)"

**Expected behavior**  
I expect it to not be marked as unused. Ideally there'd also be a way to enable setting it as the global compatibility tool so that it's not necessary to go through all of the hassle of manually editing config file.

**Desktop (please complete the following information):**  
 - **Platform**: Laptop
 - **System**: Arch Linux (rolling)
 - **Version**: 2.8.0
 - **How did you install ProtonUp-Qt?**: AUR package [`protonup-qt`](https://aur.archlinux.org/packages/protonup-qt)

**Terminal output**  
```
qt.qpa.plugin: Could not find the Qt platform plugin "wayland" in ""
qt.pysideplugin: Environment variable PYSIDE_DESIGNER_PLUGINS is not set, bailing out.
qt.pysideplugin: No instance of QPyDesignerCustomWidgetCollection was found.
Qt WebEngine seems to be initialized from a plugin. Please set Qt::AA_ShareOpenGLContexts using QCoreApplication::setAttribute and QSGRendererInterface::OpenGLRhi using QQuickWindow::setGraphicsApi before constructing QGuiApplication.
ProtonUp-Qt 2.8.0 by DavidoTek. Build Info: built from source.
Python 3.11.3 (main, Apr  5 2023, 15:52:25) [GCC 12.2.1 20230201], PySide 6.5.1
Platform: Arch rolling Linux-6.3.5-zen1-1-zen-x86_64-with-glibc2.37
Loading locale en / en_US
Loaded ctmod GE-Proton
Loaded ctmod Wine-GE
Loaded ctmod Boxtron
Loaded ctmod D8VK (nightly)
Loaded ctmod Kron4ek Wine-Builds Vanilla
[...]

PR #314, opened 2023-11-23
Tool List: Mark Tool as Global

sonic2kk:
Addresses at least part of #254.

## Overview
This PR adds `(global)` beside installed compatibility tools which are set as the global tool in the Steam client's Steam Play settings. When no compatibility tool is forced, but when Steam detects that a game/app is a Windows-only application, this is the default tool that will be used to run that app from the Steam Client. This can be any compatibility tool, including things like GE-Proton. In such a case, even though the compatibility tool is technically not *explicitly* used, it is still the default and so used by all Windows games which do not have an alternative compatibility tool forced in the Properties dialog.

image

## Implementation
This PR accomplishes marking the tool as global, in summary, by doing the following:
- Adding a property `is_global` to `SteamApp`. This is then toggled manually in `pupgui2.py` when fetching the compat tool mapping list
- We update this property using a new `steamutil` function which returns the name of the `0` entry in config.vdf's `CompatToolMapping.` It looks in `config.vdf`, gets `CompatToolMapping`, and returns the name field for the `0` entry. This should always exist when Steam Play is enabled (required to even use Proton flavours and even show the "Compatibility" section of the Properties dialog), but if it's not, we have fallbacks in place to return an empty string.
- In `pupgui2.py`, when we're fetching all of the compatibility tools, if the name of a tool matches the global tool name, we mark it as default
- In `get_displayname`, which is called to return a string to set as the tool name in the games list, this function checks if the compat tool `is_global` and if it is, it sets `(global)`. We use an if/elif so that a tool is not marked as `global` *and* `unused`.

## Future Work
There are some future things we could do for this functionality, and this PR serves as a baseline for implementing them since it allows us to track which tool is the global one:
- Use badges instead of text, as it may still not be clear which tool is the global one
- Move the global compatibility tool to the top of the list, to make it visually clearer
- Show a warning dialog when trying to remove the global compatibility tool, like we do when trying to remove a used compatibility tool
    - I'm not sure how Steam handles a missing global compatibility tool and if it tries to fall back, but if you force something like GE-Proton8-22 for a single game in the Properties dialogue for that game, and then remove it, Steam will get confused and will not display the "Play" button, and there will be text beside it saying "Available for [Windows icon, and maybe a macOS icon]".
    - This is because the tracked compatibility tool in `config.vdf` under `CompatToolMapping` is still the previous tool, which would be `GE-Proton8-22` in this case. Checking and unchecking the use of a compatibility tool, or removing this entry altogether from `config.vdf`, will resolve the issue.
    - When removing the entry in `config.vdf`, Steam will fall back to the global compatibility tool and uncheck the checkbox (Steam knows to check/uncheck the checkbox based on whether that app's AppID has an entry under `config.vdf`'s `CompatToolMapping`). But I have no idea what happens if you remove the *global* compatibility tool if that makes sense. Since the fallback for all other apps *is* the global (`0`) compatibility tool, what happens when that one is missing? Maybe it just turns that setting off?
    - And so to go full circle, I Have no idea what happens if the global compatibility tool itself is missing, but marked in the `config.vdf` still. Unless Steam has a fallback this will not get updated, which may mean all tools show the same message as when a per-app forced compatibility tool is missing, and a user would have to manually update it
    - All this to say, at the very least a warning is probably necessary when removing the global tool, if not outright disabling the remove button (maybe we could only enable it in Advanced Mode?).
- Add an option to mark a given tool as the global compatibility tool. Steam would need to be closed for this action. This was also discussed in #254, but there was not a solid consensus on where to put it. Implementation shouldn't be that tricky since, at its core, it should just be setting the `0` compat tool's `name` in `CompatToolMapping` to the given tool's internal name.
- Potentially more I'm not thinking of :smile: 

<hr>

I actually wrote most of this back in June when I left my comment on #254, but I guess I abandonded it for some reason. Maybe I wanted to try and accomplish a lot more, like some of the things mentioned in Additional Work, but for now I figured this works and is better than not showing anything. I rebased it on main and polished it up a little before making this PR and made a couple of tiny adjustments, and now it's ready for review :-) 

Thanks!

### DIFF
--- a/pupgui2/datastructures.py
+++ b/pupgui2/datastructures.py
@@ -106,13 +106,13 @@ class BasicCompatTool:
     install_dir = ''
     install_folder = ''
     ct_type = CTType.UNKNOWN
+    is_global = False
 
     def __init__(self, displayname, install_dir, install_folder, ct_type = CTType.UNKNOWN) -> None:
         self.displayname = displayname
         self.install_dir = install_dir
         self.install_folder = install_folder
         self.ct_type = ct_type
-        self.is_global = False
 
     def set_version(self, ver : str) -> None:
         self.version = ver
--- a/pupgui2/datastructures.py
+++ b/pupgui2/datastructures.py
@@ -112,21 +112,26 @@ class BasicCompatTool:
         self.install_dir = install_dir
         self.install_folder = install_folder
         self.ct_type = ct_type
+        self.is_global = False
 
     def set_version(self, ver : str) -> None:
         self.version = ver
 
-    def set_global(self, global_tr='global'):
+    def set_global(self, is_global: bool = True):
+        self.is_global = is_global
 
-        self.displayname += f' ({global_tr})'
-
-    def get_displayname(self, unused_tr='unused') -> str:
+    def get_displayname(self, unused_tr='unused', global_tr='global') -> str:
         """ Returns the display name, e.g. GE-Proton7-17 or luxtorpeda v57 """
         displayname = self.displayname
         if self.version != '':
             displayname += f' {self.version}'
-        if self.no_games == 0:
+
+        # Don't mark global tools as unused
+        if self.is_global:
+            displayname += f' ({global_tr})'
+        elif self.no_games == 0:
             displayname += f' ({unused_tr})'
+
         return displayname
 
     def get_internal_name(self) -> str:
--- a/pupgui2/datastructures.py
+++ b/pupgui2/datastructures.py
@@ -116,6 +116,10 @@ class BasicCompatTool:
     def set_version(self, ver : str) -> None:
         self.version = ver
 
+    def set_global(self, global_tr='global'):
+
+        self.displayname += f' ({global_tr})'
+
     def get_displayname(self, unused_tr='unused') -> str:
         """ Returns the display name, e.g. GE-Proton7-17 or luxtorpeda v57 """
         displayname = self.displayname
--- a/pupgui2/pupgui2.py
+++ b/pupgui2/pupgui2.py
@@ -23,7 +23,7 @@ from pupgui2.pupgui2ctinfodialog import PupguiCtInfoDialog
 from pupgui2.pupgui2customiddialog import PupguiCustomInstallDirectoryDialog
 from pupgui2.pupgui2gamelistdialog import PupguiGameListDialog
 from pupgui2.pupgui2installdialog import PupguiInstallDialog
-from pupgui2.steamutil import get_steam_acruntime_list, get_steam_app_list, get_steam_ct_game_map
+from pupgui2.steamutil import get_steam_acruntime_list, get_steam_app_list, get_steam_ct_game_map, get_steam_global_ctool_name
 from pupgui2.heroicutil import is_heroic_launcher, get_heroic_game_list
 from pupgui2.util import apply_dark_theme, create_compatibilitytools_folder, get_installed_ctools, remove_ctool
 from pupgui2.util import install_directory, available_install_directories, get_install_location_from_directory_name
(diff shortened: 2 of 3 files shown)


sonic2kk:
Forgot to mention that this PR can be tested both by setting the global compatibility tool within Steam, or (with Steam closed to prevent any overwrites) by temporarily editing the `0` entry under `CompatToolMapping` section to use a different `name`. This is how I did most of my testing. For at least GE-Proton versions since the new naming scheme (though possibly further back as well) the internal name is the same as the actual name, so it can simply be set to something like `GE-Proton8-25`.

sonic2kk:
Just noticed the "unused" counter still counts the global tool as unused. The counter says "5" but only 4 are actually unused (GE-Proton8-25 is the global tool but appears to still be counted).

I will address this tomorrow :-)

DavidoTek:
Thanks!

Code looks good at first glance and what you wrote makes sense. Will do an in-depth review of the code later.

---

> Use badges instead of text, as it may still not be clear which tool is the global one

Ah, that would be nice :smile:

> Move the global compatibility tool to the top of the list, to make it visually clearer

Good idea, that makes it more obvious for people with >10 compatibility tools.

> All this to say, at the very least a warning is probably necessary when removing the global tool, if not outright disabling the remove button (maybe we could only enable it in Advanced Mode?).

We don't want anyone to break their setup, so a warning + advanced mode seems right.

> Add an option to mark a given tool as the global compatibility tool. Steam would need to be closed for this action. [...] but there was not a solid consensus on where to put it

Probably the ct info dialog is the best place to put it. I think the main ui would get to crowded if we put it there.

> I should also just clarify that the info dialog game list is technically empty for this tool

I guess that is fine. We could add a text saying "global tool" or something but it should be obvious from the main ui already.

> maybe in this PR or a follow-up we should disable that if the compat tool is marked as global.

We could leave it as a way to change the global compatibility tool from ProtonUp-Qt. But maybe there is a cleaner way to do this, maybe from the gamelist, idk.

> We could give it a separate dialog or modify the info dialog to not show the games list or game count

We could hide the game list and maybe rename "batch update" to "change global tool" or something

pupgui2/datastructures.py
```diff
@@ -112,17 +112,26 @@ def __init__(self, displayname, install_dir, install_folder, ct_type = CTType.UN
         self.install_dir = install_dir
         self.install_folder = install_folder
         self.ct_type = ct_type
+        self.is_global = False
```
DavidoTek, line 115: I wonder how we should initialize the class attribute here, other attributes are defined and initialized like static variables above __init__...
I guess your way seems the "more correct" one, although Python has this thing called [dataclasses](https://docs.python.org/3/library/dataclasses.html) which allow initialization from outside `__init__` (we don't use the `@dataclass` decorator though)...

---

Fun fact: There are three (four) ways you can add or initialize attributes for an object

```python
# "correct" one
class A:
    def __init__(self):
        self.my_attr = 3

# using static variables, which will also init the attributes for object
# I think it works because of Python's dataclasses
class A:
    my_attr = 3

# dataclasses with correct decoration that will move initialization to __init__
@dataclass
class A:
    my_attr: int = 3

# please don't do it this way
class A:
    pass
my_obj = A()
my_obj.my_attr = 3
```

sonic2kk: D'oh, you're right about how it's initialised, that didn't even cross my mind.

I'm more in favour of following the existing "convention" here, so I'll gladly move it outside of `__init__`.

sonic2kk:
Forgot to reply to this, sorry :sweat_smile: 

> Probably the ct info dialog is the best place to put it. I think the main ui would get to crowded if we put it there.

I agree.

> We could hide the game list and maybe rename "batch update" to "change global tool" or something

I like this, I wonder if it should be a separate widget altogether that we hide by default and show when necessary. That way we don't have to do any checks for a click event :-)

The UI portion of hiding the games list is something I'm thinking about. I think it's a good idea, but to add to it, it might be nice to replace it with a greyed-out bit of text centered relative to where the games list would be, that indicates that the tool is global. Or, if possible, we could put this in the games list.

I'm essentially imagining something similar to what KDE's Dolphin file manager does when a folder is empty, since in this case the games list is also empty.

image

Something like this, although with less empty space since our games list is smaller. The text in our case would say something like "Global Compatibility Tool".

We could even extend this idea to unused compatibility tools, displaying "No Games" in a similar fashion. If we choose to put both labels in the list, it would make for a consistent bit of UX: when a list is empty, we display a label, and for the global tool, we display a different label :-)

sonic2kk:
On my last point above, on a branch I tested out implementing something like this for "No games", here's how it could look visually. The greyed out text is achieved by "disabling" the label.

image

The exact sizing, font, etc could be figured out later, this is just a visual representation of what we could have for no games and then for the global tool as well. The UX is all wired up here too (refresh will enable/disable elements).

If we want something like this I can get a PR up for the branch I have now :-)

DavidoTek:
> On my last point above, on a branch I tested out implementing something like this for "No games", here's how it could look visually
> If we want something like this I can get a PR up for the branch I have now :-)

That looks good. I also like the smaller font (18).
If you already have it, feel free do to a PR :)

pupgui2/datastructures.py
sonic2kk: Apparently I forgot to do this, sorry! Should be moved now.

DavidoTek:
Thanks. Will be merged.

Changing the Batch update / Ctinfo dialog is probably something for a later PR.

> I'm more in favour of following the existing "convention" here (we also do this in other places too, so it's a general convention I mean), so I'll gladly move it outside of __init__.

I think I need to take a look at all those attributes when I find time and maybe revise it everywhere.
Python is a bit strange regarding instance attributes, for example, you also get this: https://stackoverflow.com/questions/472000/usage-of-slots
:laughing:

sonic2kk:
Figured out how to do this for Lutris, still got a few things to iron out though. It stores this information in its cache directory, under `versions.json`.

Will get a PR up in future for it, no idea when though :-)

PR #317, opened 2023-11-26
Roberta: Use Subclass of Luxtorpeda

sonic2kk:
This PR is the same as #311 but for the Roberta ctmod. Roberta is also a sister project of Luxtorpeda, and the changes here make Boxtron and Roberta identical right down to the number of lines save for differences in URLs, dependencies, and the name/description strings :-)


DavidoTek:
Thanks! :tada:

PR #321, opened 2023-11-26
Kron4ek Wine: Use fetch_project_releases and fetch_project_release_data

sonic2kk:
Mostly the same as #318 but for Lutris' Kron4ek Wine-Builds Vanilla.

The difference here is that, like for DXVK, we have an asset condition that we pass. We only want to fetch 64bit, non-staging releases, so the existing conditional we had in `__fetch_github_data` to check the asset name was replaced with a lambda that we pass to `fetch_project_release_data`. This takes care of filtering the assets.

I tested with and without this conditional, and we end up with the `-x86` version, so this should confirm that the asset check works for us here :slightly_smiling_face: 

Once again, there should be no user-facing change and everything should still work as expected. The releases dropdown should have no change from `main`, and the correct archive should still download and extract correctly.

image
image
image

<hr>

I have put up a lot of PRs recently that tackle smaller refactors, sorry if it's a lot to manage without much visible gain... But there is no hurry, this PR and the similar ones are just small code refactors and do not contain any absolutely urgent critical fixes :-) Just code consistency.

I think the remaining ctmods may require slightly more work, so I won't be flooding with more PRs for a while :sweat_smile:

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_kron4ekvanilla.py
+++ b/pupgui2/resources/ctmods/ctmod_kron4ekvanilla.py
@@ -9,7 +9,7 @@ import requests
 from PySide6.QtCore import QObject, QCoreApplication, Signal, Property
 
 from pupgui2.util import ghapi_rlcheck, extract_tar
-from pupgui2.util import build_headers_with_authorization
+from pupgui2.util import build_headers_with_authorization, fetch_project_release_data, fetch_project_releases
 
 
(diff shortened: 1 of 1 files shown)


DavidoTek:
Thanks!

> we have an asset condition that we pass

That's smart.

> I have put up a lot of PRs recently that tackle smaller refactors, sorry if it's a lot to manage without much visible gain

No need to apologize :smile: I think it's really great to clean the code a bit so we can keep adding more features without a hassle, and I really appreciate that.

PR #332, opened 2023-12-16
Shortcut Editor: Implement Search Functionality

sonic2kk:
## Overview
This PR adds a QLineEdit to function as a search bar to the Shortcuts Editor dialog, matching what we have for the CtInfo dialog and the Games List dialog. I have some other planned PRs for the Shortcuts dialog (spoiler :wink:) , so while I was working in here I realised search was missing and figured it would be straightforward enough to add.

image
image
image

## Implementation
Unlike the CtInfo and Games List tables, which use static labels, the Shortcuts Editor uses QLineEdits. So in order to search, we have to get the widget in the cell and then get the text from that widget, instead of just getting the TableItem at that cell and getting the text.

I believe you can populate tables with TableItems in Qt and have the option to double-click to edit them, but the Shortcuts Dialog uses QLineEdits and personally I think this is more flexible anyway.

### Always Shown Search Bar
This search bar is shown unconditionally. While this doesn't match the behaviour of the CtInfo and Games List, on those dialogs we have a search button (and Ctrl+F shortcut) because the search bar may conflict with existing widgets. On the CtInfo dialog, the Search Bar is toggled because it may overlap with the "Batch Update" button, and on the Games List, it conflicts with the Steam Running warning message. The search bar kind of has to be a toggle on those dialogs, but we have no such limitation on the Shortcuts Editor dialog, so we can unconditionally show the search bar. It would also be a bit more cluttered to introduce a search button either on the same bottom row of Add/Remove/Save/Close, above the games list with a new row, or a new row above/below the Add/Remove/Save/Close options.

Another spoiler, I'm investigating the feasibility / usefulness of adding a button to refresh the shortcuts list, so having a dedicated search button would create even more clutter on the bottom row (on top of not being as necessary as it is on the CtInfo/Games List).

I will also admit, I think it just looks nice to fill that space :sweat_smile: 

### No Tooltip
There is no tooltip for the search bar, unlike for the CtInfo and Games List dialog search bars which show "e.g. Half-Life 3" and "e.g. Team Fortress 2" respectively. Initially I had planned to add a tooltip for the Shortcuts dialog (of course with something equally as witty as Half-Life 3) but I had an idea to instead refactor the tooltip logic a bit. Instead of having a fixed tooltip, these search box tooltips could show a tooltip with a random game from the current game list. So instead of always saying "e.g. Half-Life 3", on each load of the dialog it would pick a random game name and generate the tooltip.

Instead of adding a tooltip here only to strip it out in another PR, I figured I'd leave it out. Then, if that PR is declined, we could add a tooltip here separately.

<hr>

If you want to make a release with the Shortcuts Dialog as-is then we can leave this PR for review until after the next release. It may make more sense to get an initial iteration out first and then work on enhancements afterwards. Even though this isn't out in a release yet, out of habit I use ProtonUp-Qt from my fork on `main` with `python3 -m pupgui2`, so I've been using the shortcut editor a little bit even so :-)

Thanks!

### DIFF
--- a/pupgui2/pupgui2shortcutdialog.py
+++ b/pupgui2/pupgui2shortcutdialog.py
@@ -48,6 +48,7 @@ class PupguiShortcutDialog(QObject):
         self.ui.btnClose.clicked.connect(self.btn_close_clicked)
         self.ui.btnAdd.clicked.connect(self.btn_add_clicked)
         self.ui.btnRemove.clicked.connect(self.btn_remove_clicked)
+        self.ui.searchBox.textChanged.connect(self.search_shortcuts)
 
     def prepare_table_row(self, i: int, shortcut: SteamApp):
         txt_name = QLineEdit(shortcut.game_name)
@@ -167,3 +168,11 @@ class PupguiShortcutDialog(QObject):
                 if sid not in self.discarded_shortcuts:
                     self.discarded_shortcuts.append(sid)
                 self.ui.tableShortcuts.cellWidget(i, 0).setStyleSheet('QLineEdit { color: red; }')
+
+    def search_shortcuts(self, text):
+        """ Search based on the shortcut name text (App Name on Row 0) in the QLineEdit widget on each row """
+        for row in range(self.ui.tableShortcuts.rowCount()):
+            if type(row_widget := self.ui.tableShortcuts.cellWidget(row, 0)) is not QLineEdit:
+                continue
+            should_hide: bool = not text.lower() in row_widget.text().lower()
+            self.ui.tableShortcuts.setRowHidden(row, should_hide)
--- a/pupgui2/pupgui2shortcutdialog.py
+++ b/pupgui2/pupgui2shortcutdialog.py
@@ -172,7 +172,11 @@ class PupguiShortcutDialog(QObject):
     def search_shortcuts(self, text):
         """ Search based on the shortcut name text (App Name on Row 0) in the QLineEdit widget on each row """
         for row in range(self.ui.tableShortcuts.rowCount()):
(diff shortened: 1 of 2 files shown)


sonic2kk:
Hmm, it just occurred to me that we may want a fallback tooltip anyway. I'll add one now.

DavidoTek:
Thanks!

> Always Shown Search Bar
> I will also admit, I think it just looks nice to fill that space 😅

I agree, this makes sense. No need to make it more complicated than necessary. We have the space anyway.

> I'm investigating the feasibility / usefulness of adding a button to refresh the shortcuts list

Not sure about that one. I think the user would close Steam beforehand anyway as changes to the shortcuts are only registered in Steam after a restart like with compatibility tools.

> Instead of having a fixed tooltip, these search box tooltips could show a tooltip with a random game from the current game list
> Hmm, it just occurred to me that we may want a fallback tooltip anyway. I'll add one now.

I guess the search bar wouldn't actually require a tool tip as it is quite self explainatory, but I'm fine with adding one.
Possibly adding the "dynamic tooltip" makes it a bit playful, I think that's also fine as ProtonUp-Qt is targeting a gaming audence :smile:

> If you want to make a release with the Shortcuts Dialog as-is then we can leave this PR for review until after the next release

Yeah, I'm thinking of doing a release with all the features for a while now.
I think it this PR can already be merged now.

---

Tested it, works as expected.
I changed it a bit, so a user can also search for the executable name. I think that can be useful in some cases and shouldn't break anything.
The only situation where something is changed is when the searched game name is similar to the user's name which may be in the executable path in some cases.

sonic2kk:
> I changed it a bit, so a user can also search for the executable name

That's a useful change, I like it 😄  Thanks!

PR #333, opened 2023-12-16
CtInfo, Games List: Use random game name from list in tooltip

sonic2kk:
## Overview
This PR allows for dynamically setting the search box tooltip for the CtInfo Dialog and Games List. On `main`, we hardcode these in the UI file to "Half-Life 3" and "Team Fortress 2" respectively, but now we attempt to take a random game from the list of games and use that game name in the tooltip. For example, if the games list has "GUILTY GEAR -STRIVE-", the tooltip has a chance of selecting that game and so would say "e.g. GUILTY GEAR -STRIVE-".

If the game list length is zero, or if we happen to select a game that somehow doesn't have a name, we will fall back to the default tooltip defined in the UI file. There could *maybe* be an edge case for Heroic/Lutris if someone only had one game using a tool and they manually made the game name blank, in such a case we would use our fallback. But that is astronomically unlikely to happen, and we don't show the search bar when the games list length is zero, so it's pretty unlikely that a user will ever see the fallback label. Still, since we can't `random.choice` on an empty list, we had to handle this case anyway :-)

Here are some screenshots:

### Steam CtInfo
image

### Heroic Games List
image

The tooltip updates on refresh too, so that if a game is added/removed in that time, we won't display names of games that aren't there.

This is implemented for Steam, Lutris, and Heroic. If we can't find a tooltip (i.e. the name is blank for some weird reason), we fall back to the tooltip define in the UI file.

Translation should not really matter here, since it will either use a translated tooltip in the UI file (if that's even applicable) or it will use the game name, which doesn't need a dedicated translation since it's simply pulling from the game name we already store.

I don't think this will add any considerable overhead, since we just pick a random element from the list of games that we already store. We aren't generating the games list again just to do this, we're re-using the list we already have.

<hr>

This just adds a bit of fun context to the search field that I thought would be cool to add :-)

### DIFF
--- a/pupgui2/pupgui2ctinfodialog.py
+++ b/pupgui2/pupgui2ctinfodialog.py
@@ -1,12 +1,13 @@
 import pkgutil
 import os
+import random
 
 from pupgui2.constants import STEAM_APP_PAGE_URL
-from pupgui2.datastructures import BasicCompatTool, CTType
+from pupgui2.datastructures import BasicCompatTool, CTType, SteamApp, LutrisGame, HeroicGame
 from pupgui2.lutrisutil import get_lutris_game_list
 from pupgui2.pupgui2ctbatchupdatedialog import PupguiCtBatchUpdateDialog
 from pupgui2.steamutil import get_steam_game_list
-from pupgui2.util import open_webbrowser_thread
+from pupgui2.util import open_webbrowser_thread, get_random_game_name
 from pupgui2.heroicutil import get_heroic_game_list, is_heroic_launcher
 
 from PySide6.QtCore import QObject, Signal, QDataStream, QByteArray
@@ -15,7 +16,7 @@ from PySide6.QtUiTools import QUiLoader
 from PySide6.QtCore import Qt
 from PySide6.QtGui import QShortcut, QKeySequence
 
-from typing import List
+from typing import List, Union
 
 
 class PupguiCtInfoDialog(QObject):
@@ -26,7 +27,7 @@ class PupguiCtInfoDialog(QObject):
         super(PupguiCtInfoDialog, self).__init__(parent)
         self.parent = parent
         self.ctool = ctool
-        self.games = []
+        self.games: List[Union[SteamApp, LutrisGame, HeroicGame]] = []
         self.install_loc = install_loc
         self.is_batch_update_available = False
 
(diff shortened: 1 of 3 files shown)


sonic2kk:
I should also mention that, if we *really* want, we can update this each time the search bar is summoned 😉

sonic2kk:
I just noticed the change to use `self.games` here for Lutris and Heroic's game list generation fixes a regression introduced in #319 + #329. We check when to display the 'No Games' label based on the length of `self.games`, but only the Steam logic actually sets  this; Lutris and Heroic use separate variables for some reason.

Main branch (note how the number of games is correctly counted as `2`):
image

This PR:
image

This PR accidentally fixed that by setting `self.games` :sweat_smile:

DavidoTek:
Thanks!

> If the game list length is zero, or if we happen to select a game that somehow doesn't have a name, we will fall back to the default tooltip defined in the UI file
> The tooltip updates on refresh too, so that if a game is added/removed in that time, we won't display names of games that aren't there.

Very thought through!

> Translation should not really matter here, since it will either use a translated tooltip in the UI file (if that's even applicable) or it will use the game name, which doesn't need a dedicated translation since it's simply pulling from the game name we already store.

In theory we could wrap a `self.tr` around `'e.g. {GAME_NAME}'` like this: `..setToolTip( self.tr('e.g. {GAME_NAME}') .format(...) )` as `e.g.` could be translated to other languages (e.g., `z.B.` in German, `p.ej.` in Spanish, ...)
(EDIT: DONE)

> I don't think this will add any considerable overhead, since we just pick a random element from the list of games that we already store. We aren't generating the games list again just to do this, we're re-using the list we already have.

Looks fine and `random.choice` should be quite fast.

> This just adds a bit of fun context to the search field that I thought would be cool to add :-)

Yeah, makes it a bit more playful. I like it :smile:

> I should also mention that, if we really want, we can update this each time the search bar is summoned 😉

Hmm, we might actually want to do this.
Even though there shouldn't be any noticable overhead, executing as few functions as possible on startup by moving them to the button functions is welcome, a bit like Lazy Loading. That would also give us the "new random text when the search bar is summoned" as a side effect :smile:
(EDIT: DONE)

> This PR accidentally fixed that by setting self.games 😅

Good you noticed that.
It's amazing how much is overlooked, even when a PR feels well tested :smile:

PR #334, opened 2023-12-16
Shortcut Editor: Various UIX Tweaks

sonic2kk:
## Overview
This PR makes a few UIX tweaks to the Shortcut Editor dialog:
- Increase size of Shortcut Editor dialog to match Games List, both show games and it makes sense to give them an equal size imo
- Increase width of App Name column, open to discussion on adjusting this and/or adjusting the widths of other columns to give the information more breathing room! On the other dialogs, the App Name is given the most sizing priority, which is why I gave it the biggest size here.
- Always put App Name LineEdit cursor at beginning, so the start of the game name is shown instead of the end.
- Select all text in LineEdit when initially focused with mouse, this allows quick editing and also moves the cursor to the end of the LineEdit, such as when clicking on the game name, allowing a quick way to see the beginning and end of the App Name.
    - Tab Focus behaviour is preserved.
- Allow unfocusing the LineEdits by clicking on the table background or the dialog background.
- When tabbing back out, the position of the App Name cursor is reset to the beginning.
- Add tooltips to LineEdits with their values, making it easier to see their full values.
- Add placeholder text to LineEdits with their original values, so if the value is cleared entirely, the original value can still be seen
    - If a LineEdit value is cleared entirely (i.e. `text` parameter in `txt_changed` is a falsey empty string), the tooltip text falls back to the placeholder text.

To help illustrate the changes, I made some usage comparison videos against `main` and this PR.

### `main` Branch
[pupgui2 shortcut editor - main.webm](https://github.com/DavidoTek/ProtonUp-Qt/assets/7917345/6ca71929-fb00-45e2-bf90-8dac04601bb2)

### This PR
[pupgui2 shortcut editor - PR.webm](https://github.com/DavidoTek/ProtonUp-Qt/assets/7917345/fabe8cb5-6bdf-438a-9d4c-026797d1ea92)

Despite the changes, this should still be compatible with #332.

## Implementation
To accomplish the focusing changes I wanted with any amount of flexibility, I created a custom subclass of `QLineEdit` that overrides various the focus events. We track when the line edit is initially focused and use this to know to only select the entire LineEdit text on the first focus click. This allows another click to de-select everything, and avoids everything being selected again on any other mouse click event.

One of the things that tripped me up here is the order these events are called, so I'll give some background on that. When you click on the QLineEdit, `focusInEvent` is called. However directly after that, `mousePressEvent` is called to process the click. If you try to select all the text in the line edit on `focusInEvent`, the click event processed directly after in `mousePressEvent` will override this. But we have to use `focusInEvent` to preserve the selection behaviour on Tabbing into the LineEdit cell. So we have to do two things:
1. Highlight all text in the LineEdit on the *first* mouse leftclick focus event
2. Highlight all text on the Tab focus event (this can only happen once, since to tab into the cell again, you have to tab out and then back in; you can't use tab to focus the same cell twice without first switching cells)

To accomplish this, we call the helper method I created on the widget called `focusWithTextSelection` to highlight the text in the LineEdit in `focusInEvent` if the focus reason is a tab focus event. This means we ignore mouse focus events, since we shouldn't call this helper method twice. Since we know `mousePressEvent` will handle the mouse click event directly afterwards on a mouse click focus event anyway, we let that method override handle that case. Finally, we use a boolean to track if the focus was already given, which allows us to only select all text on the *first* focus (meaning a 2nd click won't keep all the text highlighted) and also avoids a case where if we tab into the cell, clicking will move the cursor and de-select all of the text, instead of causing another call to `focusWithTextSelection`.

Calling `super` on these methods is pretty important, otherwise you end up with whacky focus/duplicated cursor issues!

## Motivation
Originally, I just wanted to move the cursor for AppName back to the beginning so that I could see the start of the App Name instead of the end. I eventually fell down a rabbit hole of making some other usability changes that I happened to notice along the way.

<hr>

Of course all of this is just my own personal tastes and based on my own expectations, so to you and others these behaviours may be unexpected. I'm open to all discussion on any feedback, or on the other hand, other UX changes you think would fit the scope of this PR.

Thanks!

### DIFF
--- a/pupgui2/pupgui2shortcutdialog.py
+++ b/pupgui2/pupgui2shortcutdialog.py
@@ -97,6 +97,11 @@ class PupguiShortcutDialog(QObject):
         txt_path.setToolTip(shortcut.shortcut_startdir)
         txt_icon.setToolTip(shortcut.shortcut_icon)
 
+        txt_name.setPlaceholderText(shortcut.game_name)
+        txt_exe.setPlaceholderText(shortcut.shortcut_exe)
+        txt_path.setPlaceholderText(shortcut.shortcut_startdir)
+        txt_icon.setPlaceholderText(shortcut.shortcut_icon)
+
         txt_name.editingFinished.connect(lambda i=i: self.txt_changed(i, 0))
         txt_exe.editingFinished.connect(lambda i=i: self.txt_changed(i, 1))
         txt_path.editingFinished.connect(lambda i=i: self.txt_changed(i, 2))
--- a/pupgui2/pupgui2shortcutdialog.py
+++ b/pupgui2/pupgui2shortcutdialog.py
@@ -55,6 +55,8 @@ class PupguiShortcutDialog(QObject):
         txt_path = QLineEdit(shortcut.shortcut_startdir)
         txt_icon = QLineEdit(shortcut.shortcut_icon)
 
+        txt_name.setCursorPosition(0)
+
         txt_name.editingFinished.connect(lambda i=i: self.txt_changed(i, 0))
         txt_exe.editingFinished.connect(lambda i=i: self.txt_changed(i, 1))
         txt_path.editingFinished.connect(lambda i=i: self.txt_changed(i, 2))
--- a/pupgui2/pupgui2shortcutdialog.py
+++ b/pupgui2/pupgui2shortcutdialog.py
@@ -11,6 +11,25 @@ from pupgui2.steamutil import get_steam_shortcuts_list, write_steam_shortcuts_li
 from pupgui2.util import host_path_exists
 
 
+class ShortcutDialogLineEdit(QLineEdit):
+
+    def __init__(self, parent=None, default_cursor_position: int = -1, *args, **kwargs,):
+        super(ShortcutDialogLineEdit, self).__init__(parent, *args, **kwargs)
+
+        self.default_cursor_position = default_cursor_position
+        if self.default_cursor_position >= 0:
+            self.setCursorPosition(self.default_cursor_position)
+
+    def focusOutEvent(self, arg__1):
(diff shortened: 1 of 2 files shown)


sonic2kk:
Noticed the MP4 videos aren't embedding properly, I'll convert them to webm and reupload.

sonic2kk:
I was just looking into allowing sorting perhaps as part of this PR, but it seems like it isn't that simple. To accomplish this we may have to switch to using `QTableListItem`s like we did in #137, though my searching led me to something called `QTableListView` which is apparently better when dealing with cells that contain one (or more) *widgets*. I really don't have any idea how that all works right now, but I wanted to call out that I'm aware sorting isn't handled in this PR.

What I think I will do is learn a bit more about what ways to go about implementing this kind of sorting, and if there is a better approach than we took for the other TableWidgets in other dialogs, I will open a PR. This shortcuts dialog could act as a testing ground for a better implementation.

So yeah, sorting is missing from this PR, but hopefully more to come in a future PR :-)

sonic2kk:
I wonder if there's potential to show the Non-Steam Game AppID here... It's probably not all that useful but maybe we could add it on the tooltip, in brackets on a newline (to avoid getting in the way of the AppName).

sonic2kk:
For health reasons I will be taking a break from working on public projects, so I won't be around for a while to review any PR feedback. I can pick this up again when I'm available, or if there is work you or others would like to do that builds on this, feel free to make any necessary changes and merge. :slightly_smiling_face:

DavidoTek:
Get well soon and take time for youself!

I'll review this long waiting PR in the next weeks and if neccessary do corrections myself.

Thanks for the work you put in ProtonUp-Qt and the other projects!

PR #336, opened 2023-12-17
CtInfo: Add Button to Mark Compat Tool as Global

sonic2kk:
Should fully round off #254 (except for replacing unused/global with a "badge" which is a general and more involved change not specific to global ctool alone).

Marked this PR as draft while I work out how to write to the `config.vdf` file, and also figure out how to update the Tool List on the Main Window. I think a signal is needed similar to what we do for `batch_update_complete`.

## Overview
This PR adds a button on the CtInfo dialog that allows for marking the current compatibility tool as the Global one. This is only shown for the Steam Launcher, when the current compatibility tool is not already global, and only when the current ctool type is `CUSTOM` so that we don't allow setting an incompatible compatibility tool type as the default (i.e. Proton EasyAntiCheat Runtime).

Below are some screenshots showing the different behaviours:

image
image
image
image

We may want to make some UI tweaks to the button, I'm not super sold on the button text or positioning, but no good alternatives have come to mind as of yet. I put it beside Batch Update to keep it further out of the way to avoid accidental clicks. Batch Update is a reasonably deliberate action, but it could be easy to accidentally click this button if it was beside what might be more commonly used buttons such as the Close button, or Search/Refresh.

## Considerations
We also may want to add a warning dialog, to help in the case of accidental clicks. There is no way to revert this action from ProtonUp-Qt, the only way to undo it is to go into the Steam Client.

I'm not sure what happens if a user tries to change this while the Steam Client is running. I suspect it does the same thing as updating other CompatToolMapping values, and any values updated outside of Steam are overwritten when the Client closes. But there also isn't a great way for us to show the warning on the CtInfo dialog either. We *could* disable the button when Steam is running, but it might not be obvious that it's disabled because the Steam Client is running.

I think we have two options to deal with this:
1. Show a dialog if the user tries to do this when the Steam Client is running. If we're going to show one to confirm the action of marking the tool as global anyway, we could probably just add another dialog to say "Close the Steam Client before updating the compatibility tool."
    - The problem is that we'd have two different paths to manage here, which could add a bit of complexity.
3. Add this warning message to the warning dialog before marking a tool as global, either conditionally if the Steam Client is running or just always show it. If we always display it then it would be a very simple solution, but would not match the other warnings that only show up when the Steam Client is open, and a user may mistakenly think that their changes will not apply. Only showing the warning conditionally would mean the user only sees it when relevant. 
    - The problem with this approach is that it's possible a user will miss it. If we always display this warning message too, it would be easy for a user's eyes to gloss over it. Conditionally disabling the button to allow the user to proceed would also be a bit complex, like if we were to prevent them from applying the changes when the Steam Client is running, we'd have to make a path to check for that.

Complexity isn't a dealbreaker or anything per se, I'm just looking for the best approach, and those are my thoughts so far.

## Remaining Work
I still need to wire up the functionality, updating the global ctool VDF dictionary should be working, but it isn't writing it out yet. But I have some implementation thoughts that I wanted feedback on, so I figured this was in a reasonable state to open as draft :-)


sonic2kk:
Having taken some more time to think on it, I'm really not sure about that position for the "Set as Global" button. To be honest, I'm envisioning something more like this (mockup made in GIMP, not code):

edit mockup

Maybe for the form layout we'd have to add a 3rd column for the 1st row and put the button there, but no idea how feasible that is...

Any thoughts about the UI portion of this, I'd really appreciate! Actually editing the VDF file should be straightforward, my main concerns with this PR are to do with making the UI part actually "fit" well enough :-)

sonic2kk:
For health reasons I will be taking a break from working on public projects for a while. I won't be around to continue this work for a bit, likely a couple of weeks at least.

If someone else wants to pick this up in the meantime please go ahead, either by salvaging this or superseding it with another PR. :slightly_smiling_face:

sonic2kk:
I took another quick look at this branch this evening and I'm not happy with where I've placed the button. I tried to find some way to fanagle the button placement into where I showed in the screenshots, but I wasn't able to. Probably the only way to do that would be with some kind of Grid layout -- Qt has a Grid layout but I'm not sure it could be used in this way? I wasn't able to find his to set widget positions from Qt Designer but maybe I'm just blind 🙃

Even so with the above it might alter the uniform look of the UI. Since a button is taller it might make all the cells on a given row taller.

Having some kind of Horizontal layout that contains the FormLayout and then inserting a Vertical layout inside it was another approach I tried, but it still didn't look right... If there was some way to get this to fit the design I had in mind, we could expand it to add a "Copy to Clipboard" button for the compat tool path.

<hr>

As I'm kinda stumped right now on the UI portion, I think I'll take a quick look at the rest of the implementation again and mark it out of draft, and we can discuss how to implement this on the UI side. At worst, we could add another button along with the refresh and search buttons, but I'm conscious of accidental presses for such an action. I guess a warning dialog is a possible solution...

sonic2kk:
That rebase was scarier than it needed to be, it had conflicts across commits which was a new experience for me, but it seems to have went smoothly!

During that experience I had an idea that if we wanted to put the global compatibility tool as global option up with the search/refresh buttons, we could do it like this:
- Make it a button with an icon, like the existing buttons on that row, and perhaps use a "star" icon of some kind? I think that might convey the behaviour. We will also note in the tooltip that this button marks a tool as global.
- When the button is clicked, we can show a confirmation dialog. This would give a bit more information. We might want to make this dialog flexible so we can use different strings for different launchers (I've mostly implemented internally tracking the global compatibility tool for Lutris, pending some cleanup before PR).
    - Optionally we could have a "Don't show this again" for users who do this often, although I can't imagine this gets done often.

I moved and altered the button (but didn't commit the changes), here's what it might look like:

image

It is selected by default, but we should be able to fix that.

Even if it might ultimately be a little futile, we should help the user to be careful, at least as much as is reasonably possible without being intrusive. This is a "destructive" operation in the sense that there is no way to *revert* this change, the user has to know what tool they had selected as global before and then manually mark it again, or select it from the Steam Client if their previous selection was not a custom compatibility tool.

sonic2kk:
Ah, it seems I left the writing to the VDF file commented out. I even left a comment saying I was unsure if it would work, but it does in fact work. I threw caution to the and tested and it correctly marked GE-Proton8-28 as global with this PR. It also correctly updates on the Main Menu that the tool is global!

I pushed the uncommented line in b2ded73.

Lookking at the code again I wonder if we could create an abstracted util function to call `vdf.dump`. Instead of calling `vdf.dump(d, open(config_vdf_file, 'w'), pretty=True)` in three different places (`steam_update_ctool`, `steam_update_ctools`, and now `set_global_compat_tool`) and having to remember how to call it, we could create a small helper function like `update_vdf_file(filepath = config_vdf_file, contents = d)`. Really minor nitpick but just something I thought of.

pupgui2/steamutil.py
```diff
@@ -791,4 +791,29 @@ def determine_most_recent_steam_user(steam_users: List[SteamUser]) -> SteamUser:
...
+            d = vdf.load(open(config_vdf_file))
+            c = get_steam_vdf_compat_tool_mapping(d)
+
+            # Create or update the '0' CompatToolMapping entry
+            if not c.get('0', {}):
+                c['0'] = { 'name': ctool.get_internal_name(), 'config': '', 'priority': '75' }
+            else:
+                c['0']['name'] = ctool.get_internal_name()
+
+            vdf.dump(d, open(config_vdf_file, 'w'), pretty=True)
```
sonic2kk, line 814: Looking at the code again I wonder if we could create an abstracted util function to call `vdf.dump`. Instead of calling `vdf.dump(d, open(config_vdf_file, 'w'), pretty=True)` in three different places (`steam_update_ctool`, `steam_update_ctools`, and now `set_global_compat_tool`) and having to remember how to call it, we could create a small helper function like `update_vdf_file(filepath = config_vdf_file, contents = d)`. Really minor nitpick but just something I thought of.

pupgui2/steamutil.py
```diff
@@ -791,4 +791,29 @@ def determine_most_recent_steam_user(steam_users: List[SteamUser]) -> SteamUser:
...
+        except Exception as e:
+            print(f'Error, could not update Global Steam compatibility tool to {ctool.displayname}: {e}, vdf: {config_vdf_file}')
+            return False
+
+        return True
```
sonic2kk, line 819: We could probably move this inside of the `try` block, since it shouldn't be able to return `True` if it catches an exception. But it seems like "convention" to put these kinds of returns at the end, such as:

https://github.com/DavidoTek/ProtonUp-Qt/blob/89bdbcf9843210c6e479a2834bde80ed7637c7b4/pupgui2/steamutil.py#L636-L639

Happy either way though :smile:

pupgui2/steamutil.py
```diff
@@ -791,4 +791,29 @@ def determine_most_recent_steam_user(steam_users: List[SteamUser]) -> SteamUser:
...
     print('Warning: No Steam users found. Returning None')
-    return None
\ No newline at end of file
+    return None
+
+def set_global_compat_tool(ctool: BasicCompatTool, steam_config_folder):
+    """ Update the global compatibility tool to the ctool parameter defined in config.vdf CompatToolMapping (AppID '0') """
+
+    if ctool.ct_type == CTType.CUSTOM:
+        config_vdf_file = os.path.join(os.path.expanduser(steam_config_folder), 'config.vdf')
```
sonic2kk, line 800: It should be a separate PR I think but, we define this a lot in `steamutil`, I wonder if we should define it (and `libraryfolders.vdf` in `constants.py`.

I can't remember if we let the user select their current Steam user, but if not, the path to `shortcuts.vdf` would also fit here, since it won't change during execution.

We've made changes like this has been made in the past (#281) so might be worth considering :slightly_smiling_face:

pupgui2/pupgui2ctinfodialog.py
```diff
@@ -29,6 +29,7 @@ def __init__(self, parent=None, ctool: BasicCompatTool = None, install_loc=None)
         self.games: List[Union[SteamApp, LutrisGame, HeroicGame]] = []
         self.install_loc = install_loc
         self.is_batch_update_available = False
+        self.is_mark_global_available = (self.ctool.ct_type == CTType.CUSTOM and not self.ctool.is_global) and self.install_loc.get('launcher') == 'steam' and 'vdf_dir' in self.install_loc
```
sonic2kk, line 32: This will need revisited to account for other launchers, we could probably make a function for this instead. The first part of the check is fine, but the 2nd part outside the brackets is very Steam-specific. When we expand to Lutris, the check might turn into this, which would be ugly:

```python
(self.ctool.ct_type == CTType.CUSTOM and not self.ctool.is_global) and (self.install_loc.get('launcher') == 'steam' and 'vdf_dir' in self.install_loc) or self.install_loc.get('launcher') == 'lutris'
```

Instead we could make a function, maybe like this?

```python3
def is_mark_global_available(self):
    allowed_launchers = [ 'steam', 'lutris' ]  # Only allow marking global for these launchers

    # Only allow marking global for compatibility tools and not runtimes etc
    if self.ctool.ct_type != CTType.CUSTOM:
        return False

    # Exit early if launcher is not on our list of compatible/allowed launchers
    if self.install_loc.get('launcher') not in allowed_launchers:
        return False

    # Only allow marking global for Steam if we have the VDF directory path
    if self.install_loc.get('launcher') == 'steam' and 'vdf_dir' not in self.install_loc:
        return False
    # Could put other conditions for other launchers in an elif here, i.e. check for Lutris and ensure a specific path exists

    return True  # All other checks passed so default to true
```

pupgui2/steamutil.py
DavidoTek: We can do that.

Maybe it's time to also rename the variables `c`, `d`. Not sure how to call them though, that's why I didn't give them good names in the first place.

pupgui2/steamutil.py
DavidoTek: We should probably add a `return 1` to the `except` block as it currently will `return 0` if an exception occurs. That's not the intended behaviour as far as I can remember.

pupgui2/steamutil.py
DavidoTek: Not sure if we can do that in constants as the `steam_config_folder` variable may change.

We could create a special function like `get_special_steam_file(steam_config_folder: str, special_file_type: str, extra_arg=None)` that accepts the folder and for special file type "libraryfolders"/"shortcuts"+extra_arg=user_id/"config". Not sure if that would give any benefit.
Or do you mean just defining something like `LIB_VDF_BLAH = 'libraryfolders.vdf'`?

Is that how you intended it or is there another way?

pupgui2/pupgui2ctinfodialog.py
DavidoTek: Seems reasonable. Definitively better than a one-liner. :smile:

sonic2kk:
Sorry, I totally missed the replies here... I'll take a look into it now.

pupgui2/steamutil.py
sonic2kk: Yeah, that looks right. We `return 1` in other cases when we exit early, like here:

https://github.com/DavidoTek/ProtonUp-Qt/blob/7b673a7531781f7b01ef94d6c4ad49f535d4a332/pupgui2/steamutil.py#L601-L612

pupgui2/steamutil.py
sonic2kk: This is embarrassing, but I don't quite remember where I was going with that.

Although I think having to continuously do `os.path.join(os.path.expanduser(steam_config_folder), 'config.vdf')` is a bit repetitive, having a dedicated function for it would abstract away the need to join the paths manually and also run the appropriate `expanduser` to make sure we have the right path, those kinds of implementation details would be hidden away in `get_special_steam_file`.

It's something for another PR for sure I think and needs a bit more thought from my end :-)

pupgui2/pupgui2ctinfodialog.py
sonic2kk: I'll replace the line highlighted in this thread with `is_mark_global_available` :-)

pupgui2/steamutil.py
sonic2kk: Pushed the change

pupgui2/pupgui2ctinfodialog.py
sonic2kk: I'm going to tentatiively put this method in `util.py`, as if we ever allow marking a tool as global anywhere else outside of the ctinfo dialog, we would need to use this function.

I have not had a chance to test this very much yet, but will do so over the weekend when I have some time.

pupgui2/steamutil.py
sonic2kk: I think at the time I had a good idea on how to break it out, but it's been a while since I've been thick in the codebase, so I'll take another look when I have some time to see how best to break this out.

Having this separated out might also be good for unit testing, hint-hint-wink-wink :smile:

sonic2kk:
Addressed most of the feedback. Once done, the UI may still need a bit of consideration. I'm not overly happy with the current placement. The more I look at my mockup with the star util button the more I like that one, but I also recognise the shortcomings that it is not totally visually obvious and may be easy to press.

Another possibility may be to hide the "Set as Global" button when the Search bar is present. This would mean that a user couldn't search for a game, keep that list on screen, and then press "Mark as Global". The user would have to close the search bar before they can press the button, if the button is kept in its current position. It may not be clear to the user where the button went, or it may get reported as a bug that the button vanishes if it is not intuitive behaviour.

Basically, the actual updating of the global compat tool works, but I'm still not entirely sure on the UIX for it. Even all these months later :sweat_smile: 

<hr>

Regardless of the approach to the UI though, I think one thing I would like to include (either here or in a follow-up PR soon after and definitely before this gets into a release) is a confirmation dialog for the user to let them know that this will mark the tool as global for all compatibility tools. We should also note that for the change to take effect, the Steam Client must be closed beforehand (since this modifies `config.vdf` which gets overwritten when Steam is closed).

The wording on this dialog should be somewhat dynamic, since this will be used for Lutris as well (hopefully in the near-ish future, a couple months ago I was fiddling around with some global-tool related stuff for Lutris). The warning about closing Steam, for example, shouldn't be shown. We can probably just set custom messages based on `install_loc.get('launcher')`.

We should be able to add the dialog reasonably easily I think, we should be able to use `show_msgbox_question` the same way we do for SteamTinkerLaunch (or maybe there's another approach we have that I'm forgetting, it's been a while...)

Maybe we would also want to add a "Don't show me this again" type checkbox on the dialog (unticked by default)? Not sure if that's worth it or if it could cause more problems for the UX.

<hr>

So three things remain for this PR so far:
1. Address feedback around splitting out some of the VDF file writing into a separate function with `update_vdf_file` as a separate `steamutil` function.
2. Discuss the current UI and how it could be improved, or if it is fine to leave it the way it is.
3. Add a confirmation dialog before writing any changes to `config.vdf`.

sonic2kk:
I had some more thoughts on breaking the VDF stuff into a separate function, and to cut down on reviewing too much of a refactor, I would like to leave it for a separate PR if that's okay. What you said about giving the variables better names than `c` and `d` is something that could be tackled in a separate PR as well, a general refactor of how we do our VDF interactions and how best to cut down some of the duplication.

DavidoTek:
> Address feedback around splitting out some of the VDF file writing into a separate function with update_vdf_file as a separate steamutil function.
> I had some more thoughts on breaking the VDF stuff into a separate function, and to cut down on reviewing too much of a refactor, I would like to leave it for a separate PR if that's okay. What you said about giving the variables better names than c and d is something that could be tackled in a separate PR as well, a general refactor of how we do our VDF interactions and how best to cut down some of the duplication.
> Having this separated out might also be good for unit testing, hint-hint-wink-wink 😄

Okay, that's fine for a separate PR. 

> Discuss the current UI and how it could be improved, or if it is fine to leave it the way it is.
> Another possibility may be to hide the "Set as Global" button when the Search bar is present

Seems fine to me as it is.
I guess hiding the "Batch update" buttons makes sense at it is applied to all games shown.
The mark global button is independent of that.
What we could do is to place if more to the left so it does not jump to the right when the search bar is displayed.

> Add a confirmation dialog before writing any changes to config.vdf.
> Regardless of the approach to the UI though, I think one thing I would like to include (either here or in a follow-up PR soon after and definitely before this gets into a release) is a confirmation dialog for the user to let them know that this will mark the tool as global for all compatibility tools.

We can also move that to a follow-up PR.

> we should be able to use show_msgbox_question the same way we do for SteamTinkerLaunch (or maybe there's another approach we have that I'm forgetting, it's been a while...)

I think we can just use the QMessageBox here. The reason we don't use it for the compatibility tools is that they are running on a separate thread from the main ui.

---

> Basically, the actual updating of the global compat tool works, but I'm still not entirely sure on the UIX for it. Even all these months later 😅

Should/can we merge it for now though?

TODOs for future PRs would be:
- [ ] VDF Stuff
  - [ ] "breaking the VDF stuff into a separate function"
  - [ ] "giving the variables better names"
  - [ ] "a general refactor of how we do our VDF interactions and how best to cut down some of the duplication."
- [ ] Maybe improve UI by moving the button to the left
- [ ] Add a confirmation dialog

sonic2kk:
Makes sense to me to keep all those for future PRs!

> Should/can we merge it for now though?

Not just yet, in testing I had hoped the answer would be yes, but there are a couple of things I want to address:
1. The binding for `btnMarkGlobal` is incorrect as it's inside the block that also checks for `'Proton' in self.ctool.displayname`, when it should really be bound if `self.is_mark_global_available`. I have addressed this locally, not committed yet.
2. "Set as Global" is highlighted by default, meaning if you open the dialog and press a button like Space or Enter, the "destructive" action of setting the global compat tool will be performed. Either "Close" or something else should be highlighted by default (or better yet, no button hightlighted by default, but I think I looked into this a while ago and ran into trouble... wonder if it's feasible to simply give focus to the list if present, and in cases where it isn't present, focus the close button).
3. Luxtorpeda for some reason saves as `%name%` when set as the global compat tool and I am not sure why. All other compat tools that I have tested work fine (Proton-tkg, GE-Proton8-32, Steam-Play-None, SteamTinkerLaunch also correctly sets as `Proton-stl`).

There is another UX issue I found but that would probably depend on what direction we go with for #350. After marking a compat tool as global, the list doesn't update. It should go from whatever state it was in (showing a games list, or displaying "No games") to displaying "Tool is Global." The list on the main menu will update to say a tool is "(global)" though.

I'm unsure if the UI should update or if we should close the games list, similar to how to go with #350.

sonic2kk:
The Luxtorpeda name issue appears to be an upstream bug that was already fixed a few months ago, my bad: (luxtorpeda-dev/luxtorpeda#271). Updating has fixed the issue.

When Luxtorpeda is built it looks like it sets the name at build time in the `compatibilitytool.vdf`, based on its [`compatibilitytool.template`](https://github.com/luxtorpeda-dev/luxtorpeda/blob/a67fd1cad3d5e3d5af9fb13f63691ceda5c06846/compatibilitytool.template) file and substituting the `%name%` and `%display_name%` values with those defined in `post_build.rs` (as seen in the linked PR diff).

The name was actually being set as `%name%` in the `compatibilitytool.vdf` file, so ProtonUp-Qt was fetching it properly.

So the Luxtorpeda issue is solved, and I pushed a fix for the incorrect binding logic in 6fd4d26.

The remaining issue I want to tackle in this PR before merging is the focusing issue.

DavidoTek:
> Not just yet, in testing I had hoped the answer would be yes, but there are a couple of things I want to address:

Okay.

> I'm unsure if the UI should update or if we should close the games list, similar to how to go with #350.

Does it actually change anything about the game list inside the ct info dialog? I think it only shows the games which have the tool specifically selected and not all games which use it because it is a global tool, or am I confusing something?

sonic2kk:
When you set the tool as global, currently nothing changes in the dialog. If you close and re-open the dialog, it shows the "Tool is Global" message (although I'm considering a way to add a button to toggle this, thinking ahead to Lutris global tools because of how Lutris manages tools it can be useful to see the games anyway).

It may be unintuitive to have a games list displayed after marking as global, and then when you open the dialog again it'll show the "Tool is Global" label. Switching to this label immediately after may be more intuitive is what I'm thinking :-)

After that I think this is in a mergable state if you're still happy with the placement of the button to mark a tool as global 😄

sonic2kk:
A user would have to select a tool to mark it as global. I'm not sure where the button would go on the Main Menu either. This is also not something to have "prominently displayed" I don't think.

The tool is displayed as global on the Main Window tool list the same way "unused" and the version of tools are shown. A global indicator is also displayed on the compat tool info dialog.

There's no reason from the code side why this couldn't go on the Main Window though, it's just a case of moving the button, I'm just personally not in favour of having it there 🙂

sonic2kk:
We wouldn't need to hide Anti-Cheat runtimes, we can just disable the button to mark as global for those runtimes. We already have logic for this on the ctinfo dialog, there is a function we call to check if marking a given compat tool as global is available for the current compat tool, and we don't enable this for runtimes including the anti-cheat runtimes.

pupgui2/util.py
```diff
@@ -892,3 +892,24 @@ def get_random_game_name(games: List[Union[SteamApp, LutrisGame, HeroicGame]]) -
...
+    # Only allow marking global for compatibility tools and not runtimes etc
+    if ctool.ct_type != CTType.CUSTOM:
+        return False
+
+    # Exit early if launcher is not on our list of compatible/allowed launchers
+    if install_loc.get('launcher') not in allowed_launchers:
+        return False
+
+    # Only allow marking global for Steam if we have the VDF directory path
+    if install_loc.get('launcher') == 'steam' and 'vdf_dir' not in install_loc:
```
sonic2kk, line 909: Wonder if we need to replace this with `util#is_valid_launcher_installation`? I'd be worried about this ending up being a bit expensive with all of the filesystem checks...

DavidoTek: `util#is_mark_global_available` is only called once at the start of ProtonUp-Qt as far as I can see.
In theory, using `util#is_valid_launcher_installation` could slow down the start but (I hope!) Python's `os.path.exists` call only take a few microseconds (50µs on my pc with a ~30 byte path).

Should be fine.

pupgui2/steamutil.py
sonic2kk: We can do this refactor in a separate PR.

pupgui2/util.py
sonic2kk: Sounds good, we can leave this as-is :-)

sonic2kk:
I did the refactor described above in 9f7c1c2. I made a couple of other changes, like renaming the steautil function to be more clear that it's for Steam. Since we just import it and use it without qualifying it as `steamutil.foo`, putting `steam` in the function name imo makes it clearer that this function is _specific_ to Steam, and will help avoid any potential name clashes :slightly_smiling_face: 

I think the new approach is cleaner. We just have one function now that we can call anywhere with a given ctool and `install_loc`, and it'll be able to set the global compat tool for us.

sonic2kk:
Fixed the focusing issue by re-arranging the focus order for the CtInfo dialog. This is probably worth visiting in other screens too but I don't think it's a super high priority for now.

<hr>

So pending some further review I think this PR is complete in my eyes, finally :sweat_smile: Took me way too long.

The warning dialog, Mark Global keyboard shortcut, and VDF logic refactor can go in separate PRs. Implementing this functionality for Lutris should also be possible and can go in a follow-up PR in future too!

DavidoTek:
Thanks! :tada:
I tested it and it works great.

pupgui2/pupgui2ctinfodialog.py
```diff
@@ -161,3 +166,7 @@ def search_ctinfo_games(self, text):
         for row in range(self.ui.listGames.rowCount()):
             should_hide: bool = not text.lower() in self.ui.listGames.item(row, 1).text().lower()
             self.ui.listGames.setRowHidden(row, should_hide)
+
+    def btn_mark_global_clicked(self):
```
DavidoTek, line 170: I wonder if we should disable the button after clicking because there currently is no visual feedback of the event.

That could be as simple as `self.btnMarkGlobal.setEnabled(false)`. We would need to add another check in the constructor so that the user won't click the button if it is already global.

sonic2kk: Oh that's a good idea! 

There's no reason to click this button again after marking a tool as global, at least not in the context of the currently open dialog.

An edge case could be opening two ctinfo dialogs at once, say GE-Proton9-7 and Luxtorpeda (just to have distinct examples). This is extremely unlikely to ever happen and could be alleviated with other behaviour, it's not a reason against implementing this button disabling but an outline of a potential problem scenario (although again, I *really* don't think anyone would do this if they weren't trying to break something)
1. User opens both dialogs at once.
2. User marks GE-Proton9-7 as Global.
3. With this suggestion, the button is disabled.
4. User then marks Luxtorpeda as Global without closing the prior dialog.
5. Perhaps a user did this in error and now they want GE-Proton9-7 to be the global tool again.
6. But since the global button is disabled on the GE-Proton9-7 dialog, they cannot.
7. Re-opening the dialog would fix this but it may not be very visually obvious.

A big way to alleviate this being a problem is the confirmation dialog noted earlier, because it would vastly reduce the likelihood of accidentally setting a tool as global.

There's also the possibility of closing the dialog which could be seen as expected since it's on the bottom row of the dialog buttons, but perhaps also not.

<hr>

Disabling the button makes sense to me, I'll add it when I get home. I'm not sure if we want to close dialogs after a tool is marked as global, it might be a little jarring.

I also had a thought, I don't remember what happens if a tool is marked as global and then the game list is refreshed. If we get the "tool is global" text like if the user closed and re-opened the dialog, maybe the button also gets hidden, which I think is good too.

sonic2kk: Oh, actually it seems like we don't hide the Mark Global button when the games list is refreshed.

We check if we should display the Mark Global button when loading the ctinfo dialog with `self.ui.btnMarkGlobal.setVisible(self.is_mark_global_available)`, perhaps we should also call this in `update_game_list_ui`?

I'm thinking we should rework the logic like this:
1. Remove the initial `btnMarkGlobal.setVisible` call from `setup_ui`.
2. `setup_ui` calls `update_game_list` (updates launcher-specific UI elements) which in turn calls `update_game_list_ui` (handles generic non-launcher-specific UI elements)
3. In `update_game_list_ui` we can call check if the Mark Global button should be displayed using `self.ui.btnMarkGlobal.setVisible(self.is_mark_global_available)` (since we may want to display this for other launchers, and `is_mark_global_available` will handle checking if Mark Global is available for the current launcher anyway)
    a. Right now the logic is only set if the current launcher is Steam, but with the recent refactor, we can call this generically and call this regardless of our current launcher, because `is_mark_global_available` is responsible for checking if the current launcher supports this action
5. We also move the `self.ui.btnMarkGlobal.clicked.connect` action to connect the button logic from `update_game_list` to `update_game_list_ui`
    a. Similar to above, this is also nested in a Steam-specific launcher check, but calling it from `update_game_list_ui` means we can connect this action regardless of launcher. Then the only thing we need to care about is if `is_mark_global_available` is True or False, and we don't need to care about knowing whether or not the launcher should support it, that's taken care of by the call to `is_mark_global_available`.
6. Update `set_launcher_global_tool` to return a success boolean
7. Change the call to `set_launcher_global_tool` to actually update `is_mark_global_available`, like this: `self.is_mark_global_available = set_launcher_global_tool(self.install_loc, self.ctool)`. We *could* I guess hardcode `self.is_mark_global_available = False` but that would assume `set_launcher_global_tool` will always succeed. If `set_launcher_global_tool` fails, this approach means we will update the value correctly.
    a. We could simply leave `set_launcher_global_tool` unchanged and instead make another call like this: `self.is_mark_global_available = is_mark_global_available(self.install_loc, self.ctool)` to achieve the same thing. The approach here is more stylistic preference, although I think personally making `set_launcher_global_tool` return a success boolean is a bit cleaner.
9. This would mean we wouldn't need to disable the button. Instead, we will update `is_mark_global_available` and our existing call to `btn_refresh_games_clicked` will handle running all the checks again for whether or not the Mark Global button should be displayed, using the same refactored logic described above that we would use when the dialog opens.

Basically we move the logic to set whether the button is visible or not into `update_game_list_ui`, meaning we will update it regardless of launcher, because `is_mark_global_available` performs a launcher check for us. Then when we mark a tool as global, we update the value of `self.is_mark_global_available`. Finally our existing call to refresh the game list elements means we will pick up any changes to `is_mark_global_available` and show/hide the button.

The reason showing and hiding may be better is because we show/hide the button when the dialog opens.

sonic2kk:
Based on a comment above, I pushed some changes in 05fb631 and 08849b7 which toggles the visibility of the Mark Global button based on whether `set_launcher_global_tool` successfully updates the launcher.
- `set_launcher_global_tool` now returns a `bool`; `False` if all the checks fail or if the function ends without hitting a launcher-specific block, OR the value of whether `set_steam_global_compat_tool` succeeded (as it can fall into an `except` block and fail to write to the VDF). This means `set_launcher_global_tool` will only return True if it was able to write out an updated global tool value to a given config file; right now that's only Steam's `config.vdf`.
- When we click on the Mark Global button, we update `self.is_mark_global_available` to be the `not set_launcher_global_tool`. This is because if `set_launcher_global_tool` *succeeds*, then the current tool was marked as global and so is *not* available to be marked as global.
- Now that we update `is_mark_global_available` when the button is clicked, we can move our logic to show/hide the Mark Global button into `update_game_list_ui`. which is called as a result of our button click calling `btn_refresh_games_clicked`. This means we'll run through the same logic used when the dialog opens to determine if the Mark Global button should be shown/hidden, and we use the updated value of `is_mark_global_available` to do it.

What all this means is if the Mark Global button is clicked and the tool is successfully marked as Global, we will hide the button.

sonic2kk:
Damn, I just realised that because we're using `set_launcher_global_tool` to update `is_mark_global_available`, if the global tool is updated externally from ProtonUp-Qt, the refresh button won't account for this because `is_mark_global_available` is only set when the dialog opens, and in `btn_mark_global_clicked`.

This could be fixed by setting `self.is_mark_global_available = is_mark_global_available(self.install_loc, self.ctool)` in  `update_game_list_ui` I guess...

pupgui2/pupgui2ctinfodialog.py
```diff
@@ -161,3 +167,7 @@ def search_ctinfo_games(self, text):
...
+        self.is_mark_global_available = not set_launcher_global_tool(self.install_loc, self.ctool)
+        self.btn_refresh_games_clicked()
```
sonic2kk, line 173: The reference to `self.ctool` is updated when the refresh button is clicked. This means `self.ctool.is_global` is still `False`. So when the tool is marked as global, or when the refresh button is generally clicked (i.e. if the global Steam compat tool was updated externally while the CtInfo dialog was open, and the refresh button was pressed after that) the dialog does not updated to reflect that the tool is global. 

So if GE-Proton9-7's CtInfo dialog is open, and the global compat tool in `config.vdf` is updated to be GE-Proton9-7, pressing the refresh button will not show/hide the Mark Global button, and it will not update the game list to display "Tool is Global".

However, it does update the compat tool list on the Main Menu to display `(global)` because the compat tool name!

And re-opening the dialog of course works as expected.

<hr>

I'm not sure how to fix this one, it is a bit of an edge-case and at worst something we can update in a later PR.

sonic2kk:
I think the last remaining issue is that when we refresh the CtInfo dialog, the `self.ctool` reference is not updated. This is a problem because after marking as global (which calls the Refresh method) we don't have an updated reference to `self.ctool.is_global`, so many UI show/hide functions still act as though we don't have a global compat tool.

We will need a way to update our reference to `self.ctool` to be more up-to-date. In `update_game_list_steam` we emit to `batch_update_completed` which correctly updates the display on the Main Menu to mark the tool as "(global)" (as it will re-parse the config file and so pick up that the tool is now global). However to my understanding Qt signals are asynchronous, so there is no way to know when the data is actually updated. This also only applies to Steam, meaning that for other launchers, 

Perhaps the best way to do this in a launcher-agnostic way is to return the ctool object in `set_launcher_global_tool` alongside the boolean success value.
- This would be straightforward but would not allow us to update the UI for other launchers to display that a tool is global. We can do this for Steam already because we unconditionally emit to batch_update_completed when we call `update_game_list_steam`.

We could emit to a slot that calls `MainWindow#update_ui` and then have `MainWindow#update_ui` emit to a slot that we listen for on CtInfo that gives us an updated compat_tool_index_map which we could then parse our ctool out of. This would allow us to:
- Update the UI for all launchers when we mark a tool as global, so they will all display `(global)`. If we are able to mark a tool as global from CtInfo, the launcher should always be able to display a tool as global, otherwise we wouldn't be able to mark it as global in the first place.
- Remove the unconditional call to the Batch Update Completed signal when we update the Steam Game List
- But this all has the drawback of `update_ui` returning us a compat_tool_index_map that we have to parse. We could get around this by making a util method to find a given `BasicCompatTool` object from a `compat_tool_index_map`

PR #337, opened 2023-12-17
CtInfo: Allow Label Text Selection

sonic2kk:
A little quality-of-life change to allow selecting the text in the CtInfo dialog with the mouse. This can be useful for quickly getting an install directory path if a user (or a nerd :nerd_face:) wants to get there a little bit quicker. For consistency, all three label values can be selected (the ones on the right, not their parent labels on the left), as even though it's much less common a user would actually select these imo, it would be a bit unexpected if they couldn't.

image

Selected text can also be dragged from ProtonUp-Qt into other applications, such as a terminal. This may be handy in cases like Steam Deck Game Mode, where if a user needs to visit their Steam installation files, they could more quickly get there.

In addition to, or instead of, we could add an icon button to the right of this label with a "copy to clipboard" type icon, to copy this path, which may be more touchscreen/gamescope friendly (as a small aside, this button would match the UI idea proposed in #336, but I'm getting ahead of myself). This may make the dialog wider than it needs to be though, and may add more clutter than anything else.

The keyboard selection flag was not enabled because it adds a cursor to the label, which looks strange. This however has the downside of not allowing selection changes with Shift+Arrow Keys. Though I think a triple-click to select all is probably the more common use-case for this :-)

### DIFF
--- a/pupgui2/resources/ui/pupgui2_ctinfodialog.ui
+++ b/pupgui2/resources/ui/pupgui2_ctinfodialog.ui
@@ -32,6 +32,9 @@
        <property name="text">
         <string/>
        </property>
+       <property name="textInteractionFlags">
+        <set>Qt::LinksAccessibleByMouse|Qt::TextSelectableByMouse</set>
+       </property>
       </widget>
      </item>
      <item row="1" column="0">
@@ -46,6 +49,9 @@
        <property name="text">
         <string/>
        </property>
+       <property name="textInteractionFlags">
+        <set>Qt::LinksAccessibleByMouse|Qt::TextSelectableByMouse</set>
+       </property>
       </widget>
      </item>
      <item row="2" column="0">
@@ -59,6 +65,9 @@
       <widget class="QLabel" name="txtInstallDirectory">
        <property name="text">
         <string/>
+       </property>
+       <property name="textInteractionFlags">
+        <set>Qt::LinksAccessibleByMouse|Qt::TextSelectableByMouse</set>
        </property>
       </widget>
      </item>


sonic2kk:
> Actually, I'm wondering how well that works with GameScope, did you test that by chance?

I didn't, but I checked just now, and you can't drag out of a GameScope session. However, at least in a nested session (where you run GameScope inside of another desktop session, i.e. from a terminal with `gamescope -- python3 -m pupgui2`, and not running it from a TTY), you can copy and paste by default at least with a build of GameScope from source. So from inside of your current desktop session (i.e. GNOME) if you run ProtonUp-Qt inside of GameScope from the terminal, you can select text and Ctrl+C/Ctrl+V to copy it to clipboard.

I'm not sure what happens in the Embedded Session though. If you, for example, run ProtonUp-Qt from Steam in Steam Deck Game Mode and try to copy and paste text then into Desktop Mode, I'm not sure what happens. Historically, the clipboard did not carry over, but I know there was a significant amount of work related to shared clipboard between the KDE Desktop and GameScope session on Steam Deck. In the past though, the clipboard did not sync this way. That may have changed.

Either way, this does rely on desktop keyboard shortcuts, which while absolutely available on SteamOS, would not be set up by default for ProtonUp-Qt and would require a keyboard by default (not ideal for a handheld). I *believe* in the Steam Input settings for a game you can bind Copy/Paste actions (though Copy is what we'd be interested in here), but this wouldn't be set by default (though a community controller config can be created and uploaded, even for Non-Steam Games!). A copy to clipboard button may be a good idea for this use-case :-)

PR #338, opened 2023-12-17
misc: Avoid shadowing 'map' name

sonic2kk:
`map` is a keyword, so this PR replaces a couple usages of `map` as a variable name to avoid that, and hopefully uses clearer variable names for this as well :-)


DavidoTek:
Thanks! :christmas_tree:

PR #354, opened 2024-02-11
FIX: awacy_icon denied problem

ahmed-abdelsalam:
# Overview

Fixing the problem when opening the game list, that contains a game with status denied.

## ScreenShot

image


sonic2kk (approved):

DavidoTek:
Thanks.

Not sure how we missed that :smile:

Looks like the cause for https://github.com/DavidoTek/ProtonUp-Qt/issues/235

Issue #363, opened 2024-03-09
games using dxvk or vkd3d in Lutris aren't listed


seeeeew:
**Describe the bug**  
ProtonUp-Qt erroneously reports no games are using dxvk or vkd3d in Lutris, despite them being used.

**To Reproduce**  
1. List installed compatibility tools for Lutris
2. "Show info" for any installed version of dxvk or vkd3d that is used by at least one game in Lutris

**Expected behavior**  
Either the games actually using the selected version of dxvk or vkd3d in Lutris should be listed, or there should be an indication that ProtonUp-Qt doesn't detect which games are using the selected version of dxvk or vkd3d in Lutris.

**Screenshots**  
![Screenshot from 2024-03-09 03-20-38](https://github.com/DavidoTek/ProtonUp-Qt/assets/639508/1cc13ead-a1a7-4526-b52b-9cfa50df3e92)

**Desktop:**  
 - **Platform**: Desktop PC
 - **System**: Ubuntu 22.04.4
 - **Version**: ProtonUp-Qt 2.9.1
 - **How did you install ProtonUp-Qt?**: Flatpak

**Terminal output**  
```
Gtk-Message: 03:23:24.786: Failed to load module "canberra-gtk-module"
Gtk-Message: 03:23:24.787: Failed to load module "canberra-gtk-module"
qt.qpa.qgnomeplatform: Could not find color scheme  ""
Qt: Session management error: Could not open network socket
qt.pysideplugin: Environment variable PYSIDE_DESIGNER_PLUGINS is not set, bailing out.
qt.pysideplugin: No instance of QPyDesignerCustomWidgetCollection was found.
qt.pysideplugin: No instance of QPyDesignerCustomWidgetCollection was found.
ProtonUp-Qt 2.9.1 by DavidoTek. Build Info: DavidoTek Flathub build.
Python 3.10.13 (main, Nov 10 2011, 15:00:00) [GCC 12.2.0], PySide 6.5.2
Platform: KDE Flatpak runtime 6.5 Linux-6.5.0-21-generic-x86_64-with-glibc2.35
Loading locale de / de_DE
Loaded ctmod GE-Proton
Loaded ctmod Wine-GE
Loaded ctmod Boxtron
Loaded ctmod D8VK (nightly)
Loaded ctmod Kron4ek Wine-Builds Vanilla
Loaded ctmod Lutris-Wine
Loaded ctmod Luxtorpeda
[...]

Issue #347, opened 2024-01-26
Introduce Tests into ProtonUp-Qt


sonic2kk:
**Is your feature request related to a problem? Please describe.**
Not really :stuck_out_tongue: 

**Describe the solution you'd like**
ProtonUp-Qt has become a very mature project in terms of functionality. The codebase is clean and readable imo and I frequently use it as reference when working on my own projects.

As the project is no longer in early stages of development, it might worth spending some time in investigating some unit tests. There has been discussion around re-organizing parts of the code (such as moving some of the util functions around, and restructing ctmods a little). Having a test framework in place could help reduce the risk of regressions during refactors.

Unit tests can also generally be a good way to "check your work", as sometimes when you're writing test cases, it helps you realise something might not be architected in the best way, or you might've taken the wrong approach. (EDIT: I realised when reading this back how this may sound, this is not directed at you or other contributors; it's actually directed a little at myself!)

Using unit tests *could* also help us catch AppImage compatibility breakages I think, if they're integrated as a workflow that runs on a specific Python instance (if that's not going to be expensive and cost you money :sweat_smile:).

Of course, unit tests are only as good as how they're written and used. Whether or not unit tests are enforced/encouraged as part of a PR, reviewing tests, it is more of a maintenance burden, but it _can_ help. Also, it's important to be vigilant and not write pointless tests. Not everything needs a test, and 100% code coverage need not be the goal, especially for a GUI program.

Which leads me to: ProtonUp-Qt is heavily Qt-based, and writing UI tests may not be worth anyone's time. What I'm more interested in is having tests for utility functions primarily.

**Describe alternatives you've considered**
Unit tests are, plainly speaking, a pain in the ass sometimes! Writing a new feature and then thinking "jeez, how am I gonna write a test case for that..." especially for complex cases that require mocking. Simply not bothering is an alternative. There's no burning need, this is really just a bit of spit-and-shine for project polish.

This also won't really have any direct user-facing changes. As such, I can entirely see and understand the mindset of, well: *who cares?* - This is entirely a wishlist-y proposal, but one that I think long-term could help keep with maintenance and housekeeping.

Does ProtonUp-Qt *need* tests? In my opinion, it doesn't. But, I could see some cases where they could be helpful. I have caused regressions before, and I should've absolutely been more careful, but having some tests might've helped me catch them.

**Additional context**
[...]

PR #364, opened 2024-03-09
Lutris: Add some util functions for conditional game list parsing

sonic2kk:
Some readability improvements that may help extending these checks later for future work, hopefully not too premature! :smile: 

When I was digging into #363 I noticed that some of the conditional logic was a bit long and seemingly arbitrary. To help with readability I made some generic utility functions that should explain better than the checks we were doing before.

For #363 we may need to extend the conditional logic i.e. in `pupgui2ctinfodialog.py`, we may need to add a check for whether a given runtime is in use. This would get pretty long, but with this format, we could make `is_lutris_game_using_runtime('dxvk', game.dxvk_version)` (or something along these lines, idk how we would want to do this just yet, and not too important for this PR :-) ).

The first function added is to check what runner a `LutrisGame` is using, with `not None` and length sanity checks. We were already doing this for the Games List

(https://github.com/DavidoTek/ProtonUp-Qt/blob/6cf1def3bf1ee9775b2639ed41aad1ff96ae2afa/pupgui2/pupgui2gamelistdialog.py#L178)

But this function now means anytime we check the runner name, we perform these checks.

The 2nd function is more specific, and it checks if a `LutrisGame` is using Wine as its runner, with an optional parameter to check if it's using a specific Wine version. I figured there's no harm in making this function a bit more flexible as it was pretty straightforward to extend.

This could make our checks a lot cleaner should they be extended further. At least in my opinion, `if is_lutris_game_using_wine(game, self.ctool.displayname) or is_lutris_game_using_runtime(game, ...)` is a lot cleaner than extending our existing check.

### DIFF
--- a/pupgui2/lutrisutil.py
+++ b/pupgui2/lutrisutil.py
@@ -47,3 +47,28 @@ def get_lutris_game_list(install_loc) -> List[LutrisGame]:
     except Exception as e:
         print('Error: Could not get lutris game list:', e)
     return lgs
+
+
+def is_lutris_game_using_runner(game: LutrisGame, runner: str) -> bool:
+
+    """ Determine if a LutrisGame is using a given runner. """
+
+    is_runner_name_valid = game.runner is not None and len(game.runner) > 0
+    is_using_runner = game.runner == runner
+
+    return is_runner_name_valid and is_using_runner
+
+
+def is_lutris_game_using_wine(game: LutrisGame, wine_version: str = '') -> bool:
+
+    """ Determine if a LutrisGame is using a given wine_version string. """
+
+    is_using_wine = is_lutris_game_using_runner(game, 'wine')
+
+    # Only check wine_version if it is passed
+    if len(wine_version) > 0:
+        is_using_wine_version = game.get_game_config().get('wine', {}).get('version', '') == wine_version
+    else:
+        is_using_wine_version = True
+
+    return is_using_wine and is_using_wine_version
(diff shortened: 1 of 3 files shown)


sonic2kk:
There are some typing issues in this file, but I think we can clean this up as part of a general sweep of typing changes, once we have Python 3.10 (#359).

sonic2kk:
I'm wondering if these would be better as methods on the `LutrisGame` class itself, since we have to pass it a `LutrisGame` anyway...

DavidoTek:
Thanks! Looks good.

Makes sense to create utility functions for that, especially if we also need `is_lutris_game_using_runtime('dxvk', game.dxvk_version)` later.

I noticed that the sanity checks also catch situations where the game.runner is None. Can't remember if that is the case anywhere. If yes, we might want to add a type hint like `runner: Union[str, None]` to `datastructures.py#LutrisGame` in the future.

> The 2nd function is more specific, and it checks if a LutrisGame is using Wine as its runner, with an optional parameter to check if it's using a specific Wine version. I figured there's no harm in making this function a bit more flexible as it was pretty straightforward to extend.
> This could make our checks a lot cleaner should they be extended further. At least in my opinion, if is_lutris_game_using_wine(game, self.ctool.displayname) or is_lutris_game_using_runtime(game, ...) is a lot cleaner than extending our existing check.

Yes, that's definitively worth adding the utility functions for making the other code cleaner.

(I just noticed that these functions are a great example for something that could use unit tests https://github.com/DavidoTek/ProtonUp-Qt/issues/347. Haven't gotten around to look into that yet...)

> I'm wondering if these would be better as methods on the LutrisGame class itself, since we have to pass it a LutrisGame anyway...

I feel like that it is fine as you did for now. I wonder if at some point we might actually want to move the data structures into the launcher specific files (i.e., move `LutrisGame` to `lutrisutil.py`). We could then move the functions into `LutrisGame` and just rename the `game` parameter into `self`.

PR #371, opened 2024-04-08
Custom Install Directory: Use Inline `if` for Safer Default Launcher Selection

### LINKED ISSUE #370: Steam not visible and "..." crashes app on first run

**Describe the bug**  
./ProtonUp-Qt-2.9.1-x86_64.AppImage 
ProtonUp-Qt 2.9.1 by DavidoTek. Build Info: Official AppImage by DavidoTek.
Python 3.8.2 (default, Mar 13 2020, 10:14:16) [GCC 9.3.0], PySide 6.2.4
Platform: Ubuntu 22.04 Linux-6.5.0-14-generic-x86_64-with-glibc2.29
Gtk-Message: 14:41:27.549: Failed to load module "canberra-gtk-module"
Gtk-Message: 14:41:27.556: Failed to load module "canberra-gtk-module"
Loading locale en / en_US
Loaded ctmod GE-Proton
Loaded ctmod Wine-GE
Loaded ctmod Boxtron
Loaded ctmod D8VK (nightly)
Loaded ctmod Lutris-Wine
Loaded ctmod Luxtorpeda
Loaded ctmod Proton Tkg
Loaded ctmod Roberta
Loaded ctmod Steam-Play-None
Loaded ctmod vkd3d-lutris
Loaded ctmod vkd3d-proton
Loaded ctmod Wine Tkg (Valve Wine)
Loaded ctmod Wine Tkg (Vanilla Wine)
Loaded ctmod DXVK
Loaded ctmod DXVK Async
Loaded ctmod DXVK (nightly)
qt.pysideplugin: Environment variable PYSIDE_DESIGNER_PLUGINS is not set, bailing out.
qt.pysideplugin: No instance of QPyDesignerCustomWidgetCollection was found.
Gamepad error: No gamepad found.
qt.pysideplugin: No instance of QPyDesignerCustomWidgetCollection was found.
  File "/tmp/.mount_ProtonIvo955/usr/lib/python3.8/site-packages/pupgui2/pupgui2.py", line 381, in btn_manage_install_locations_clicked
    customid_dialog = PupguiCustomInstallDirectoryDialog(install_directory(), parent=self.ui)
  File "/tmp/.mount_ProtonIvo955/usr/lib/python3.8/site-packages/pupgui2/pupgui2customiddialog.py", line 32, in __init__
    self.setup_ui()
  File "/tmp/.mount_ProtonIvo955/usr/lib/python3.8/site-packages/pupgui2/pupgui2customiddialog.py", line 53, in setup_ui
    self.set_selected_launcher(self.install_locations_dict[self.launcher] or 'steam')  # Default combobox selection to "Steam" if unknown launcher for some reason
KeyError: ''

 - **Platform**:  Desktop Ubuntu 22.04
 - **Version**: For example ProtonUp-Qt 2.9.1
 - **How did you install ProtonUp-Qt?**: same error for Flatpak / AppImage

sonic2kk:
Fix #370.

Almost forewent my usual PR format, but that would be no fun, right? :wink: 

## Overview
Changes the check in the Custom Install Dialog that selects the default launcher to use an inline `if`, which fixes a crash if `self.launcher` is not a key in `self.install_locations_dict`. This currently occurs when there are no available launchers installed. We get a `KeyError` if no launchers are found, because `self.launcher` is not a valid key in `self.install_locations_dict`.

## Problem
Apparently, in Python, `dictionary[key] or 'value'` will result in a `KeyError` if `key` is not defined. However in other cases, this does not happen. For example with an inline `if` I guess the conditional gets parsed first, like a "full" `if`, so you can do `dictionary[key] if key in dictionary else 'value'` safely, as Python won't process the `dictionary[key]` first. Makes sense having thought it through (you wouldn't want `print('hello') if condition else print ('goodbye')` to print `hello`, so `condition` has to be processed fiirst), but tripped me up a little.

## How to Replicate Error
You can test this fix by forcefully returning False at the beginning if `util#is_valid_launcher_installation`. This will result in ProtonUp-Qt not finding any launchers. On `main`, opening the Custom Install Directory dialog will give a crash. With this change, it should work correctly.

I did some testing and this prevents the crash, and the crash I was encountering matches the one described in #370 (`KeyError: ''` and noting line 53, where this fix is implemented). I am very confident this is the cause of the crash. What I am not sure of is how no Steam installation was found in #370. For some reason this seems to be coming up more in recent months for no clear reason...

## Future Work
There are some UIX issues I'd like to tackle in separate PRs that won't depend on the changes here. For example, adding a custom install directory will not immediately find all compatibility tools, you have to restart the launcher. If you force no launchers to be found and then add a valid path to a Steam installation's `compatibilitytools.d`, you have to restart ProtonUp-Qt for any tools to display.

<hr>

Small change, big PR description :sweat_smile: 

Thanks!

### DIFF
--- a/pupgui2/pupgui2customiddialog.py
+++ b/pupgui2/pupgui2customiddialog.py
@@ -50,7 +50,7 @@ class PupguiCustomInstallDirectoryDialog(QObject):
             display_name for display_name in self.install_locations_dict.values()
         ])
 
-        self.set_selected_launcher(self.install_locations_dict[self.launcher] or 'steam')  # Default combobox selection to "Steam" if unknown launcher for some reason
+        self.set_selected_launcher(self.install_locations_dict[self.launcher] if self.launcher in self.install_locations_dict else 'steam')  # Default combobox selection to "Steam" if unknown launcher for some reason
 
         self.ui.btnSave.clicked.connect(self.btn_save_clicked)
         self.ui.btnDefault.clicked.connect(self.btn_default_clicked)


DavidoTek:
Thanks!

Oh, that fix makes sense. I guess the idea of the `custom install dialog` is to have the ability to add a launcher manually if it is not detected automatically.

> Apparently, in Python, dictionary[key] or 'value' will result in a KeyError if key is not defined. However in other cases, this does not happen

Hmm, a `self.install_locations_dict.get(self.launcher) or 'steam'` probably would do the same. I guess your solution achieves the same effect.

> I did some testing and this prevents the crash, and the crash I was encountering matches the one described in #370 (KeyError: '' and noting line 53, where this fix is implemented). I am very confident this is the cause of the crash.

I didn't explicitly test that scenario, but the current code is definitively defective. Seems reasonable that this is the defect which causes the error.

> What I am not sure of is how no Steam installation was found in #370. For some reason this seems to be coming up more in recent months for no clear reason...

Yes, that's a bit strange. I haven't encountered anything similar yet, but I don't exactly have the most exotic setup either.

> For example, adding a custom install directory will not immediately find all compatibility tools, you have to restart the launcher
> If you force no launchers to be found and then add a valid path to a Steam installation's compatibilitytools.d, you have to restart ProtonUp-Qt for any tools to display.

Okay. I agree, this could use some reworking.

> Almost forewent my usual PR format, but that would be no fun, right? 😉
> Small change, big PR description 😅

I like it. Prevents any confusion about the content of the PR 😉

This one will be a quick merge.

sonic2kk:
Ah damn you're right, `self.install_locations_dict.get(self.launcher, 'steam')` would probably also work (so it would fall back to `'steam'`). Totally didn't occur to me 😅

PR #379, opened 2024-04-21
Games List: Add tooltip to Deck Compatibility column

sonic2kk:
Depending on the length of the compatibility tool name, or more importantly, the font selected on the user's system and the sizing of this font, the text for the Deck Compatibility tool name may be cut off.

This PR adds a tooltip to the cells on the Deck Compatibility column that displays the text of the cell. Below is a screenshot to illustrate the problem and the tooltip implemented here. For reference, the font used here is Inter Display Medium 10pt.

image

The cell could be resized to see it, but we already display a tooltip for the Game Name, and resizing columns can create a misaligned UI, so I think this is a slight usability improvement. I discovered this because I tried to mouse over the cell to view the name in the tooltip to avoid resizing the columns. This could also go hand-in-hand with resizing the columns, as if a user made this column smaller to view others, the text would be cut off. This makes it easier to see the text on this column without resizing.

### DIFF
--- a/pupgui2/pupgui2gamelistdialog.py
+++ b/pupgui2/pupgui2gamelistdialog.py
@@ -147,8 +147,13 @@ class PupguiGameListDialog(QObject):
             self.ui.tableGames.setItem(i, 4, QTableWidgetItem())
             self.ui.tableGames.setCellWidget(i, 4, btn_fetch_protondb)
 
+            # Steam Deck compatibility 
             lbltxt = self.get_steamdeck_compatibility(game)
-            self.ui.tableGames.setItem(i, 2, QTableWidgetItem(lbltxt))
+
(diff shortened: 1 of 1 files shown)


sonic2kk:
Just noticed, for very minor consistency, we might want to use "Deck **C**ompatibility", with an uppercase "C" (matches "Compatibility Tool")

DavidoTek:
Good catch, that is a nice addition.
Just something that occurred to me: Is there a way to view the tooltips on Steam Deck?

> Just noticed, for very minor consistency, we might want to use "Deck Compatibility", with an uppercase "C" (matches "Compatibility Tool")

I added a commit that changes it. I hope I didn't break the i18n again with that... :smile:

PR #381, opened 2024-04-23
util: Use os.path.abspath for is_valid_launcher_installation

### LINKED ISSUE #378: 2.9.2 x86_64 crashes on launch
**Describe the bug**  
Unable to launch ProtonUp-Qt-2.9.2-x86_64.AppImage

**To Reproduce**  
Launch it with any method

**Expected behavior**  
Program opens

**Screenshots**  

 - **Platform**: PC
 - **System**: Mint 21.3, kernel 6.5.0-26-generic, last mesa from the PPA.
 - **Version**: ProtonUp-Qt 2.9.2
 - **How did you install ProtonUp-Qt?**: AppImage

**Additional context**  
2.8.0 work fine, tried to update and latest version doesn't work 2.9.2. I can test other version if you think it would help?

**Terminal output**  
```
$ ./ProtonUp-Qt-2.9.2-x86_64.AppImage 
ProtonUp-Qt 2.9.2 by DavidoTek. Build Info: Official AppImage by DavidoTek.
Python 3.10.4 (main, Apr  2 2022, 09:04:19) [GCC 11.2.0], PySide 6.7.0
Platform: LinuxMint 21.3 Linux-6.5.0-26-generic-x86_64-with-glibc2.35
qt.qpa.plugin: From 6.5.0, xcb-cursor0 or libxcb-cursor0 is needed to load the Qt xcb platform plugin.
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vkkhrdisplay, vnc, wayland-egl, wayland, xcb.

[1]    2597968 IOT instruction (core dumped)  ./ProtonUp-Qt-2.9.2-x86_64.AppImage
```

sonic2kk:
Fix #378.

If the compatibilitytools.d dir is a symlink, `realpath` will resolve the symlink before expanding and moving up a directory, meaning we will end up looking for the Steam data files in the symlink's parent directory instead of the Steam root, as we assumed the parent folder of compatibilitytools.d would always be a Steam root directory.

To fix this, we use `abspath` which will resolve the path without expanding symlinks.

We probably use `realpath` in a lot of places where symlinks might be used, so we might want to consider revisiting these. Anywhere we might expand an install directory path before moving up a directory is likely unsafe. I thought we did this for some cases with Lutris/Heroic (to install in the runtimes folder instead of the runners folder, for example) but I didn't spot any with a quick glance. It's also pretty late so I might've just missed it :smile: 

Reference:
- Python docs: https://docs.python.org/3/library/os.path.html#os.path.abspath
- StackOverflow answer with better explanation: https://stackoverflow.com/questions/37863476/why-would-one-use-both-os-path-abspath-and-os-path-realpath/40311142#40311142

### DIFF
--- a/pupgui2/util.py
+++ b/pupgui2/util.py
@@ -210,7 +210,7 @@ def is_valid_launcher_installation(loc) -> bool:
     #
     # In future we could expand this to other Steam flavours and other launchers.
     if loc['display_name'] == 'Steam':  # This seems to get called many times, why?
-        return is_valid_steam_install(os.path.realpath(os.path.join(install_dir, '..')))
+        return is_valid_steam_install(os.path.abspath(os.path.join(install_dir, '..')))
     
     return os.path.exists(install_dir)  # Default to path check for all other launchers
 


sonic2kk:
Not sure if it's worth adding a comment on why we're using `abspath` instead of `realpath` here...

DavidoTek:
Thanks!

That's interesting, I wouldn't have expected this to cause an issue. I thought if there was a symlink, it would include the entire Steam folder and not just the `compatibilitytools.d`.
Good that it is fixed now.

> We probably use realpath in a lot of places where symlinks might be used, so we might want to consider revisiting these

There are cases where this behavior is actually desired:
If there is a Steam installation at `~/.steam/root` and a symlink `~/.local/share/Steam` that points to the actual folder, `os.path.realpath` in `constants.py` will resolve both paths to the actual path at `~/.steam/root` and get rid of the duplicates using `list(dict.fromkeys(...))`\*
As far as I can remember, this is also the reason why we put `~/.local/share/Steam` first in the list because it is more likely that a symlink at `~/.steam/root` will point to `~/.local/share/Steam` and not the other way around like I described above (I think we already had the second situation in some issue).

https://github.com/DavidoTek/ProtonUp-Qt/blob/23b60df1ee379103d6652af08630d89a2a6ae1b6/pupgui2/constants.py#L28-L33

> Not sure if it's worth adding a comment on why we're using abspath instead of realpath here...

Might actually be worth adding it, there are already a few other comments, so it should be fine.
I think it is easy to google the difference, but it could be helpful to know why exactly we don't want to resolve symlinks here (I guess we should also link/ID this PR in the comment then).

---

\*there is a funny behavior: If you create a symlink `~/.steam/root` -> `~/.var/app/com.valvesoftware.Steam/data/Steam`, it will display the correct path (`~/.var/app/...`), but it detects it as a native/non-Flatpak installation. I don't think that is anything we need to worry about though :smile:

Issue #387, opened 2024-04-26
Unable to download GE-Proton versions 6 or older


robohappy:
**Describe the bug**  
On the version scroll down menu for GE-Proton, it only goes down far enough for  7.0rc3-GE-1, no GE Protons from 6 or older show on the menu.

**To Reproduce**  
1. Click add version
2. Have compatibilty tool set to GE-Proton
3. On the version scroll down menu, scroll to the bottom
4. Only goes down far enough for  7.0rc3-GE-1

**Expected behavior**  
Should be able to access all older versions of GE-Proton

**Screenshots**  
![Screenshot_20240425_213924](https://github.com/DavidoTek/ProtonUp-Qt/assets/112583688/9a062bb7-3e28-4bf9-8629-c36006374b59)

**Desktop (please complete the following information):**  
 - **Platform**: PC
 - **System**: Manjaro
 - **Version**: 2.9.2-1
 - **How did you install ProtonUp-Qt?**: Manjaro Repository
**Additional context**  

**Terminal output**  
```
protonup-qt                                                      ✔ 
ProtonUp-Qt 2.9.2 by DavidoTek. Build Info: built from source.
Python 3.11.8 (main, Feb 12 2024, 14:50:05) [GCC 13.2.1 20230801], PySide 6.7.0
Platform: ManjaroLinux 23.1.4 Linux-6.8.7-1-MANJARO-x86_64-with-glibc2.39
Loading locale en / en_US
Loaded ctmod GE-Proton
Loaded ctmod Wine-GE
Loaded ctmod Boxtron
Loaded ctmod D8VK (nightly)
Loaded ctmod Kron4ek Wine-Builds Vanilla
Loaded ctmod Lutris-Wine
Loaded ctmod Luxtorpeda
Loaded ctmod Northstar Proton (Titanfall 2)
Loaded ctmod Proton Tkg
Loaded ctmod Proton Tkg (Wine Master)
Loaded ctmod Roberta
Loaded ctmod Steam-Play-None
Loaded ctmod SteamTinkerLaunch
Loaded ctmod SteamTinkerLaunch-git
Loaded ctmod vkd3d-lutris
Loaded ctmod vkd3d-proton
Loaded ctmod Wine Tkg (Valve Wine Bleeding Edge)
Loaded ctmod Wine Tkg (Wine Master)
Loaded ctmod DXVK
Loaded ctmod DXVK Async
Loaded ctmod DXVK (nightly)
[...]

PR #398, opened 2024-05-16
Add a "Load more" button to the Install dialog

DavidoTek:
Fix https://github.com/DavidoTek/ProtonUp-Qt/issues/387

Adds a "Load more..." option to the install dialog when there are possibly further versions available (`len(vers) < RELEASES_PER_PAGE`). A `page` parameter is added to the `fetch_releases` method of all ctmods for this purpose.

grafik


pupgui2/pupgui2installdialog.py
```diff
@@ -56,24 +63,63 @@ def btn_install_clicked(self):
...
+            if len(vers) < RELEASES_PER_PAGE:
+                self.more_releases_loadable = False
+
+            # Stops install dialog UI elements from being enabled when rate-limited to prevent switching/installing tools
+            if len(vers) > 0:
+                self.ui.comboCompatToolVersion.addItems(vers)
+                self.ui.comboCompatToolVersion.setCurrentIndex(0)
+
+                if self.more_releases_loadable:
+                    self.ui.comboCompatToolVersion.addItem(self.tr('Load more...'))
```
sonic2kk, line 95: It might be nice if we could add a separator here, for a bit of visual clarity.

```
separator_index = self.ui.comboCompatToolVersion.count()
self.ui.comboCompatToolVersion.insertSeparator(separator_index)
```

With the KDE Breeze system theme, this separator is invisible, but it shows up fine for the dark theme. It does add some strange padding to the bottom of the dropdown though, but that happens when the separator is inserted anywhere.

Although adding this would be a bit ugly in the code with the logic earlier to `removeItem`. We'd have to call it twice.

Unfortunately I can't take a screenshot of the dropdown, it doesn't display...

DavidoTek: Good idea. I will take a look at that.
The alternative is to highlight the text somehow, maybe by adding an unicode/png icon or by changing the font.

sonic2kk: This is way better, I wasn't sure if you could add icons and I didn't see any in the project that would be fitting anyway. You could've fooled me, I think the icon looks pretty spiffy :smile:

PR #400, opened 2024-05-18
Add Unit Testing Workflow using PyTest

DavidoTek:
Fix https://github.com/DavidoTek/ProtonUp-Qt/issues/347

- Adds a CI workflow for Unit Testing using `pytest`
  - Runs on *Pull Requests* and *Pushes to main*
  - Currently tests on Ubuntu 22.04 and Python 3.10
- Adds an initial test for `util.py#get_random_game_name`
- Produces Markdown Test Report using [pytest-md](https://github.com/hackebrot/pytest-md)

---

<details>
  <summary>Fixed Qt import for PyTest by installing libEGL and libxcbcommon and setting QT_QPA_PLATFORM=offscreen</summary>

**Previously**: Currently fails due to missing `libEGL`, i.e. the runner is headless but the Qt imports expect a graphics library. I will investigate that.

```sql
==================================== ERRORS ====================================
_____________________ ERROR collecting tests/test_util.py ______________________
ImportError while importing test module '/home/runner/work/ProtonUp-Qt/ProtonUp-Qt/tests/test_util.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_util.py:1: in <module>
    from pupgui2.util import *
pupgui2/util.py:20: in <module>
    from PySide6.QtWidgets import QApplication, QStyleFactory, QMessageBox, QCheckBox
E   ImportError: libEGL.so.1: cannot open shared object file: No such file or directory
---------------- generated Markdown report: /tmp/tmp.t8xQFKNH8p ----------------
```
</details>

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -1,0 +1,22 @@
+from pupgui2.util import *
+
+from pupgui2.datastructures import SteamApp, LutrisGame, HeroicGame
+
+
+def test_get_random_game_name():
+    """ test whether get_random_game_name returns a valid game name """
+    names = ["game", "A super cool game", "A game with a very long name that is very long", "0123456789"]
+
+    steam_app = [SteamApp() for _ in range(len(names))]
+    lutris_game = [LutrisGame() for _ in range(len(names))]
+    heroic_game = [HeroicGame() for _ in range(len(names))]
+
+    for i, name in enumerate(names):
+        steam_app[i].game_name = name
+        lutris_game[i].name = name
+        heroic_game[i].title = name
+
+    for i in range(10):
+        assert get_random_game_name(steam_app) in names
+        assert get_random_game_name(lutris_game) in names
+        assert get_random_game_name(heroic_game) in names
--- a/.github/workflows/testing-ci.yml
+++ b/.github/workflows/testing-ci.yml
@@ -24,13 +24,17 @@ jobs:
         cache: "pip"
 
     - name: Install dependencies
-      run: pip install pytest pytest-md pytest-emoji
+      run: |
+        sudo apt install -y libegl1 libxkbcommon0
+        pip install pytest pytest-md pytest-emoji
 
     - name: Install ProtonUp-Qt
       run: pip install -e .
 
     - name: Run pytest
       uses: pavelzw/pytest-action@v2
+      env:
+        QT_QPA_PLATFORM: "offscreen"
       with:
         verbose: true
         emoji: true
--- a/.github/workflows/testing-ci.yml
+++ b/.github/workflows/testing-ci.yml
@@ -18,9 +18,10 @@ jobs:
     - uses: actions/checkout@v4
 
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v5
       with:
         python-version: "${{ matrix.python-version }}"
+        cache: "pip"
 
     - name: Install dependencies
       run: pip install pytest pytest-md pytest-emoji
--- a/.github/workflows/testing-ci.yml
+++ b/.github/workflows/testing-ci.yml
@@ -1,0 +1,39 @@
+name: Test using Pytest
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  tests:
+    name: "Run PyTest"
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        python-version: ["3.10"]
+
+    steps:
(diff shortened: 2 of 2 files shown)


tests/test_util.py
```diff
@@ -0,0 +1,22 @@
...
+    lutris_game = [LutrisGame() for _ in range(len(names))]
+    heroic_game = [HeroicGame() for _ in range(len(names))]
+
+    for i, name in enumerate(names):
+        steam_app[i].game_name = name
+        lutris_game[i].name = name
+        heroic_game[i].title = name
+
+    for i in range(10):
+        assert get_random_game_name(steam_app) in names
```
sonic2kk, line 20: So this is just "regular" Python code, but PyTest will read and run these functions, and collate the rest results, tell you where certain tests fell over and why (e.g. expected X list of values, got Y). Is that right?

That means for "basic" tests there is not always a requirement to import or use anything from PyTest? There is probably some specific PyTest stuff for mocking and such, but it was interesting to me to see this as just simple Python code! That would be a big question mark I had around unit testing out of the way...

I'm definitely hoping to learn a lot more about PyTest and to go back and write unit tests for the code I've contributed, so figured I'd ask here :smile:

DavidoTek: > So this is just "regular" Python code, but PyTest will read and run these functions, and collate the rest results, tell you where certain tests fell over and why (e.g. expected X list of values, got Y). Is that right?
> That means for "basic" tests there is not always a requirement to import or use anything from PyTest?

That is correct.
PyTest imports are only required for e.g. [fixtures](https://docs.pytest.org/en/latest/explanation/fixtures.html) which provide reliable/constant example data.
It will look for files prefixed with `test_` by default.

> I'm definitely hoping to learn a lot more about PyTest and to go back and write unit tests for the code I've contributed, so figured I'd ask here 😄

I have the same goal. I've worked with tools like [Vitest](https://vitest.dev/) before but I'm also new to PyTest :smile:

There are also a few things I still need to figure out. For example, I'm calling the `assert` statement 30 times. I've seen that (with Java I think) and it definitively works as it will just throw an `AssertionError` in case the parameter is `False`. But I don't know yet whether that is the correct/best way to do so.

sonic2kk:
Woohoo, once v2.10.0 is out and I get some of the ideas I have off my plate, I'll definitely see about submitting unit tests!

sonic2kk:
Do we need to add PyTest as a dependency in `requirements.txt`? On one hand it's not needed for actually bundling the AppImage and Flatpak, but it is needed for local development. I'm not sure how other projects handle this, maybe this is just something for a future `CONTRIBUTING.md` file to note that PyTest should be installed [following the steps in their Documentation](https://docs.pytest.org/en/stable/getting-started.html).

DavidoTek:
> Do we need to add PyTest as a dependency in requirements.txt?

Good question. Following what I read online, we may create a `requirements_dev.txt` or similar file. I'm unsure whether we want to do that for a single dependency.
The `CONTRIBUTING.md` method you mentioned is also okay I guess.

sonic2kk:
It looks like PyTest has a lot of plugins, some of which we might find useful.
- PyTest-Dev GitHub Organization: https://github.com/pytest-dev
- PyTest-Mock: https://github.com/pytest-dev/pytest-mock
- PyTest-Qt: https://github.com/pytest-dev/pytest-qt
- PyFakeFS: https://github.com/pytest-dev/pyfakefs
- PyTest-Randomly: https://github.com/pytest-dev/pytest-randomly

Outside of that, we may also want to use [requests-mock](https://requests-mock.readthedocs.io/en/latest/overview.html#installation).

Adding all of these all at once may not make much sense, but in order to make sure these are available on a user's machine (so they don't have to hunt through the code to see what we might be importing, especially as the test suite get more robust) and in order to make sure our test dependencies are available on CI, it could be an idea to make a `requirements_dev.txt`.

DavidoTek:
Good to know.

> Outside of that, we may also want to use [requests-mock](https://requests-mock.readthedocs.io/en/latest/overview.html#installation), or alternatively the [responses](https://github.com/getsentry/responses#basics) library - I would probably lean towards using responses, from looking over the usage it seems very clean and very natural and complementory to the requests usage!

Ah, wasn't aware of responses. Seems to be the one which is used more often (4.1k stars and 24M downloads per month).

> it could be an idea to make a requirements_dev.txt

Yes, we should create that file and add dependencies one by one when we need them.

PR #401, opened 2024-05-18
Games List: Add checks for skipping invalid games

### LINKED ISSUE #397: Lutris Games List Crashes if Game Has No Runner, Install Location, or Installation Time
**Describe the bug**  
**What the problem is**: When viewing the Games List for Lutris, if there is a game with no runner, it will crash because we attempt to run `capitalize()` on the runner name to display it. Even if we try to skip games that have `game.runner == None` or infer a game's runner based on the Lutris config file (if a Lutris config has `game.appid` then we can assume it's a Steam game), we run into problems where we're missing other information like `game.install_loc` and `game.installed_at`.

**What causes the problem**: Valid games should always have a runner, but Lutris will store uninstalled Steam games in its sqlite DB. So if you link your Steam library to Lutris, uninstalled Steam games will be there but they will be missing a `runner`, `install_loc`, `installed_at`, and I believe a `slug`.

**When did this start**: I am not sure if Lutris always stored Steam games like this. Attempting to sync my games list always caused Lutris to crash, but at the same time I don't remember it missing any games. So I think this might be new behaviour that it stores these games in its DB.

**Possible solutions**:
1. One solution might be to filter the games to show on the games list even further. If a game doesn't have a runner, and install location, OR an install time, then we can skip them. We can make a method in `PupguiGameListDialog` to do this, and we can call it when we're generating the Lutris game list:
    https://github.com/DavidoTek/ProtonUp-Qt/blob/db33194616ac32b986ec3e12929803c48f6de294/pupgui2/pupgui2gamelistdialog.py#L183
    In this case I have found the issue of all these missing fields only being applicable for not-installed Steam games that are in my library, but it could theoretically happen for any other game using any runner. That said, in my head it's equally valid to assume any game from any runner is invalid if it's missing all of this data (invalid meaning, _not relevant to display on the Game List_).
2. Another solution is to be more "None-safe" and display labels missing this data as "Unknown", but I am slightly less in favour of this. If a game is missing so much information, like runner and install location, it might be better to assume that it is either invalid or not installed. In other words, in Solution 1 we filter out games missing all this data, which means we effectively consider games missing all this data as not valid to display on the Games List. Sure, we could just set the labels as "Unknown", but I don't know. I think at least right now we're better off simply skipping these games, and we could revisit this if users start reporting that games which they expect to be present in the Games List are missing.

I have already prototyped a patch based on Solution 1, creating a `PupguiGameListDialog#is_valid_lutris_gameslist_game` method that returns False if `game.runner`, `game.install_loc`, and `game.installed_at` are Falsey, as well as if the current runner is `steam`. But I'm opening this issue to get a bit of discussion going on the problem, to make sure I've put across what the problem is properly as it took me a little while to figure out the issue, and also to discuss the solution in case there's a better approach.

But regardless of where we land I'm happy to be the one to tackle this :-)

<hr>

In short, uninstalled Steam games can be in the Lutris DB file, and these are missing a bunch of information. Even though we filter out Steam games, they're missing a runner to do this filtering. We can make some assumptions to add this runner, but even then these games are missing a lot of information. And while this issue I have found is isolated to not-installed Steam games being missing this data, games from other runners could also be missing this information for any other reason, but it is likely just as relevant to ignore those games. If any game is missing this data we probably don't want to display it on the Games List, it just happens that for the purposes of this issue, I have only found this with uninstalled Steam games.

**To Reproduce**  
1. Take note of the Lutris sqlite DB content and size first (i.e. at `~/.local/share/lutris/pga.db`). For context, mine was around 1mb.
2. Connect your Steam Account to Lutris
4. Hover over the "Steam" source on the left and click on the "Refresh" button that appears
5. Sync can take a while, if Lutris crashes, try again
6. Inspect the Lutris sqlite DB (i.e. at `~/.local/share/lutris/pga.db`). My DB size increased by around 1.2mb, and now includes uninstalled Steam games
7. Open ProtonUp-Qt and try to open the Games List
8. There will be a crash for trying to run `.capitalize()` on 

**Expected behavior**  
Avoid this crash by handling games that are missing this data.

**Screenshots**  
The following backtrace is shown in the error dialog:
image

 - **Platform**: PC
 - **System**: Arch Linux
 - **Version**: db33194616ac32b986ec3e12929803c48f6de294
 - **How did you install ProtonUp-Qt?**: Running from source with `python3 -m pupgui2`

**Additional context**  
Not sure if this affects the CtInfo dialog. I think it probably wouldn't since it should rely on a runner being present in the first place and would skip if no runner is present? At least I haven't ran into any issues with the CtInfo dialog.

**Terminal output**  
The following backtrace is printed, and also shows up in the error dialog.
```python3
  File "/home/emma/Programming/ProtonUp-Qt/pupgui2/pupgui2.py", line 362, in btn_show_game_list_clicked
    gl_dialog = PupguiGameListDialog(install_directory(), self.ui)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/emma/Programming/ProtonUp-Qt/pupgui2/pupgui2gamelistdialog.py", line 41, in __init__
    self.setup_ui()
  File "/home/emma/Programming/ProtonUp-Qt/pupgui2/pupgui2gamelistdialog.py", line 53, in setup_ui
    self.setup_lutris_list_ui()
  File "/home/emma/Programming/ProtonUp-Qt/pupgui2/pupgui2gamelistdialog.py", line 94, in setup_lutris_list_ui
    self.update_game_list_lutris()
  File "/home/emma/Programming/ProtonUp-Qt/pupgui2/pupgui2gamelistdialog.py", line 196, in update_game_list_lutris
    runner_item = QTableWidgetItem(game.runner.capitalize())
                                   ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'capitalize'
```

sonic2kk:
Fixes #397.

## Overview
This PR fixes a crash in the Lutris Games List if some games have data as `None` that we expect to be defined.

I covered this a lot in #397 but to reiterate briefly: When loading the games list, we get our list of `LutrisGame`s with `lutrisutil#get_lutris_game_list`. We filter out Steam games from this list. Then we extract data from each `LutrisGame` and display it on the UI; the name, runner, install directory, and installation time, as well as some other data for tooltips.

However some games can be missing this information. Lutris stores information about its games in an sqlite3 database, `pga.db`, which we query. Some games in this DB may be missing a lot of information though. The example I encountered this with was by syncing my Lutris Steam list. It stored information about all my Steam games including ones that weren't installed, and the ones not installed had lots of empty information including an empty runner.

This missing runner meant the check we do to skip Lutris games gotten from Steam with `lutrisutil#is_lutris_game_using_runner` were not catching these games. Since the runner was `None` and not `'steam'` then it was missing this.

## Solution
The fix I ended up going with was to create a method, `PupguiGameListDialog#is_valid_lutris_gameslist_game`. We then use this in our list comprehension when storing our list of Lutris games.

### Stricter checking with `is_valid_lutris_gameslist_game`
The purpose of this function is to check if a `LutrisGame` is valid to be displayed on the Games List. In 99.9999% of cases, this is essentially just ensuring a game is _installed_, as an entry in Lutris' DB does not guarantee that a game is installed. We have to try an infer this (this is even a problem in Lutris itself, where many a missing game may not always be marked as missing, and will simply fail to launch). Lutris does have a column in its DB to track if a game is installed but this is not very reliable from what I have observed from querying, and from what I recall in the past, which is why we don't simply check this.

So now we have `is_valid_lutris_gameslist_game` that will essentially perform a stricter check to make sure games we want to display on the Games List have all the data we need. Instead of working around with `None` checks all over the place, we can assume early when generating the `self.games` list, that if a game is missing a `runner` and an `install_dir`, and if the runner is `'steam'`, then we don't want to display it on the games list.

A small benefit to this method is that if later on we decide we want other criteria, for example if Lutris started displaying Bottles or Heroic games for some reason, we can exclude those with an additional runner check. This keeps the list comprehension cleaner and also using this method name should make it a bit more obvious why the conditional check is in place.

### Safer Logic for Parsing `LutrisGame#installed_at`
Despite what I said above, I decided that a safer check for `installed_at` wasn't out of the question. Based on my extensive querying of my own Lutris DB in https://github.com/DavidoTek/ProtonUp-Qt/issues/397#issuecomment-2116195449, a game that is installed and valid for our purposes (in other words, a game that passes `is_valid_lutris_gameslist_game`) should always have `installed_at`. However in case this is ever missing, it is a less important factor in determining if a game is valid, so I kept it out of `is_valid_lutris_gameslist_game` and instead did a slight refactor of our logic for setting this.

Instead of setting all the data when setting up the `QTableWidgetItem`, we instead store the data in variables. Then we set the variables to different values based on whether  `installed_at` is Truthy (not `None` and not an empty string). If we have an `installed_at` then we set it up as before. Otherwise, we set the text to `self.tr('Unknown')`, set the tooltip to "`Install Date is Unknown`", and then set the data to `0`. The data is used for sorting to know how to sort by a given column, so we simply set to `0`. here.

After this check we set up the `QTableWidgetItem` as before but we give it the variables defined above. This keeps our conditional logic out of the widget item setup at the expense of some extra lines.

### Minor Type Hinting Refactor
We recently updated our base Python version to 3.10, meaning we can finally start to use the type hinting in Python3.10 without all the imports or capitalizations (https://docs.python.org/3.10/library/typing.html). Just because I was working in this file I decided to update our type hintings in `pupgui2gamelistdialog.py` to use things like `list` and `tuple` instead of `List` and `Tuple`, meaning we can also remove those imports as well.

We couldn't replace the `Callable` type hint, but you win some, you lose some :smile: 

<hr>

In my testing this isn't missing any games in the Lutris Games List, all the games I would expect to be there are present (there are games I know is missing, but they were displayed before, and Lutris also does not display them as missing, but the files do not exist anymore).

Let me know if you have any concerns about the implementation here or if we need to do any additional work in the `is_valid_lutris_gameslist_game` check. I know we had some discussion in #397 already but sometimes it's easier to look at an implementation and realise there are problems :sweat_smile:  

Thanks!

### DIFF
--- a/pupgui2/lutrisutil.py
+++ b/pupgui2/lutrisutil.py
@@ -38,9 +38,16 @@ def get_lutris_game_list(install_loc) -> List[LutrisGame]:
             lutris_install_dir = g[5]
             if not lutris_install_dir:
                 lg_config = lg.get_game_config()
-                working_dir = lg_config.get('game', {}).get('working_dir')
-                exe_dir = lg_config.get('game', {}).get('exe')
+                lg_game_config = lg_config.get('game', {})
+
+                working_dir = lg_game_config.get('working_dir')
+                exe_dir = lg_game_config.get('exe')
+
                 lutris_install_dir = working_dir or (os.path.dirname(str(exe_dir)) if exe_dir else None)
+
+                # If a LutrisGame config has an 'appid' in its 'game' section in its yml, assume runner is Steam
+                if lg_game_config.get('appid', None) is not None:
+                    lg.runner = 'steam'
 
             lg.install_dir = os.path.abspath(lutris_install_dir) if lutris_install_dir else ''
             lgs.append(lg)
--- a/pupgui2/pupgui2gamelistdialog.py
+++ b/pupgui2/pupgui2gamelistdialog.py
@@ -454,8 +454,9 @@ class PupguiGameListDialog(QObject):
     def is_valid_lutris_gameslist_game(self, game: LutrisGame) -> bool:
         """
         Check if a LutrisGame should be displayed on the Games List based on some criteria.
-        * Steam Games are excluded
-        * Games with no runner, install_loc, or installed_at are exluded
+        * Steam Games are excluded.
+        * Games with no runner are excluded.
+        * Games with no install_dir are excluded.
 
         Return Type: bool
         """
@@ -463,7 +464,7 @@ class PupguiGameListDialog(QObject):
         # Lutris DB may store Steam games even if they are not installed
         #
         # This results in only the game.name being available, and many
-        # other values like game.runner and game.install_loc being 'None'
+        # other values like game.runner and game.install_dir being 'None'
         #
         # We can assume a game is not valid to display on the Games List
         # if the runner and install_dir are Falsey, and if 
--- a/pupgui2/pupgui2gamelistdialog.py
+++ b/pupgui2/pupgui2gamelistdialog.py
@@ -180,13 +180,13 @@ class PupguiGameListDialog(QObject):
         """ update the game list for the Lutris launcher """
         # Filter blank runners and Steam games, because we can't change any compat tool options for Steam games via Lutris
         # Steam games can be seen from the Steam games list, so no need to duplicate it here
-        self.games = [game for game in get_lutris_game_list(self.install_loc) if not is_lutris_game_using_runner(game, 'steam')]
+        self.games = [game for game in get_lutris_game_list(self.install_loc) if self.is_valid_lutris_gameslist_game(game)]
 
         self.ui.tableGames.setRowCount(len(self.games))
 
         # Not sure if we can allow compat tool updating from here, as Lutris allows configuring more than just Wine version
         # It lets you set Wine/DXVK/vkd3d/etc independently, so for now the dialog just displays game information
-        for i, game in enumerate(self.games): 
+        for i, game in enumerate(self.games):
             name_item = QTableWidgetItem(game.name)
             name_item.setToolTip(f'{game.name} ({game.slug})')
             if game.installer_slug:
@@ -212,16 +212,25 @@ class PupguiGameListDialog(QObject):
                 runner_item.setToolTip(tooltip)
 
             # Some games may be in Lutris but not have a valid install path, though the yml should *usually* have some path
-            install_dir_text = game.install_dir or self.tr('Unknown')
+            install_dir_text = game.install_dir
             install_dir_item = QTableWidgetItem(install_dir_text)
             self.set_item_data_directory(install_dir_item, install_dir_text)
 
-            install_date = datetime.fromtimestamp(int(game.installed_at)).isoformat().split('T')
-            install_date_short = f'{install_date[0]}'
-            install_date_tooltip = self.tr('Installed at {DATE} ({TIME})').format(DATE=install_date[0], TIME=install_date[1])
+            # Populate Install Date column if we have game.install_date
+            # Otherwise set to some safe Unknown values
+            if game.installed_at:
+                install_date = datetime.fromtimestamp(int(game.installed_at)).isoformat().split('T')
+                install_date_short = f'{install_date[0]}'
+                install_date_tooltip = self.tr('Installed at {DATE} ({TIME})').format(DATE=install_date[0], TIME=install_date[1])
+                install_date_data = int(game.installed_at)
+            else:
+                install_date = self.tr('Unknown')
+                install_date_short = self.tr('Unknown')
+                install_date_tooltip = self.tr('Install Date is Unknown')
+                install_date_data = 0
 
             install_date_item = QTableWidgetItem(install_date_short)
-            install_date_item.setData(Qt.UserRole, int(game.installed_at))
+            install_date_item.setData(Qt.UserRole, install_date_data)
             install_date_item.setToolTip(install_date_tooltip)
             install_date_item.setTextAlignment(Qt.AlignCenter)
 
@@ -441,3 +450,31 @@ class PupguiGameListDialog(QObject):
                 return self.tr('Verified for {compat_tool}').format(compat_tool=deckt)
         else:
             return ''
+
+    def is_valid_lutris_gameslist_game(self, game: LutrisGame) -> bool:
+        """
+        Check if a LutrisGame should be displayed on the Games List based on some criteria.
+        * Steam Games are excluded
+        * Games with no runner, install_loc, or installed_at are exluded
+
+        Return Type: bool
+        """
+
+        # Lutris DB may store Steam games even if they are not installed
+        #
+        # This results in only the game.name being available, and many
+        # other values like game.runner and game.install_loc being 'None'
+        #
+        # We can assume a game is not valid to display on the Games List
+        # if the runner and install_dir are Falsey, and if 
+
+        if not game.runner:
+            return False
+
+        if not game.install_dir:
+            return False
+
+        if is_lutris_game_using_runner(game, 'steam'):
+            return False
+
+        return True
--- a/pupgui2/pupgui2gamelistdialog.py
+++ b/pupgui2/pupgui2gamelistdialog.py
@@ -1,7 +1,7 @@
 import os
 import pkgutil
 
-from typing import List, Callable, Tuple, Union
+from typing import Callable
 from datetime import datetime
 
 from PySide6.QtCore import QObject, Signal, Slot, QDataStream, QByteArray, Qt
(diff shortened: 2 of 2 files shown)


DavidoTek:
Thanks!

> A small benefit to this method is that if later on we decide we want other criteria, for example if Lutris started displaying Bottles or Heroic games for some reason, we can exclude those with an additional runner check

That would be a good addition if Lutris adds that. Good you kept that in mind! :smile:

> `installed_at` However in case this is ever missing, it is a less important factor in determining if a game is valid, so I kept it out of is_valid_lutris_gameslist_game and instead did a slight refactor of our logic for setting this.

Makes sense. Good, we should be safe then in case that ever happens.

> We recently updated our base Python version to 3.10, meaning we can finally start to use the type hinting in Python3.10 without all the imports or capitalizations (https://docs.python.org/3.10/library/typing.html).

Okay. It's probably the simplest to introduce such changes gradually as you did.
There shouldn't be anyone left running older versions of Python anymore now that the AppImage is updated. Arch/AUR is running the latest Python version anyway and I doubt any Pacstall installations are affected either.

@DavidoTek Note to myself: Remember to include a note about the updated Python requirement in the release notes.

> In my testing this isn't missing any games in the Lutris Games List, all the games I would expect to be there are present
> Let me know if you have any concerns about the implementation here or if we need to do any additional work in the is_valid_lutris_gameslist_game check. I know we had some discussion in #397 already but sometimes it's easier to look at an implementation and realise there are problems 😅

Looks good. I tested it with my (small) Lutris game library and everthing was displayed as expected. The code looks good as well.


---

Off topic: I'm wondering, did Lutris Flatpak change it's config location? I got an error and noticed that my lutris config is not stored in `config` but in `data`. Hmm.

```
$ python3 -m pupgui2
...
FileNotFoundError: [Errno 2] Datei oder Verzeichnis nicht gefunden: '/home/david/.var/app/net.lutris.Lutris/config/lutris/games'

$ ls /home/david/.var/app/net.lutris.Lutris/config/lutris/games
ls: cannot access '/home/david/.var/app/net.lutris.Lutris/config/lutris/games': No such file or directory. 

$ ls /home/david/.var/app/net.lutris.Lutris/data/lutris/games
grand-theft-auto-v-rockstar-games-launc-1706250034.yml
...
```

sonic2kk:
It's using `user_data_dir` as a fallback, see here:

https://github.com/lutris/lutris/blob/d3fd93dfe6d4017592c3296ae04d4d23e4dba98d/lutris/settings.py#L19-L26

It changes `CONFIG_DIR` to `DATA_DIR` if the Lutris `CONFIG_DIR` doesn't exist (i.e. for fresh installs I guess, which applies to a Flatpak).

Some Lutris Flatpak issues (https://github.com/flathub/net.lutris.Lutris/issues/422) are also pointing to `data/lutris`. On my system, A fresh Lutris Flatpak is also using `data/lutris`, but ProtonUp-Qt is able to pick it up fine (ProtonUp-Qt Flatpak and running from source).

DavidoTek:
> It's using user_data_dir as a fallback, see here:

Oh, not sure how I missed that. I should have ready two more lines of code :)


> On my system, A fresh Lutris Flatpak is also using `data/lutris`, but ProtonUp-Qt is able to pick it up fine (ProtonUp-Qt Flatpak and running from source).

Can you check if a `games` folder also exists in `config/lutris` on your system?
`lutrisutil.py#get_lutris_game_list` reads the game configuration for every installed game. The game configuration is read in `datastructures.py#LutrisGame:get_game_config`. It uses the `config_dir` from `constants.py`, which is hardcoded to `~/.var/app/net.lutris.Lutris/config/lutris`.
That means it should not be possible that it reads the game list when `config/lutris/games` does not exist. Or do we in fact check if the configuration is stored in `data/lutris` and overwrite the `config_dir` somewhere? Cannot remember.

https://github.com/DavidoTek/ProtonUp-Qt/blob/fcf0eaa66ae2d6e096be6e4df1b96183ac49c0f2/pupgui2/datastructures.py#L178-L185


I tried reinstalling Lutris.
The message changed a bit, now it is handled by the try-except of `get_lutris_game_list` (doesn't say `FileNotFoundError` anymore). Not sure why.

```
$ python3 -m pupgui2
...
Error: Could not get lutris game list: [Errno 2] Datei oder Verzeichnis nicht gefunden: '/home/david/.var/app/net.lutris.Lutris/config/lutris/games'
```

sonic2kk:
> Can you check if a `games` folder also exists in `config/lutris` on your system?

I checked and I only have a `games` folder under `~/.var/app/net.lutris.Lutris/data/lutris/games`.

Perhaps we should, for the time being, add support for both directories, as `config` is deprecated but may still be in use by existing installations.

This will likely also apply to the non-Flatpak Lutris installations as well. So we'd need to add whatever the XDG data path is as well.

PR #402, opened 2024-05-18
Theming: Use Base Palette Colour for Tooltip Background Instead of `Qt.white`

sonic2kk:
When using the Dark theme, the `ToolTipBase` colour was set to the same colour as the `ToolTipText` colour. Both were set to `Qt.white`. This meant the tooltip text would not be readable against the tooltip background.

This PR changes the `ToolTipBase` colour to match the Palette base colour (`12, 12, 12`), which is very close to pure black (`0, 0, 0`), making it contrast well against the tooltip text and making it readable. We could use either of these two, or `Qt.black`, but I opted to go with the window base colour. The difference is so minute my eyes cannot tell the difference :smile: 

#### Before (db33194)
image

#### After (This PR)
image

<hr>

Here are also some comparisons with other options, `Qt.black` and `QColor(0, 0, 0)`:

#### Comparison with `Qt.black` (NOT implemented in this PR)
image

#### Comparison with `QColor(0, 0, 0)` (NOT implemented in this PR)
image

<hr>

I caught this because I have switched to using the Dark theme. I put this down to Qt Wayland bugs with Plasma 5+Qt5, but now that we're onto Plasma 6+Qt6 with vastly improved Wayland support, the issue persisting felt a little strange, so I dug into it and this is what I found.

Thanks!

### DIFF
--- a/pupgui2/constants.py
+++ b/pupgui2/constants.py
@@ -65,12 +65,14 @@ POSSIBLE_INSTALL_LOCATIONS += [
 
 def PALETTE_DARK():
     """ returns dark color palette """
+    palette_base = QColor(12, 12, 12)
+
     palette_dark = QPalette()
     palette_dark.setColor(QPalette.Window, QColor(30, 30, 30))
     palette_dark.setColor(QPalette.WindowText, Qt.white)
-    palette_dark.setColor(QPalette.Base, QColor(12, 12, 12))
+    palette_dark.setColor(QPalette.Base, palette_base)
     palette_dark.setColor(QPalette.AlternateBase, QColor(30, 30, 30))
-    palette_dark.setColor(QPalette.ToolTipBase, Qt.white)
+    palette_dark.setColor(QPalette.ToolTipBase, palette_base)
     palette_dark.setColor(QPalette.ToolTipText, Qt.white)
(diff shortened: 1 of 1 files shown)


DavidoTek:
Thanks!

`(12, 12, 12)` seems like a valid choice as it matches with the base color.
I got that palette from somewhere, it seems like a gray that is slightly not black is preferred in UI design for some reason.

I've noticed that you added the line `palette_base = QColor(12, 12, 12)`. I am wondering whether it would make sense to do the same for the other colors to maintain consistency. On the other hand, `Base` and `ToolTipBase` are related while e.g. `Window` and `Button` are probably not.
I will merge it for now, we can look into that in case we need to introduce additional colors at some point.

sonic2kk:
> it seems like a gray that is slightly not black is preferred in UI design for some reason.

Heh, yup, I'm not sure why this is but I also heard this trick from a friend a while back too. Maybe they just look less "harsh" and that extra buffer from being absolutely any colour makes all the difference somehow 😄 

> I've noticed that you added the line `palette_base = QColor(12, 12, 12)`. I am wondering whether it would make sense to do the same for the other colors to maintain consistency. 

That is something we could look into. Maybe there's something we can borrow from other Qt projects and how they handle their theming :-)

PR #404, opened 2024-05-24
Bump Python version requirement to 3.10

DavidoTek:
- `setup.cfg`: `python_requires = >=3.6` -> `python_requires = >=3.10`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -18,7 +18,7 @@ classifiers =
 
 [options]
 include_package_data = True
-python_requires = >=3.6
+python_requires = >=3.10
 install_requires =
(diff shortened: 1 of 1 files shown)


sonic2kk:
I feel like part of this is my fault :sweat_smile: But makes sense, Python 3.10 has been out for a while now, and anyone running from source should hopefully be using an up-to-date Python version.

DavidoTek:
It's all ok.
I guess it is fine to move to a newer Python version because 3.8 will be EOL by the end of this year.
As the Flatpak runtime and AppImage use Python 3.10, we may now use all of its features.

PR #413, opened 2024-06-23
Readme: Add warning about fake Proton-GE website that references ProtonUp-Qt

sonic2kk:
There is a fake website that has been around for a little while now claiming to be the website for GE-Proton. Most, if not all, of the content on the site is more than likely AI generated.

[GloriousEggroll added a big disclaimer to the top of their Readme about it](https://github.com/GloriousEggroll/proton-ge-custom?tab=readme-ov-file#myself-gloriouseggroll-and-this-project-proton-ge-are-not-affiliated-with-httpsprotongecom-that-is-a-spamfake-website-there-is-no-existing-website-for-proton-ge-other-than-this-github-repository-proton-ge-does-not-collect-any-user-data-what-so-ever-and-is-not-a-company-or-organization-of-any-type). However when visiting the site from the Readme I noticed that some pages actually reference ProtonUp-Qt. This means a user could miss the warning on the GE-Proton readme.

The warning at the moment is a bit "low key". I didn't want to do something quite as loud as what is on the GE-Proton repo, not because it isn't warranted, but just as a matter of organization. If you'd like a warning that loud though we can do it. Perhaps a "Note" / "Disclaimer" subheading to draw attention and putting this warning under it would be a good idea.

I also mentioned that other compatibility tools are not affiliated with this site, including GE-Proton.

<hr>

We might also want to add the same warning to the Flatpak, once we settle on wording and visibility. While the site directs users on some pages to the repository it also mentions going to the Software Center Frontends, so putting the warning there as well means users are more likely to see it and become aware that the site is fake.

I don't really know what malicious things this site might do, but I think adding a note to the Readme to distance this project from it is a good idea. Although the other side of this is that we could end up drawing more attention to the site by doing this. 

It is up to you whether we put this warning in, and how we format it. I just thought it would be good to get it out there that this site isn't affiliated with your project :-)


DavidoTek:
Thanks for letting me know, I wasn't even aware that this site exists!

I don't think we should add a warning that specific, it only gives them more attention. What we can do, however, is to add a "generic" disclaimer regarding affiliation with other compatibility tools/websites.

_What about_:

**[Affiliation Note](https://github.com/DavidoTek/ProtonUp-Qt/pull/413)**: ProtonUp-Qt is an independent tool for managing gaming compatibility tools. It is neither directly affiliated with the compatibility tool creators nor with the providers of the individual game launchers. However, we try to work with them where possible. The official development takes place at https://github.com/DavidoTek/ProtonUp-Qt and the official website is https://davidotek.github.io/protonup-qt/. We distribute ProtonUp-Qt on Flathub and as the AppImage in the releases section. Additionally, we check the integrity of the AUR and [Pacstall](https://pacstall.dev/packages/protonup-qt-app) distribution on an irregular basis.

sonic2kk:
I think a more general note is a great idea, I didn't even consider that! We can also include that on the Flatpak page.

And indeed giving more credit to the services used is also a good idea.

<hr>

The disclaimer you wrote is perfect imo, I will put that into the Readme. We could put that part under its own heading, a smaller one under the main section? I would normally suggest to put it just above the "Licensing" section, but that *might* hurt visibility.

Putting further credit to the services used could go above the Licensing section. Maybe an "Acknowledgements" section, although that might be a bit too formal :-)

<hr>

I will push a commit with the changes mentioned and we can continue from there.

sonic2kk:
Pushed the changes, I didn't add the Steam API because it seems like the `STEAM_API_GETAPPLIST_URL` constant isn't actually used.

PR #418, opened 2024-06-29
constants: Use constant to determine if Flatpak

sonic2kk:
We use a lot of `os.path.exists('/.flatpak-info')` calls to determine if we're running inside Flatpak. This can be stored as a constant. Maybe you could consider this a micro-optimisation as we don't call `os.path.exists` as much anymore, but my main motivation here was readibility.

I didn't test this in Flatpak, please feel free to do so to make sure I didn't screw anything up (or direct me on how to do it :smile:)

Thanks!

<hr>

P.S. - This touches the dreaded SteamTinkerLaunch ctmod :wink: I have a laptop again (that I installed SteamTinkerLaunch on using ProtonUp-Qt!) so I am planning to tinker around with that ctmod again and make the code a bit less scary.


DavidoTek:
Thanks!

That makes it very clean. If we ever need it, we could add the variables `IS_APPIMAGE` and `IS_SNAP` in the future.

PR #419, opened 2024-06-29
Flatpak Metainfo: Add Affiliation Note

sonic2kk:
Adds the affilication note about the fake website to the Flatpak Metainfo which should also, if my understanding is correct, make it appear on the Flathub page (and possibly elsewhere that can pull this data, such as on the commandline?).

I think I got it right in terms of styling but I haven't ever written anything like this before. I am unsure if the links for example work this way, and if the markup is fairly standard, or if there are some Flatpak (or more specifically, Flathub) specific guidelines in place for excessive linking/linking at all.

### DIFF
--- a/share/metainfo/net.davidotek.pupgui2.appdata.xml
+++ b/share/metainfo/net.davidotek.pupgui2.appdata.xml
@@ -50,6 +50,16 @@
     </ul>
     <p>Double click compatibility tools to show more information and games using the tool.</p>
     <p>Advanced settings and the color theme can be managed in the About dialog.</p>
+
+    <hr/>
+
+    <p>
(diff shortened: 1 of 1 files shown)


DavidoTek:
Thanks!

> Adds the affilication note about the fake website to the Flatpak Metainfo which should also, if my understanding is correct, 

Hmm. I think it is enough for now that we have added it to the GitHub page. If people found their way to the Flatpak page, there isn't much they can do wrong anymore.

> make it appear on the Flathub page (and possibly elsewhere that can pull this data, such as on the commandline?).

It's displayed on the Flathub page and inside the distros app store (GNOME Software, KDE Discover, ...). Not sure if that can also be done using the cli.

> I am unsure if the links for example work this way,

The `<a></a>` tag is unsupported by appstream. I think only plaintext URLs are allowed. I have seen somewhere that even this can be a warning if I remember correctly.

---

I will close this PR for now as I don't think we absolutely need it at the moment and also only plaintext URLs are allowed in the Flatpak description.

PR #421, opened 2024-07-06
util: Don't download AWACY list if offline

sonic2kk:
When troubleshooting to see if another issue was related to offline behaviour, I found out that if we open ProtonUp-Qt while offline, there is an ugly error coming from requests if we try to download the game list while offline.

This PR returns early from `download_awacy_gamelist` if we are offline. We do this before we even create the thread that tries to download the game list, which should be very, very slightly more efficient than doing this after the thread is created :sweat_smile:

<hr>

This has the side-effect of not creating the file at all if we open ProtonUp-Qt while offline. We use this file in `update_steamapp_awacystatus` to populate `SteamApp` anti-cheat info and we have a check that if the file doesn't exist, we return early. So no worries here.

Something to consider for the future is that the AWACY list is not updated if we don't create it on startup. Since the file won't be populated while offline, even if the user re-connects later, we don't re-download the file. We might want to consider doing this, but I am not sure how. Just a note from being in and around this part of the code :smile:  

<hr>

Since this calls out directly to the raw file data we shouldn't have to worry about any GitHub rate limits. But we might want to look in future at making `_download_awacy_gamelist_thread` a bit safer, to avoid writing any data if we don't get a success code, just to avoid writing out bad data to the file.

This `update_steamapp_awacystatus` function does account for the file not having the fields we expect by wrapping the logic in a try/except block, so we raise an exception if we do have a file but it has bad data (so if we ever end up writing garbage data like a rate-limit or GitHub-down response it'll be safe).

<hr>

### DIFF
--- a/pupgui2/util.py
+++ b/pupgui2/util.py
@@ -456,6 +456,10 @@ def download_awacy_gamelist() -> None:
         r = requests.get(AWACY_GAME_LIST_URL)
         with open(LOCAL_AWACY_GAME_LIST, 'wb') as f:
             f.write(r.content)
+
+    if not is_online():
+        return
+
     t = threading.Thread(target=_download_awacy_gamelist_thread)
     t.start()
 


DavidoTek:
Thanks!

I wonder if it would make more sense to catch the error directly instead of calling `is_online`. We're making a request anyway, so doing another is unnecessary.
Also, if `api.github.com` is reachable, but `raw.githubusercontent.com` is not, we will get an error nevertheless. Not sure if that will ever happen. (Just came to my mind: We're doing the online check using GitHub for the ctmods, but it is in theory possible that the user is online, but Git*Lab* is down. Hmm)

I haven't reviewed https://github.com/DavidoTek/ProtonUp-Qt/pull/423 yet, but I noticed that you included a wrapper for `requests.get` which handles errors. Maybe we should use this? Although that would mean we need to pass an empty lambda function to the `progress_callback` parameter, so a simple `try-except` may be simpler. I will leave this open for discussion.

---

> Since the file won't be populated while offline, even if the user re-connects later, we don't re-download the file

Well, it should be created after restarting ProtonUp-Qt, right? I think that is fine.

> Since this calls out directly to the raw file data we shouldn't have to worry about any GitHub rate limits. But we might want to look in future at making _download_awacy_gamelist_thread a bit safer, to avoid writing any data if we don't get a success code, just to avoid writing out bad data to the file.
> This update_steamapp_awacystatus function does account for the file not having the fields we expect by wrapping the logic in a try/except block, so we raise an exception if we do have a file but it has bad data (so if we ever end up writing garbage data like a rate-limit or GitHub-down response it'll be safe).

We can do that in the future, but seems fine for now if it doesn't cause any errors. The file will be re-downloaded the next time ProtonUp-Qt is started and if there isn't any rate-limit anymore then, it should work again.

sonic2kk:
Ah the first half is a great point. I think using `download_file` from that PR would make sense, with some caveats:
- I just noticed that the function from that PR only catches an `OSError`, which was all we caught before in ctmods. But we don't handle failed requests or timeouts, like we do in `is_online`. So before we use `download_file` we would need to catch these exceptions. That would entirely negate the need for our `is_online` check in this context, like you said we're already making the request, we can just ping.
- For the empty lambda function, I did mess around with making it optionally `None`. We can do that as long as we do the check beforehand, but writing `if progress_callback: progress_callback()` just looked messy. But we could absolutely do it, it would make the function usable in this context instead of passing an empty lambda, which makes the usage a bit messy, whereas making the lambda optional only adds length to the implementation.

So `download_file` might need a bit of a refactor, but it should work for this as well, yes. We can hold off on this PR and wait until we can use `download_file`, then implement it here.

> Well, it should be created after restarting ProtonUp-Qt, right? I think that is fine.

Yup, restarting will re-download the file.

> We can do that in the future, but seems fine for now if it doesn't cause any errors. The file will be re-downloaded the next time ProtonUp-Qt is started and if there isn't any rate-limit anymore then, it should work again.

Yes I believe the file will be re-created when ProtonUp-Qt starts again. At least on my system in `/tmp/pupgui2.blahblah` that `pupgui2` folder is removed when ProtonUp-Qt closes. I assume this happens in the Flatpak environment as well? And regardless the file contents should be overwritten with the latest request the next time ProtonUp-Qt opens.

<hr>

> Also, if api.github.com is reachable, but raw.githubusercontent.com is not, we will get an error nevertheless. Not sure if that will ever happen.

I think this is safe, `githubusercontent` used to/maybe still does host images, so I am not sure if only one can go down. I think this one is safe. However:

(Just came to my mind: We're doing the online check using GitHub for the ctmods, but it is in theory possible that the user is online, but GitLab is down. Hmm)

This is a great point, especially when considering that a ctmod in future could potentially use a separate GitLab instance. GitHub and GitLab could be down, but this one specific GitLab instance might be available.

I think the logic to check against the GitHub API was because at the time, we *only* had GitHub. So we would ping to see if we could access the one service we called out to, as if we can't access GitHub, we *are* effectively offline. But now that we have GitLab... I'm not sure what the best course of action is.
- I don't think changing to check a generic host works, because we still really only care about the sites we use. Showing as online but being unable to download tools might be strange to a user.
- I am not sure if changing the check to check for GitHub and GitLab is a good idea either:
    - This would lengthen the online check.
    - This would become a bit of a maintenance burden, if we ever added ctmods hosted on other platforms we'd have to account for that too.
    - We would have to either remove the (unused) optional `host` value and just hardcode the online check to specific APIs, *or* return whether ONLY the host can be reach if specified, otherwise check whether GitHub and any known GitLab instance can be reached 
    - This would end up making a lot of requests...

It hasn't caused issues for now, and a user can always click on the CtInfo's "Info" button to see if a tool's page is unavailable. If it gives a 404 that should let them know there's an issue with that specific page.

So perhaps keeping this in mind as a "known limitation" for now is fine :-)

DavidoTek:
> Ah the first half is a great point. I think using download_file from that PR would make sense, with some caveats:

Okay. I thought about it and it might actually make it more complicated.
`is_online()` will handle any Timeout/Connection related issues. A 404 doesn't cause any harm.
**In short: I will merge this PR as it is.** If we ever want to use the download function from networkutil in the future, we can still change that at a later point.

> I think this is safe, githubusercontent used to/maybe still does host images, so I am not sure if only one can go down. I think this one is safe. However:

Okay, then `is_online()` should be fine.
I also doubt that the `githubusercontent.com` will become unregistered in the near future, causing a ConnectionError.

> > (Just came to my mind: We're doing the online check using GitHub for the ctmods, but it is in theory possible that the user is online, but GitLab is down. Hmm)
> This is a great point, especially when considering that a ctmod in future could potentially use a separate GitLab instance. GitHub and GitLab could be down, but this one specific GitLab instance might be available.

Okay. Then we can look into that in the future.

Issue #415, opened 2024-06-28
`appinfo.vdf` Magic Changed on Steam Client Beta (June 2024) - Needs Resolved Upstream in `ValvePython/steam`


sonic2kk:
Relevant upstream PRs:
- ValvePython/vdf#61
- ValvePython/steam#464 (depends on above VDF PR)

<hr>

Please fill out following when reporting a new bug:

**Describe the bug**  
The upstream magic for `appinfo.vdf` changed again on the latest Steam Client Beta, meaning `parse_appinfo` from [ValvePython/steam](https://github.com/ValvePython/steam) is broken. This happened before (ValvePython/steam#418) and this new issue has already been reported upstream (ValvePython/steam#462).

I am making this issue for awareness for anyone who runs into this, and also for us to track this so that we can update the Steam dependency once a version with a fix is available.

**To Reproduce**  
1. Open ProtonUp-Qt while on the latest Steam Client Beta
2. With Steam selected, it will open significantly faster than normal
3. Open the Game List
4. Only Steam Shortcuts will be listed

**Expected behavior**  
ProtonUp-Qt should be able to parse the `appinfo.vdf`, but cannot pending an upstream fix.

**Screenshots**  
N/A.

**Desktop (please complete the following information):**  
 - **Platform**: PC
 - **System**: Arch Linux
 - **Version**: 2.9.2, but hapepens on any version
 - **How did you install ProtonUp-Qt?**: Discover

**Additional context**  
Took a stab at trying to submit a patch for this upstream, but no guarantees my fix will be the one that gets merged as it is still in progress :smile: 

**Terminal output**  
The `parse_appinfo` gives the following output because it cannot parse the magic in the file.
```
Error getting ctool map from appinfo.vdf: Invalid magic, got b')DV\x07'
Error updating SteamApp info from appinfo.vdf: Invalid magic, got b')DV\x07'
```

The fix is not as straightforward as adding the magic to the file though, as we need to skip past some parts of the file. And then there are some issues around parsing the binary VDF information in this file under `data`.

sonic2kk:
A small effort by me and a much larger effort by the Protontricks maintainer have created a working implementation *however* changes are required for both ValvePython/steam and ValvePython/vdf. So we will need to bump both versions.

sonic2kk:
[...]

PR #422, opened 2024-07-08
ctmod: Remove D8VK

sonic2kk:
D8VK has now, at long last, been merged into DXVK (doitsujin/dxvk#3411, [D8VK readme](https://github.com/AlpyneDreams/d8vk/blob/c8bbb32ab2b6184b17d6ffd23bed03c48463321c/README.md))! There is no need for us to offer D8VK anymore, as DXVK Nightly will cover the basis here. We hooked into D8VK CI anyway because at the time there were no releases, and we never added the stable releases. Plus, there hadn't been any commits in so long since it was waiting on upstream DXVK to review and merge, so all CI artifacts had expired anyway...

I updated the DXVK ctmod descriptions to note that it supports Direct3D 8, 9, 10, and 11 so that anyone looking for D8VK and was not aware of the change should hopefully see that description change and know that DXVK now supports Direct3D 8. We can also note on the changelog for the next release that D8VK is now available upstream.

**Note**: The description update is *technically* only accurate for DXVK Nightly right now. Support for Direct3D 8 is not yet in a release at time of writing, although I imagine there will be a release very soon. And the DXVK Async patch gets updated whenever there is a new release of DXVK, so it'll be updated afterwards too. Personally I think it's fine to be a little "ahead" here in the interest of helping users migrate from D8VK to DXVK,, instead of being under the impression that D8VK was removed, but I am happy to change it if you think this is too misleading :-)

Thanks!


DavidoTek:
Thanks, that was a fast PR!

> D8VK has now, at long last, been merged into DXVK

I read the news on sunday and was excited!

> I updated the DXVK ctmod descriptions to note that it supports Direct3D 8, 9, 10, and 11 so that anyone looking for D8VK and was not aware of the change should hopefully see that description change and know that DXVK now supports Direct3D 8. We can also note on the changelog for the next release that D8VK is now available upstream.

Great. Yes, we should include a note in the changelog.

> Note: The description update is technically only accurate for DXVK Nightly right now. Support for Direct3D 8 is not yet in a release at time of writing, although I imagine there will be a release very soon

I think it makes sense to include this once there is a new release out. I've added a reminder to my calender in 5 days to check if there will be a release of DXVK then. It's probably also time for a new ProtonUp-Qt release soon :smile:

---

@DavidoTek PR is ready to merge once a new version of DXVK has been released: https://github.com/doitsujin/dxvk/releases

sonic2kk:
> I've added a reminder to my calender in 5 days to check if there will be a release of DXVK then

I will also keep an eye out. I think it's fine to wait for a new DXVK release. I don't know if we need to wait on a new DXVK Async release, but I imagine Async is much less important for Direct3D 8 :wink: 

> It's probably also time for a new ProtonUp-Qt release soon 😄

Heh, that would be great :smile: But! Maybe we should wait to see the progress on #415 ? It's not our bug, but if it's merged upstream, we can bump our dependency version and have it fixed without needing to do a hotfix-release.

sonic2kk:
They made that release faster than I expected 😄

PR #392, opened 2024-05-06
Change AppImage PYTHONPATH to local lib

DavidoTek:
Fix https://github.com/DavidoTek/ProtonUp-Qt/issues/390

Change `PYTHONPATH` from `${APPDIR}/usr/lib/python3.10/site-packages` to `${APPDIR}/usr/local/lib/python3.10/dist-packages`.

I need to check whether that really fixes it, marked as draft for now.

PR #426, opened 2024-07-27
Use steam and vdf dependency by  Solstice Game Studios

DavidoTek:
Fix https://github.com/DavidoTek/ProtonUp-Qt/issues/415

Use `steam` and `vdf` dependency by  Solstice Game Studios:

- https://github.com/solsticegamestudios/vdf (`master` branch)
- https://github.com/solsticegamestudios/steam (`master` branch)

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ ProtonUp-Qt|GPL-3.0
 [PySide6](https://pypi.org/project/PySide6/)|LGPL-3.0/GPL-2.0
 [inputs](https://pypi.org/project/inputs/)|BSD
 [pyxdg](https://pypi.org/project/pyxdg/)|LGPLv2
-[vdf](https://pypi.org/project/vdf/)|MIT
+[vdf@solstice](https://github.com/solsticegamestudios/vdf/)|MIT
+[steam@solstice](https://github.com/solsticegamestudios/steam/)|MIT
 [requests](https://pypi.org/project/requests/)|Apache 2.0
 [PyYAML](https://pypi.org/project/PyYAML/)|MIT
--- a/pupgui2/steamutil.py
+++ b/pupgui2/steamutil.py
@@ -257,7 +257,7 @@ def _get_steam_ctool_info(steam_config_folder: str) -> Dict[str, Dict[str, str]]
     compat_tools = {}
     try:
         with open(appinfo_file, 'rb') as f:
-            header, apps = parse_appinfo(f)
+            header, apps = parse_appinfo(f, mapper=dict)
             for steam_app in apps:
                 if steam_app.get('appid') == 891390:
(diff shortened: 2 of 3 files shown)


sonic2kk:
Out of interest, do we want to point to the `master` branch, or do we want to point to specific tags? For example the [v3.5 tag](https://github.com/solsticegamestudios/vdf/releases/tag/v3.5).

We can point to the tag with, for example `git+https://github.com/solsticegamestudios/vdf@v3.5`.

This can help with consistency and making sure the same version is used when, i.e. the Flatpak is built, the AppImage is built, and for other packages. This will also make bumping the dependency version easier.

But maybe you already considered this and there's a reason to point directly to `master`? I'm not used to maintaining dependencies so there might be a good reason why tags weren't used :smile:

sonic2kk:
I left a comment on the AUR packages for ProtonUp-Qt, [`protonup-qt`](https://aur.archlinux.org/packages/protonup-qt#comment-984045) and `[protonup-qt-bin](https://aur.archlinux.org/packages/protonup-qt-bin#comment-984047)`, to let them know that we are using a fork of these dependencies now. It probably doesn't apply *as much* to `protonup-qt-bin` but I wanted to let them know just in case (not a package maintainer, don't want to risk making faulty assumptions).

DavidoTek:
> Out of interest, do we want to point to the master branch, or do we want to point to specific tags? For example the v3.5 tag.
> We can point to the tag with, for example git+https://github.com/solsticegamestudios/vdf@v3.5.

Good point.
I thought about that too. Our convention until now was to only specify a minimum version in the `requirements.txt` (e.g., `requests>=2.27.0`). I think I originally put it in because people are installing it directly on the system (AUR, Pacstall, dev build without venv).
This also means we do not need to worry about updating the dependencies. But I get the point, if there are changes breaking compatibility (like suddently requiring the `mapper=dict` parameter), they will appear without notice and it can be harder to track the issue as there can be inconsistencies between the Flatpak and AppImage build.

In case something like this happens, we may consider changing it from master to a fixed version. I think the risk of waiting for such an event is low as existing builds are not affected.

> I left a comment on the AUR packages for ProtonUp-Qt, protonup-qt and protonup-qt-bin, to let them know that we are using a fork of these dependencies now.

Great, thanks.

I just released a new version (`v2.10.0` :tada:), but apparently the new AppImage build (https://github.com/DavidoTek/ProtonUp-Qt/pull/359, https://github.com/DavidoTek/ProtonUp-Qt/pull/382, https://github.com/DavidoTek/ProtonUp-Qt/pull/392, https://github.com/DavidoTek/ProtonUp-Qt/pull/430) is still broken. AUR can use it just fine though and Flatpak too (there just isn't any "AppImage Beta Testing" right now).

PR #432, opened 2024-07-28
CID: Add WineZGUI

sonic2kk:
WineZGUI was missing from the Custom Install Dialog dropdown. Sorry that I didn't find this in time for v2.10.0 (v.2.10.2?)'s release :sweat_smile:


DavidoTek:
Thanks!

PR #434, opened 2024-07-28
test_util: Improve type hints for `test_get_random_game_name`

sonic2kk:
Improve type hints for the unit test `test_get_random_game_name`, which also leads to improving the type hints for the `util#get_random_game_name` function, which was incorrect (it was hinting for a single list which could have Steam Games, Lutris Games, and/or Heroic Games, instead of a list which would _only_ contain one of these types rather than any combination).

This also means we get rid of some deprecated types in `util.py`.

Adding hints to unit tests and the functions they are testing is a good pairing imo; we are improving the hints of the actual function, and being explicit about the types we are expecting in our tests.

### DIFF
--- a/pupgui2/util.py
+++ b/pupgui2/util.py
@@ -883,14 +883,14 @@ def create_missing_dependencies_message(ct_name: str, dependencies: List[str]) -
     return tr_msg, False
 
 
-def get_random_game_name(games: List[Union[SteamApp, LutrisGame, HeroicGame]]) -> str:
+def get_random_game_name(games: list[SteamApp] | list[LutrisGame] | list[HeroicGame]) -> str:
     """ Return a random game name from list of SteamApp, LutrisGame, or HeroicGame """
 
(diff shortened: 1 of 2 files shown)


DavidoTek:
Thanks!

> it was hinting for a single list which could have Steam Games, Lutris Games, and/or Heroic Games, instead of a list which would only contain one of these types rather than any combination

Good catch.

> `random_game: SteamApp | LutrisGame | HeroicGame`

I wonder if we want to add type definitions for this to `constants.py`/`types.py ?` in the future. Something like `AnyGame = SteamApp | LutrisGame | HeroicGame`.

sonic2kk:
I like that idea, we may also want to do this for other things like parsing JSON and VDF files (arguably both of these should be upstream and perhaps we can ask upstream VDF to improve type hinting if possible).

Definitely room to improve our typing 😄 Not that it's majorly urgent, just something that can be worked on overtime

PR #444, opened 2024-08-12
heroic_util: Remove deprecated type hints

sonic2kk:
This PR removes the deprecated type hints `List` and `Dict` imports from `typing` and replaces them with standard usage of `list` and `dict` that are available since Python 3.9. It also introduces some further type hints to other parts of the util file, although incomplete as typing some parts of this is a little complex given our reliance on parsing JSON (ideally we'll eventually have a JSON type alias, and we will more specifically hint in places where we know the structure of the JSON coming back, like if we know certain objects that we're performing `.get` on are always expected to be `dict[str, str]` for example). The rationale for further type hinting in the file is to make it easier for unit tests once we eventually get to that stage.

### DIFF
--- a/pupgui2/heroicutil.py
+++ b/pupgui2/heroicutil.py
@@ -2,26 +2,24 @@ import os
 import json
 import re
 
-from typing import List, Dict
-
 from pupgui2.datastructures import HeroicGame
 from pupgui2.constants import EPIC_STORE_URL
 
 
-def get_heroic_game_list(heroic_path: str) -> List[HeroicGame]:
+def get_heroic_game_list(heroic_path: str) -> list[HeroicGame]:
     """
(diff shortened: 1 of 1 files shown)


DavidoTek:
Thanks!

> although incomplete as typing some parts of this is a little complex given our reliance on parsing JSON (ideally we'll eventually have a JSON type alias, and we will more specifically hint in places where we know the structure of the JSON coming back, like if we know certain objects that we're performing .get on are always expected to be dict[str, str] for example).

If we know that the returned type will be `dict[str, str]` we can use that. The only problem is that the JSON could potentially change and the data could be `dict[str, Any]` (or even `Any`).
Ideally, we would check if the type is correct before returning it. Not sure if it is worth the effort (I would like if Python had something like https://zod.dev integrated... :smile:)
-> I think it is okay to type-hint `dict[str, str]` in the future if we assume that is the expected type.

PR #451, opened 2024-08-31
pupgui2: Display install location combobox text in tooltip

sonic2kk:
Adds a tooltip to the installation location dropdown `comboInstallLocation` on the Main Menu which displays the dropdown's text. This can be useful if the path in the dropdown text is long, usually for paths. Resizing the window to see the full path is awkward, and not possible in some scenarios (like in GameScope). Some paths are so long that they are truncated in the dropdown itself at the Main Window's default size.

Steam

Heroic

One thing not implemented here that might be cool to display is if the launcher path is custom. We could maybe add `[CUSTOM]` somewhere in the tooltip. But I think making it more obvious when a chosen launcher path is custom is a general enhancement for the future, so we could tackle it then.

### DIFF
--- a/pupgui2/pupgui2.py
+++ b/pupgui2/pupgui2.py
@@ -296,6 +296,10 @@ class MainWindow(QObject):
         self.ui.txtUnusedVersions.setText(self.tr('Unused: {unused_ctools}').format(unused_ctools=unused_ctools) if unused_ctools > 0 else '')
         self.ui.txtInstalledVersions.setText(f'{len(self.compat_tool_index_map)}')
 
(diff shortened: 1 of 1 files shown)


DavidoTek:
Thanks!

> One thing not implemented here that might be cool to display is if the launcher path is custom. We could maybe add [CUSTOM] somewhere in the tooltip. But I think making it more obvious when a chosen launcher path is custom is a general enhancement for the future, so we could tackle it then.

Yeah, it would make sense to also display it in the combo box itself, not only in the tooltip.

Issue #424, opened 2024-07-23
ProtonUp-Qt Flatpak ver. 2.9.2 - UnicodeDecodeError


Megatron-X:
Please fill out following when reporting a new bug:

**Describe the bug**  
When I launch ProtonUp-Qt (Flatpak) I keep getting a UnicodeDecodeError. I also tried to launch it in terminal and still get the UnicodeDecodeError. Prior to a couple of days, I didn't have this problem at all and was able to launch it. 

**To Reproduce**  
1. Launch ProtonUp-Qt either by Icon or Terminal.
2. See Screenshot and Terminal output

**Screenshots**  
![Captured on 2024-07-22 21-51-02](https://github.com/user-attachments/assets/24cb8894-b7bb-4d4a-8716-2a58491619ad)

**Desktop (please complete the following information):**  
 - **Platform**: PC
 - **System**: Arch Linux
 - **Version**: ProtonUp-Qt 2.9.2
 - **How did you install ProtonUp-Qt?**: GNOME Software

**Terminal output**  
```
ProtonUp-Qt 2.9.2 by DavidoTek. Build Info: DavidoTek Flathub build.
Python 3.11.9 (main, Nov 10 2011, 15:00:00) [GCC 13.2.0], PySide 6.7.0
Platform: KDE Flatpak runtime 6.7 Linux-6.6.40-1-lts-x86_64-with-glibc2.38
Loading locale es / es_ES
Loaded ctmod GE-Proton
Loaded ctmod Wine-GE
Loaded ctmod Boxtron
Loaded ctmod D8VK (nightly)
Loaded ctmod Kron4ek Wine-Builds Vanilla
Loaded ctmod Lutris-Wine
Loaded ctmod Luxtorpeda
Loaded ctmod Northstar Proton (Titanfall 2)
Loaded ctmod Proton Tkg
Loaded ctmod Proton Tkg (Wine Master)
Loaded ctmod Roberta
Loaded ctmod Steam-Play-None
cat: /etc/lsb-release: No existe el fichero o el directorio
Loaded ctmod SteamTinkerLaunch
cat: /etc/lsb-release: No existe el fichero o el directorio
Loaded ctmod SteamTinkerLaunch-git
Loaded ctmod vkd3d-lutris
Loaded ctmod vkd3d-proton
Loaded ctmod Wine Tkg (Valve Wine Bleeding Edge)
Loaded ctmod Wine Tkg (Wine Master)
[...]

PR #455, opened 2024-09-12
Add vdf_safe_load function

DavidoTek:
Fix https://github.com/DavidoTek/ProtonUp-Qt/issues/424

Adds a new function `vdf_safe_load(vdf_file: str) -> dict` to `steamutil.py`

It opens and parses vdf files using the vdf library. Defective UTF-8 characters are replaced with a unicode question mark symbol. The error is printed in case of other Exceptions and an empty dict is returned.

### DIFF
--- a/pupgui2/datastructures.py
+++ b/pupgui2/datastructures.py
@@ -150,7 +150,9 @@ class BasicCompatTool:
         """
         compat_tool_vdf_path = os.path.join(self.install_dir, self.install_folder, 'compatibilitytool.vdf')
         if os.path.exists(compat_tool_vdf_path):
-            compat_tool_vdf = vdf.load(open(compat_tool_vdf_path))
+            # TODO: Move this somewhere else and use steamutil.py#vdf_safe_load
+            with open(compat_tool_vdf_path, 'r', encoding='utf-8', errors='replace') as f:
+                compat_tool_vdf = vdf.loads(f.read())
             if 'compatibilitytools' in  compat_tool_vdf and 'compat_tools' in compat_tool_vdf['compatibilitytools']:
                 return list(compat_tool_vdf['compatibilitytools']['compat_tools'].keys())[0]
 
--- a/pupgui2/steamutil.py
+++ b/pupgui2/steamutil.py
@@ -58,8 +58,8 @@ def get_steam_app_list(steam_config_folder: str, cached=False, no_shortcuts=Fals
     apps = []
 
     try:
-        v = vdf.load(open(libraryfolders_vdf_file))
-        c = get_steam_vdf_compat_tool_mapping(vdf.load(open(config_vdf_file)))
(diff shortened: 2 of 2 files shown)


sonic2kk:
Is this a bug upstream out of interest? Why exactly is this needed? Just wondering as I have never ran into this issue, and have some games with emojis etc in their names.

I remember a similar issue was related to incorrectly configured system locales. In the past I have also fallen victim to application issues caused by locale problems.

I saw the discussion in the linked issue but I still don't really understand why this is needed.

DavidoTek:
There was a similar issue upstream where defective UTF-8 characters would cause the problem. The fix didn't help with vdf.load though. I don't think there is an upstream issue for that yet.

> I remember a similar issue was related to incorrectly configured system locales. In the past I have also fallen victim to application issues caused by locale problems.
> I saw the discussion in the linked issue but I still don't really understand why this is needed.

It's most likely caused by a misconfigured system and not directly the vdf dependency. The problem is that the vdf file contains invalid UTF-8 characters that cannot be parsed for some reason. The vdf parser will throw an error and crash ProtonUp-Qt. This patch will replace the invalid characters with a valid character (I think �).

There are multiple people who reported the issue, so it seems like a good idea to ensure that it won't crash ProtonUp-Qt.

sonic2kk:
Okay, I think I understand. So this isn't strictly a problem caused by us or the VDF library, but is probably some system issue. That's okay then, I just wanted to make sure there wasn't a deeper problem.

Is this something we should report upstream?

DavidoTek:
> Is this something we should report upstream? Other applications using the VDF library are probably going to bump into this.

I don't think upstream can do much about it because the unicode-error handling takes place in the `open` function which is called before the vdf library. `vdf.load` is passed the file object.
I'm thinking about opening an issue nevertheless in case others encounter a similar issue.

sonic2kk:
Ohh you're right, I was thinking that this was called when the VDF library was trying to open it, but I guess this is actually to do with how Python tries to `open` the file, which it can't do on systems with locale issues.

A number of years ago I encountered locale issues with Protontricks, and some years ago I also encountered locale issues with SteamTinkerLaunch. I think the solution in this PR makes sense, and opening an issue upstream could lead to some discussion around the cause of the problem and some solutions.

sonic2kk:
I saw the upstream discussion and agree that this is a result of the `open()` function, but I really wonder why it seems to have only recently become an issue. Maybe Valve changed something, or suddenly more people have incorrect locales? I'm trying to understand if something broke here, but I don't know what could have broke.

How likely is it that Valve changed something that has suddenly made systems with incorrect locales run into this issue, I wonder.

<hr>

I think this PR covers all of the VDF loads where we would need this from what I can tell. Will take a look at #407 closer to the time to see how it would integrate with this. :-)

Issue #454, opened 2024-09-11
IndexError: List index out of range


vadimen:
Please fill out following when reporting a new bug:
![image](https://github.com/user-attachments/assets/18ee8b61-4606-4c54-af5d-cf051e0777a7)

**Describe the bug**  
happens when I try to set any folder as the location for installation

**To Reproduce**  
try to set a location

**Desktop (please complete the following information):**  
 - **Platform**: PC/Laptop?
 - **System**: Ubuntu 23.10
 - **Version**: ProtonUp-Qt 2.10.0
 - **How did you install ProtonUp-Qt?**: 
apt-get update
apt-get install wget
wget https://github.com/DavidoTek/ProtonUp-Qt/releases/download/v2.10.2/ProtonUp-Qt-2.10.2-x86_64.AppImage
chmod +x ProtonUp-Qt-2.10.2-x86_64.AppImage
./ProtonUp-Qt-2.10.2-x86_64.AppImage --appimage-extract
cd squashfs-root

apt-get install libegl1 libxcb-xinerama0 libxcb-cursor0 libxkbcommon-x11-0 libxcb-xinerama0 libxcb-cursor0 libx11-xcb1 libxrender1 libxi6 libxrandr2 libxkbcommon-x11-0 libxcb-xinerama0 libxcb-cursor0 libx11-xcb1 libxrender1 libxi6 libxrandr2 libxkbcommon-x11-0 libxcb-icccm4 libxcb-keysyms1 libxcb-shape0 python3

### for ubuntu 23.10 also
apt-get install libglib2.0-0 libgl1 libfontconfig1 libdbus-1-3

./AppRun

**Additional context**  

**Terminal output**  
```
ProtonUp-Qt 2.10.0 by DavidoTek. Build Info: Official AppImage by DavidoTek.
Python 3.10.4 (main, Apr  2 2022, 09:04:19) [GCC 11.2.0], PySide 6.7.2
Platform: Ubuntu 23.10 Linux-5.15.0-119-generic-x86_64-with-glibc2.38
Loading locale C / C
Loaded ctmod GE-Proton
Loaded ctmod Wine-GE
Loaded ctmod Boxtron
Loaded ctmod Kron4ek Wine-Builds Vanilla
Loaded ctmod Lutris-Wine
Loaded ctmod Luxtorpeda
Loaded ctmod Northstar Proton (Titanfall 2)
Loaded ctmod Proton Tkg
[...]

PR #457, opened 2024-09-15
CID: Disable `btnDefault` if no Custom Install Directory path is set

sonic2kk:
Based on suggestion in https://github.com/DavidoTek/ProtonUp-Qt/issues/454#issuecomment-2351673220.

This PR disables the "Default" button (`btnDefault`) if no Custom Install Directory has been configured. It also disables the button on click, if it successfully clears the value from the INI (i.e. if after clicking, the Custom Install Directory is successfully removed, toggle `btnDefault` enabled/disabled based on the `bool` cast of the value in the INI file).

The Custom Install Directory path is saved to the ProtonUp-Qt config file, and we can retrieve it using `config_custom_install_location`. I believe this is also responsible for writing out the path as well to the INI file. We can toggle whether `btnDefault` is enable on dialog load based on if `custom_install_directory` is defined.

`custom_install_directory: str = config_custom_install_location().get('install_dir', '')` will retrieve the value of `install_dir` based on the value in the INI. If this is not set, trying to fetch `install_dir` will actually return `None`. I am not _entirely_ sure why this is, I would've thought calling `.get('install_dir', '')` would return `''` for `False`y values, but I guess not. The `config_custom_install_location` function can take an `install_dir` keyword argument and this defaults to `install_dir=None`, so that's where the `None` comes from. If no Custom Install Directory is set, `config_custom_install_location` returns `None`.

Instead of refactoring behaviour in `config_custom_install_location` for this PR and potentially breaking things (i.e. changing the default argument to `install_dir=''`), I rolled with it and called `setEnabled` with a `bool` cast of `custom_install_directory`. If `custom_install_directory` returns `None` (which it should if there is no Custom Install Directory is set) then we call `bool(None)` to toggle the enabled state of `btnDefault`, which will disable it (`bool(None)` is `False`). Also, if `custom_install_directory` contained other `False`y values like an empty string (`''`) (which shouldn't be possible normally, but could happen if a user manually edited their INI file), we would also disable `btnDefault`. I think this is desirable; `False`y values should be considered equal to having no Custom Install Directory set, and so should disable `btnDefault`

Since this functionality is also used on `btnDefault` click, I broke it out into a method that handles checking if a `custom_install_directory` is set to a `Truth`y value. This makes the code a bit cleaner, as instead of doing the `bool` cast ourselves, we just tell `btnDefault.setEnabled` to toggle based on `self.has_custom_install_directory` return value, and that method handles the implementation detail. This also means if we wanted to change this implementation in future it would be more straightforward (e.g. if we ever wanted to validate the path in any way).

I have attached some screenshots to show this change.

### Before (`main`)
image

### After (This PR)
image

It is also enabled again when a Custom Install Directory is saved and the dialog is re-opened.

image

<hr>

Thanks!

### DIFF
--- a/pupgui2/pupgui2customiddialog.py
+++ b/pupgui2/pupgui2customiddialog.py
@@ -45,7 +45,10 @@ class PupguiCustomInstallDirectoryDialog(QObject):
         self.txtIdBrowseAction = self.ui.txtInstallDirectory.addAction(QIcon.fromTheme('document-open'), QLineEdit.TrailingPosition)
         self.txtIdBrowseAction.triggered.connect(self.txt_id_browse_action_triggered)
 
-        self.ui.txtInstallDirectory.setText(config_custom_install_location().get('install_dir', ''))
+        custom_install_directory = config_custom_install_location().get('install_dir', '')
+
+        self.ui.txtInstallDirectory.setText(custom_install_directory)
+        self.ui.btnDefault.setEnabled(bool(custom_install_directory))  # Don't enable btnDefault if there is no Custom Install Directory set
 
         self.ui.comboLauncher.addItems([
             display_name for display_name in self.install_locations_dict.values()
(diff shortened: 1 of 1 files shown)


DavidoTek:
Thanks!

Issue #412, opened 2024-06-22
Add support for RTSP GE Proton


CorneliusCornbread:
**Is your feature request related to a problem? Please describe.**
For VRChat players we have to install a fork of GE proton which patches RTSP players. These patches include many fixes, especially to encoders found in the windows media foundations libraries. Installing these manually can be a bit of a pain

**Describe the solution you'd like**
I propose two possible solutions:
- Allow for installing from an archive: this would make it so that any forked, custom version of proton can be installed easily to Steam
- Simply support downloading RTSP proton in the same manner that regular GE proton is supported.

sonic2kk:
> Allow for installing from an archive: this would make it so that any forked, custom version of proton can be installed easily to Steam

I like this idea a lot, actually. If there was a way we could do this "ModOrganizer 2 style" where after selection we can show the contents of the archive and allow the user to select what folder should be extracted (defaulting to the top-level one) I think this would be great.

You could open a separate enhancement request for that feature, if you'd like.

> Simply support downloading RTSP proton in the same manner that regular GE proton is supported.

This is probably equally valid for this use case, but doesn't invalidate the above proposal. Implementing this as a Ctmod would be straightforward I think, it could likely just inherit from the GE-Proton CtMod.

Maybe this one should be under Advanced Mode?

sonic2kk:
Yeah, this was very straightforward to add. I made a branch here (forgot to put it under Advanced Mode): https://github.com/sonic2kk/ProtonUp-Qt/tree/rtsp-ge-proton

I verified that it showed up, downloaded, and extracted correctly. The extracted contents have the same structure as GE-Proton, it uses tagging properly, uses the same archive format, so it was straightforward.

As a disclaimer: I am not sure if we want to go this direction though, so I won't open a PR yet. That branch is **only** there to serve as a proof-of-concept that it is possible, not a confirmation or endorsement :-)

CorneliusCornbread:
> As a disclaimer: I am not sure if we want to go this direction though, so I won't open a PR yet. That branch is only there to serve as a proof-of-concept that it is possible, not a confirmation or endorsement :-)

[...]

PR #470, opened 2024-10-23
Add RTSP Proton CtMod

DavidoTek:
Fix https://github.com/DavidoTek/ProtonUp-Qt/issues/412

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_rtspgeproton.py
+++ b/pupgui2/resources/ctmods/ctmod_rtspgeproton.py
@@ -1,0 +1,23 @@
+# pupgui2 compatibility tools module
+# Roberta
+# Copyright (C) 2021 DavidoTek, partially based on AUNaseef's protonup
+
+from PySide6.QtCore import QCoreApplication
+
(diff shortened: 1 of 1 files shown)


DavidoTek:
Should this be hidden behind `advmode`?

sonic2kk:
I guess in this case, whether or not it goes behind Advanced Mode is more a question of "clutter" in the dropdown.

In terms of codebase "precedent", we put NorthStar Proton behind Advanced Mode, so I suppose we can do it here, too. Makes sense to me :-)

sonic2kk:
Went ahead and pushed that change and rebased as well. Tested and it seems to work. Didn't launch any games with it but ProtonUp-Qt was able to download and extract the archive, and it looks correct and matches a standard Proton compatibility tool file structure.

image

<hr>

As an aside, I've mentioned a GE-Proton ctmod refactor that I've been working on in the past. This change wouldn't get in the way of that, we would be able to continue using this ctmod in the same way (since the goal would be having a GE-Proton ctmod as a base class, exactly as we do here). Just wanted to call this out :-)

Issue #472, opened 2024-11-24
proton-cachyos


Delor3an91:
**Is your feature request related to a problem? Please describe.**

No possibility to add a GitHub Proton source.

**Describe the solution you'd like**

1) Add the proton-cachyos GitHub as another source to download and install their Proton versions.
2) Add the possibility to manually add a GitHub Proton repository.

Thank You for your amazing work!

Wish You well

DavidoTek:
Sorry for the late reply.

As you pointed out, the version of Proton does not provide any binary releases. There are two possible approaches:

1. Implement a compatibility tool module (ctmod) that downloads the sources and builds them on the users machine
2. Request that proton-cachyos (and other tools too for that matter) provide pre-build binaries

I think it makes the most sense to build each tool only once and go with approach (2) in the case of proton-cachyos. Furthermore, the compilation takes a while and that would not be ideal for the user experience.

I understand that there are compatibility tools that are not used by many and that it can be interesting to build individual forks/branches/tags. I don't have spare time to build such a ctmod at the moment but I'm open to accepting a PR if someone builds it. It could work by supplying a local folder or a URL+commit in a custom popup that opens once the Install button in the install dialog is pressed.

loathingKernel:
The reason `proton-cachyos` exists is because we wanted a version that didn't depend on the Steam runtime, but is running on the host system outside of pressure-vessel. As such it contains modifications that wouldn't allow it to be built as a redistributable until now. Recently we decided to start providing Steam Linux Runtime based releases which can be found [here](https://github.com/CachyOS/proton-cachyos/releases). At this point it is not decided yet that we want to keep doing this. If this is going to be come a regular release, I will let you know.

DavidoTek:
> we wanted a version that didn't depend on the Steam runtime, but is running on the host system outside of pressure-vessel

What is the reason for this? I assume you're doing that so it uses the optimized libraries of CachyOS?

> Recently we decided to start providing Steam Linux Runtime based releases which can be found [here](https://github.com/CachyOS/proton-cachyos/releases). At this point it is not decided yet that we want to keep doing this. If this is going to become a regular release, I will let you know.

Okay. That would change a lot as it would allow us to use the existing mechanisms of ProtonUp-Qt.

---

> 1. Implement a compatibility tool module (ctmod) that downloads the sources and builds them on the users machine

[...]

PR #486, opened 2025-02-23
Add ctmod for Proton-CachyOS

loathingKernel:
Based on the ctmod for ProtonGE, it add support for Proton-CachyOS as discussed in #472.

The different architectures are currently shown as different releases in the selection list. `x86_64_v3` downloads will only be shown if the host CPU supports the required instruction sets.

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_protoncachyos.py
+++ b/pupgui2/resources/ctmods/ctmod_protoncachyos.py
@@ -3,14 +3,14 @@
 # Copyright (C) 2021 DavidoTek, partially based on AUNaseef's protonup
 
 import os
-import requests
-import hashlib
+from typing import Dict, Optional, Set
 
-from PySide6.QtCore import QObject, QCoreApplication, Signal, Property
+from PySide6.QtCore import QCoreApplication, Signal
+from PySide6.QtWidgets import QMessageBox
 
+from pupgui2.networkutil import download_file
 from pupgui2.util import ghapi_rlcheck, extract_tar
-from pupgui2.util import build_headers_with_authorization
-
+from .ctmod_00protonge import CtInstaller as ProtonGECtInstaller
 
 CT_NAME = 'Proton-CachyOS'
 CT_LAUNCHERS = ['steam', 'heroicproton', 'bottles']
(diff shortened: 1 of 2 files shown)


sonic2kk:
I wonder if there's a way we could subclass the GE-Proton ctmod for this, like we do for [RTSP-Proton](https://github.com/DavidoTek/ProtonUp-Qt/blob/64e25fa3aa3ab8aa678e8b6d1708439075042977/pupgui2/resources/ctmods/ctmod_rtspgeproton.py).

It might not be applicable with how releases are fetched and the extra `get_hwcaps` method, but just something I observed :-)

pupgui2/resources/ctmods/ctmod_protoncachyos.py
```diff
@@ -0,0 +1,225 @@
...
+
+    download_canceled = Property(bool, get_download_canceled, set_download_canceled)
+
+    def __set_download_progress_percent(self, value : int):
+        if self.p_download_progress_percent == value:
+            return
+        self.p_download_progress_percent = value
+        self.download_progress_percent.emit(value)
+
+    def __download(self, url, destination):
```
sonic2kk, line 63: I left a comment around if this could be replaced with a subclass of GE-Proton, so this only applies if that is not applicable :-)

<hr>

I wonder if we can replace this with a call to `download_file` from `networkutil`. An example usage: https://github.com/DavidoTek/ProtonUp-Qt/blob/64e25fa3aa3ab8aa678e8b6d1708439075042977/pupgui2/resources/ctmods/ctmod_luxtorpeda.py#L65-L82

We don't use this in many places, I should really get around to looking into why and replacing those usages.

loathingKernel:
> I wonder if there's a way we could subclass the GE-Proton ctmod for this

Even with the custom methods in this ctmod, more than half of the class's methods wouldn't need to be duplicated. So I will be looking into it. As for using the `download_file` method, even with subclassing it should be inherited here unless there are other reasons for not using it in ProtonGE.

sonic2kk:
I vaguely remember it not being used because I wanted to do a refactor of that class overall :-) It was a bit of a task, which is probably why I never got around to using it here. But if we do go the route of subclassing here, it would come for free once we eventually move that ctmod over

DavidoTek:
Thanks, that was quick :smile:

Looks good.

> `ctmods: subclass ProtonGE ctmod for Proton-CachyOS`

I just wanted to comment on that. Great, that seems to be resolved now too. I will do a quick review of the changes.

In the future, we could also take a look at moving `get_hwcaps` in some util file in case any other tool needs that too. But I think we should leave it in the cachyos ctmod for now, that will be up to the PR for the "other tool" then.

PR #487, opened 2025-02-23
Add a 3px padding to the main list and combobox items

loathingKernel:
Adds a padding of 3px to the installed compat tool list in the main window and the installation dialog to help with mouse navigation.

### DIFF
--- a/pupgui2/pupgui2.py
+++ b/pupgui2/pupgui2.py
@@ -142,6 +142,7 @@ class MainWindow(QObject):
         self.ui.btnClose.clicked.connect(self.btn_close_clicked)
         self.ui.listInstalledVersions.itemDoubleClicked.connect(self.list_installed_versions_item_double_clicked)
         self.ui.listInstalledVersions.itemSelectionChanged.connect(self.list_installed_versions_item_selection_changed)
+        self.ui.listInstalledVersions.setStyleSheet('QListWidget::item { padding: 3px; }')
         self.ui.btnShowCtInfo.clicked.connect(self.btn_show_ct_info_clicked)
         self.ui.btnSteamFlatpakCtools.clicked.connect(self.btn_steam_flatpak_ctools_clicked)
 
--- a/pupgui2/pupgui2installdialog.py
+++ b/pupgui2/pupgui2installdialog.py
@@ -1,14 +1,13 @@
 import os
-import threading
 import pkgutil
+import threading
 
 from PySide6.QtCore import Signal, QLocale, QDataStream, QByteArray
-from PySide6.QtGui import QIcon, QPixmap
-from PySide6.QtWidgets import QDialog
+from PySide6.QtGui import QIcon, QPixmap, Qt
 from PySide6.QtUiTools import QUiLoader
+from PySide6.QtWidgets import QDialog
 
 from pupgui2.util import open_webbrowser_thread, config_advanced_mode, get_combobox_index_by_value
-
 
 RELEASES_PER_PAGE = 50  # Number of releases to fetch per page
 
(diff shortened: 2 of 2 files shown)


sonic2kk:
I wonder if this is generally useful for other dropdowns, and if it's possible to do in the `.ui` files. I know we already had that call to `setStyleSheet` here, was just a thought I had when  taking a peek at this PR :-)

loathingKernel:
My experience with Qt stylesheets is that once you start going too global, the stylesheet's explicitness needs to increase. Missing values from the stylesheet will start overriding Fusion to use Qt defaults, making for a very weird and ugly UI in some cases. I have worked around this in two ways, the very targeted approach in a global stylesheet where I am theming by class and object name which I used for things that needed to be themed at all times, and in the case of an optional global theme, the theme styled almost everything with explicitly set values.

> and if it's possible to do in the .ui files
In my case, I am trying to keep `.ui` files as minimal as possible, and while Qt Designer offers the option to set a stylesheet for some widgets, I always chose to do it on the python side. This is also in part because I tend to use transpiled to python UI files which are instantiated as class attributes instead of dynamically loading the ui file itself.

I think a global stylesheet only for a minimum amount elements could work, for other specialized usages I would suggest either stylesheets specific to the widget or depending on the extent, some other overly complicated approach like the one I described above. If it looks interesting I could dive into the details of how it works.

DavidoTek:
Thanks for the contribution.

I compared this to main and noticed that there is already a small margin with `comboCompatTool` that is missing for `comboCompatToolVersion`. I assume that it is caused by `QComboBox { combobox-popup: 0; }` (Ubuntu GNOME 46, native pupgui2)

Maybe it would be better to let the system theme decide on the padding and somehow fix the combobox popup thing that removes the padding. However, I agree that `listInstalledVersions` looks better with the added padding...

> My experience with Qt stylesheets is that once you start going too global, the stylesheet's explicitness needs to increase

There is currently an (old) draft PR that adds a global stylesheet with a Steam Deck-like style for ProtonUp-Qt. It also overrides the style for `QListWidget::item`. Which one would be prioritized in that case?

https://github.com/DavidoTek/ProtonUp-Qt/pull/316/files#diff-f9099ee1ce4c6749b7f2b19567ac9ca09e4c21a9961a79d4251455a0e332f45f

loathingKernel:
>  I assume that it is caused by QComboBox { combobox-popup: 0; } 

Indeed, this is likely the reason, I do not remember the specifics though to be certain. I assumed that the popup style is a choice, hence the need for the padding. The padding in `comboCompatTool` was maybe overzealous of me to be consistent.

> Maybe it would be better to let the system theme decide on the padding and somehow fix the combobox popup thing that removes the padding.

I would like that too, but this is what I meant about explicitness. If we set the `combo-popup` property, in qss, it overrides other things too. Although I might have some other idea.

> There is currently an (old) draft PR that adds a global stylesheet with a Steam Deck-like style for ProtonUp-Qt. It also overrides the style for `QListWidget::item`. Which one would be prioritized in that case?
I think the way it goes is that the global stylesheet is set on every widget. Once the widget has a local stylesheet, the widget loses the global stylesheet (or maybe the local is appended to the global, I am not 100% sure). If the stylesheet is specialized by object name, then the specialized stylesheet takes priority.

My opinion is that you should try to not mix palette and global qss as much as possible, from a maintenance perspective. It can get messy. A palette with styling certain widgets is acceptable though. I did this by styling certain objects only. In the [static_css](https://github.com/RareDevs/Rare/tree/main/rare/resources/static_css) for Rare you can see my approach. On the other hand, compare it with Rare's full-blown [stylesheet](https://github.com/RareDevs/Rare/tree/main/rare/resources/stylesheets/RareStyle) you can see what I mean about having to go all the way. In my approach the static_css is loaded after the stylesheet if it is enabled, and that way it works fine.

In either case, I would suggest avoiding hand-written qss and using `qstylizer` to generate the style, it will allow you SASS levels of freedom with some introspection features, like grabbing the object names from existing classes almost automatically.

I am open to trying to figure out something more global, but I would need to look deeper into the code to see the extent of the modifications.

PR #489, opened 2025-03-02
networkutil: Round up chunk_count to ensure it cannot be

sonic2kk:
This PR fixes a bug in `networkutil#download_file` where our calculation for `current_chunk` could return `0`. Instead of `int`, we use `math.ceil` to make sure fractional results are rounded up.

An alternative solution could be to append `or 1` so that if the value is `0`, it is treated as Falsey and we default to `1`, but I think using `math.ceil` is a bit more elegant even if it means the extra import :-) But happy if we want to avoid the import to just use `int(...) or 1`.

I have tested with Luxtorpeda (Steam) and DXVK (Lutris), and this should not cause any regressions.

<hr>

This bug was discovered while I was working on porting `download_file` to Steam-Play-None, where the `current_chunk` calculation was returning `0` and causing a Division by Zero error. I have also confirmed that this PR fixes the issue.

All feedback is welcome. Thanks! :-)

### DIFF
--- a/pupgui2/networkutil.py
+++ b/pupgui2/networkutil.py
@@ -1,6 +1,8 @@
 import os
-from PySide6.QtCore import Property
 import requests
+import math
+
+from PySide6.QtCore import Property
 
 from typing import Callable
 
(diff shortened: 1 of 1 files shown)


DavidoTek:
Thanks.

> I think using math.ceil is a bit more elegant even if it means the extra import :-) 

That seems like the best solution. With `int()` it was possible in theory that the download progress was `> 100%`. This should also be fixed now (not sure if that ever happened, but it was possible).
The extra import is fine, it's not adding a new dependency.

> where our calculation for current_chunk could return 0

There is still one case where it can return 0 now: When `file_size=0`, `math.ceil(file_size/buffer_size) == 0`. This shouldn't happen because the file size will never be 0 unless something else breaks.

PR #491, opened 2025-03-02
GE-Proton: Use release_format

sonic2kk:
This PR simplifies the NorthStar Proton Ctmod by subclassing the GE-Proton ctmod, removing a lot of overlapping code.

As part of some cleanup to achieve this, the GE-Proton Ctmod was refactored slightly to hold a `release_format`. NorthStar Proton and GE-Proton both use `.tar.gz` formats, but in future I hope to allow more Ctmods to inherit from GE-Proton. Since I was here I figured it would be good housekeeping to implement it now that we have a couple of Ctmods taking from GE-Proton :-)

I tested both GE-Proton and NorthStar Proton and they should still be downloading correctly.

image

<hr>

Let me know if there are any concerns with this, or if I have missed something. Thanks!

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_00protonge.py
+++ b/pupgui2/resources/ctmods/ctmod_00protonge.py
@@ -30,6 +30,8 @@ class CtInstaller(QObject):
         super(CtInstaller, self).__init__()
         self.p_download_canceled = False
 
+        self.release_format = 'tar.gz'
+
         self.rs = requests.Session()
         rs_headers = build_headers_with_authorization({}, main_window.web_access_tokens, 'github')
         self.rs.headers.update(rs_headers)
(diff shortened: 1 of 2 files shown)


sonic2kk:
I wonder if instead of specifying `release_format`, would it be better to have an `self.__extract_tool` method that child classes would be expected to override.

pupgui2/resources/ctmods/ctmod_00protonge.py
```diff
@@ -159,7 +161,7 @@ def get_tool(self, version, install_dir, temp_dir):
         if source_checksum and (download_checksum not in source_checksum):
             return False
 
-        if not extract_tar(proton_tar, install_dir, mode='gz'):
+        if not extract_tar(proton_tar, install_dir, mode=self.release_format.split('.')[-1]):
```
DavidoTek, line 164: This may need some modifications in case we get more compatibility tools that use `.zip` like Proton-Tkg or some other non-tar archive.
We can change that once we inherit a tool that needs it.

DavidoTek:
> I wonder if instead of specifying release_format, would it be better to have an self.__extract_tool method that child classes would be expected to override.

I think the `release_format` variable is fine. The variable makes it obvious what format the tool uses. Adding a reference to some function wouldn't be that clean.

PR #494, opened 2025-03-03
Kron4ek: Simplify by Inheriting GE-Proton

sonic2kk:
This PR simplifies the Kron4ek Ctmod by inheriting from the GE-Proton Ctmod. A lot of the logic between these Ctmods is shared, similar to Proton-CachyOS (#486) and NorthStar Proton (#491), so this PR reduces the duplicated code. This was not possible before #491, where we introduced the `release_format` variable into GE-Proton.

Note that like in #491, we override `fetch_releases` and `__fetch_github_data`. I think this is a clean enough approach as it makes sense that these ctmods would have their own conditions for fetching releases, but let me know if this just makes things messy.

I still envision eventually that we can move to some kind of "base" ctmod, but I think incrementally moving existing Ctmods over to inheriting where it makes sense and cuts code duplication already is a good approach. But please let me know if it's just noisy :-) I am mainly interested here in simplifying our existing ctmods where possible.

Thanks!

<hr>

P.S. - The most obvious candidate to do this with may seem like the Wine-GE Ctmod, and I haven't forgotten about that Ctmod ;-) It just needs a lil bit of extra work to... work. But I'm... working on it :smile:

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_00protonge.py
+++ b/pupgui2/resources/ctmods/ctmod_00protonge.py
@@ -113,7 +113,7 @@ class CtInstaller(QObject):
                 values['size'] = asset['size']
         return values
 
-    def is_system_compatible(self):
+    def is_system_compatible(self) -> bool:
         """
         Are the system requirements met?
         Return Type: bool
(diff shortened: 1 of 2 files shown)


DavidoTek:
Thanks!

> I still envision eventually that we can move to some kind of "base" ctmod, but I think incrementally moving existing Ctmods over to inheriting where it makes sense and cuts code duplication already is a good approach. But please let me know if it's just noisy :-) I am mainly interested here in simplifying our existing ctmods where possible.

I thought about the same thing. At some point, it would be nice to have a `BaseCtMod` that cannot be installed itself but which other ctmods inherit from.

PR #496, opened 2025-03-04
GE-Proton: Allow downloading for Lutris. TODO add note to release notes

### LINKED ISSUE #191: If is Steam, ProtonGE7-49 is available but if is Lutris, ProtonGE7-37 and older is available

sonic2kk:
Fixes #191.

This PR enables downloading GE-Proton for Lutris. No other changes were needed to facilitate this, just adding `lutris` to the launchers list in the Ctmod. ProtonUp-Qt is able to list, download, and extract GE-Proton for Lutris.

Note that GE-Proton appears at the top of the list and is the first selected tool for Lutris. This is probably fine, given that [GE-Proton + UMU is intended to supersede Wine-GE](https://github.com/GloriousEggroll/wine-ge-custom/releases/tag/GE-Proton8-26). Wine-GE has [even been archived as of January this year](https://github.com/GloriousEggroll/wine-ge-custom). I suppose that raises a question of if we want to remove Wine-GE entirely from ProtonUp-Qt and replace it with GE-Proton?

If we want to replace WIne-GE with GE-Proton, I can remove that ctmod in the scope of this PR :-) And if we do this, we should note on the changelog that Wine-GE has been removed and that users should use GE-Proton instead (something something "Replaced Wine-GE with GE-Proton for Lutris (already available for other launchers) - Upstream project has been archived")

Also, if we don't want GE-Proton to be the default for Lutris, do we ant to add a shortcut for Lutris to open the Ctmod dialog with GE-Proton?

Thanks!


sonic2kk:
I did not have a strong opinion when opening this PR but having let it sit in my head for a bit, I think removing Wine-GE is probably a good idea.

DavidoTek:
Thanks!

> If we want to replace WIne-GE with GE-Proton, I can remove that ctmod in the scope of this PR :-)
> I did not have a strong opinion when opening this PR but having let it sit in my head for a bit, I think removing Wine-GE is probably a good idea. We did similar for D8VK a while back (#422).

I first thought about keeping Wine-GE and putting it behind advanced mode. On the other hand, I don't think there are many reasons to use "ancient" compatibility tools and people can always go back and download an old version of the ProtonUp-Qt AppImage. Moreover, the choice between GE-Proton and Wine-GE could confuse people.

I will quickly remove it and merge this.

>  And if we do this, we should note on the changelog that Wine-GE has been removed and that users should use GE-Proton instead (something something "Replaced Wine-GE with GE-Proton for Lutris (already available for other launchers) - Upstream project has been archived")

Yes, we should remember that. I'll rename this PR so it is in the commit history, that's what I use to write the changelog for the release notes.

PR #497, opened 2025-03-04
Lutris: Allow checking data directory for game configs

### LINKED ISSUE #495: Can't Open CtInfo or Games List Dialog for Lutris

**Describe the bug**  
When trying to open the CtInfo or Games List for Lutris, ProtonUp-Qt crashes because it cannot find the Lutris games folder.

ProtonUp-Qt is searching for the Lutris games folder in `~/.config/lutris/games` (iirc to parse the game's yaml file), however on my laptop, it appears that this folder is at `~/.local/share/lutris/games`.

While the quick-fix would be to return an empty dict in [`datastructures.py:187`](https://github.com/DavidoTek/ProtonUp-Qt/blob/71885f61dba55c5dce2c8d7c21aa3fd47946a014/pupgui2/datastructures.py#L187), I am opening an issue because there may be a broader problem here. Did the Lutris game folder path change, and if so, when? This is more of a wider question, and something I plan to look into if I have time and no one beats me to it, but something I wanted to report all the same.

If the path has changed, we should investigate if Lutris migrates existing game yml's over and if it that is the case, I don't know if we need to worry about "backwards compatibility" on our part; we can just expect users to upgrade their Lutris version and have launched it in order to migrate the yml's over.

If Lutris does not, we may need to look into checking both paths; the `~/.config/lutris/games` path and the `~/.local/share/lutris/games` path (and the Flatpak equivalents).

Either way, too, the check in `datastructures.py` is probably a bit naive to begin with. We assume that `os.path.join(os.path.expanduser(lutris_config_dir), 'games')` will exist when we pass it to `os.listdir`, which we probably _shouldn't_ do. We can address this as part of this issue when resolving it though; whether the path has changed or not we should probably store `os.path.join(os.path.expanduser(lutris_config_dir), 'games')` in a variable and return an empty dict if it doesn't exist.

**To Reproduce**  
1. Select the "Lutris" launcher from ProtonUp-Qt.
2. Attempt to open the CtInfo or Games List dialog.
3. ProtonUp-Qt will crash if `~/.config/lutris/games` does not exist.

**Expected behavior**  
ProtonUp-Qt should not crash, but more specifically we should investigate the exact cause here and see if we need any code changes beyond just checking if the path exists. We may need to tell ProtonUp-Qt to check both `~/.config/lutris/games` and `~/.local/share/lutris/games`, or we may be safe to simply move to whichever is the newest if Lutris did indeed specifically change this path.

And as part of this change, whether we need to change the path or check both, we should make our `os.listdir` parameter safer by checking that the folder(s) actually exist before we attempt to search them.

**Screenshots**  
Image

 - **Platform**: Laptop (did not verify on my Desktop PC yet, which has a 5+ year old Lutris installation)
 - **System**: Arch Linux
 - **Version**: ProtonUp-Qt @ 71885f61dba55c5dce2c8d7c21aa3fd47946a014
 - **How did you install ProtonUp-Qt?**: Running from source using Python 3.10 and a pyenv-virtualenv

**Additional context**  
Plan to check the paths and behaviour on my Desktop PC in the future.

**Terminal output**  
```
ProtonUp-Qt 2.11.1 by DavidoTek. Build Info: built from source.
Python 3.10.14 (main, Jun 21 2024, 21:57:01) [GCC 14.1.1 20240522], PySide 6.8.2.1
Platform: EndeavourOS rolling Linux-6.13.5-arch1-1-x86_64-with-glibc2.41
Loading locale en / en_GB
Loaded ctmod GE-Proton
Loaded ctmod Wine-GE
Loaded ctmod Boxtron
Loaded ctmod Lutris-Wine
Loaded ctmod Luxtorpeda
Loaded ctmod Proton-CachyOS
Loaded ctmod Proton Tkg
Loaded ctmod Roberta
Loaded ctmod RTSP Proton
Loaded ctmod Steam-Play-None
Loaded ctmod vkd3d-lutris
Loaded ctmod vkd3d-proton
Loaded ctmod Wine Tkg (Valve Wine Bleeding Edge)
Loaded ctmod Wine Tkg (Wine Master)
Loaded ctmod DXVK
Loaded ctmod DXVK Async
Loaded ctmod DXVK (nightly)
Attribute Qt::AA_ShareOpenGLContexts must be set before QCoreApplication is created.
  File "/home/emma/Programming/ProtonUp-Qt/pupgui2/pupgui2.py", line 387, in btn_show_game_list_clicked
    gl_dialog = PupguiGameListDialog(install_directory(), self.ui)
  File "/home/emma/Programming/ProtonUp-Qt/pupgui2/pupgui2gamelistdialog.py", line 41, in __init__
    self.setup_ui()
  File "/home/emma/Programming/ProtonUp-Qt/pupgui2/pupgui2gamelistdialog.py", line 53, in setup_ui
    self.setup_lutris_list_ui()
  File "/home/emma/Programming/ProtonUp-Qt/pupgui2/pupgui2gamelistdialog.py", line 94, in setup_lutris_list_ui
    self.update_game_list_lutris()
  File "/home/emma/Programming/ProtonUp-Qt/pupgui2/pupgui2gamelistdialog.py", line 200, in update_game_list_lutris
    game_cfg = game.get_game_config()
  File "/home/emma/Programming/ProtonUp-Qt/pupgui2/datastructures.py", line 187, in get_game_config
    for game_cfg_file in os.listdir(os.path.join(os.path.expanduser(lutris_config_dir), 'games')):
FileNotFoundError: [Errno 2] No such file or directory: '/home/emma/.config/lutris/games'
```

sonic2kk:
Fixes #495.

## Overview
This PR addresses two issues with how we currently fetch game config files for Lutris, which is used to display the CtInfo dialog and Games List dialog.

I tested this change on my laptop (uses the newer `~/.local/share/lutris/games` path), and this PR fixes the crash in #495 and finds the game config for the one game I have installed (GE-Proton9-25 for Battlenet displays properly in the Games List and CtInfo dialog). I have also tested this setup on my Desktop PC (uses the older `~/.config/lutris/games` path), which works correctly and displays the same games with and without this change.

## Problem
Currently, we check for games in the `~/.config/lutris/games` folder and that folder only. However, we assume that if we get a Lutris config folder back at all from `self.install_loc.get('config_dir')`, then we will have a `~/.config/lutris/games`. However this is _not_ the case. So if `~/.config/lutris/games` does not exist, there will be a crash when attempting to call `os.listdir` on that folder. This crash was the reason for opening #495.

However this was a symptom of another problem: Lutris no longer uses `~/.config/lutris/games` by default and instead will use `~/.local/share/lutris` for new installations. This is why I was experiencing the crash described, because my Lutris game config folder was at `~/.local/share/lutris` on my laptop but at `~/.config/lutris` on my Desktop PC.

The Lutris codebase has a comment noting that `~/.config/lutris` is deprecated: https://github.com/lutris/lutris/blob/6b968e858955c0638bf93b3a72fec5ae650f0932/lutris/settings.py#L20-L25

So the overall problem was that:
- We were making an assumption that `~/.config/lutris/games` would always exist, which is not the case.
- Lutris also no longer uses this folder, and instead we also need to check for `~/.local/share/lutris/games` which is the newer folder. However if `~/.config/lutris/games` exists, we should prefer that, and Lutris cannot use both `~/.config/lutris/games` and `~/.local/share/lutris/games` at the same time.

## Solution
We now search `~/.local/share/lutris/games` to find game configuration files (and Flatpak equivalent), if `~/.config/lutris/games` does not exist (and Flatpak equivalent). Lutris will prefer `~/.config/lutris/games` if it exists, but this path is deprecated and Lutris will default to using `~/.local/share/lutris/games` (XDG Data Directory) if the Config directory does not exist. 

As a result of this, we also fix the crash which was a symptom of this problem and the catalyst for opening #495. The crash comes from `os.listdir` when searching for our game config file, if `~/.config/lutris` exists but not `~/.config/lutris/games`, caused by the assumption that the games folder will exist in the config folder if the config folder exists. We no longer get as far as `os.listdir` unless we find a valid Lutris game config folder, either in the Config Directory or Data directory.

The overall implementation here will "fall up" to `~/.local/share/lutris/games` if `~/.config/lutris/games` is not found, different to before where we assumed the `games` folder existed if we got _any truthy value_ back from `install_loc.get('config_dir')`.

## Implementation
Most of the implementation is fairly straightforward "if no Lutris games folder in config dir, check the data dir for the games folder, and if neither exist then return an empty dict". One detail worth calling out though is how we get the data directory. We do this by going two directories up from the `install_loc.get('install_dir')`, which in the case of Lutris will be something like `~/.local/share/lutris/runners/wine/../..games/`. We get the `abspath` and `expanduser` of this so we end up with a clean path, like `/home/gaben/.local/share/lutris/games`.

This works, and _should_ be safe to do, and we do similar logic in Ctmods to get the install folder we want to use ([such as the vkd3d-proton Ctmod](https://github.com/DavidoTek/ProtonUp-Qt/blob/913c7a6a4c3ac7660adfd28db3ec9f1d16191c31/pupgui2/resources/ctmods/ctmod_vkd3dproton.py#L146-L147)). So there is precedent for doing this in the codebase, and it should be safe to do for Flatpak paths as well, since `install_dir` for Lutris should always be `lutris/runners/wine`. Two directories up should always give us the Lutris root install. It is worth noting that we make similar assumptions about the `install_dir` path and directory structure around that for Heroic in [`heroicutil`](https://github.com/DavidoTek/ProtonUp-Qt/blob/913c7a6a4c3ac7660adfd28db3ec9f1d16191c31/pupgui2/heroicutil.py#L22). But let me know if there are any concerns with this kind of approach!

For consistency, the whole `LutrisGame#get_game_config` method now works with `/path/to/lutris/games` instead of just getting the Lutris config folder and `os.path.join`ing on the `games` folder. This means the method a bit easier to read as well imo, since we only build the path to the game in one place.

<hr>

Let me know if I missed anything or if there are any concerns/edge cases I did not consider. All feedback is welcome.

Thanks! :-)

### DIFF
--- a/pupgui2/datastructures.py
+++ b/pupgui2/datastructures.py
@@ -177,24 +177,52 @@ class LutrisGame:
 
     install_loc = None
 
-    def get_game_config(self):
-        lutris_config_dir = self.install_loc.get('config_dir')
-        if not lutris_config_dir:
+    def get_game_config(self) -> dict:
+        
+        """
+        Get a Lutris game config .yml file from either the config directory or the data directory.
+        i.e., ~/.config/lutris/games or ~/.local/share/lutris/games
+
+        Return Type: dict
+        """
+
+        # Lutris will prefer the config directory if it exists, but will use the data directory if config does not exist.
+        # On newer Lutris installations, only the data directory is used, as the config directory is deprecated.
+        # However, Lutris does not migrate these installations, so we need to check for both and prefer config if it exists.
+        #
+        # https://github.com/lutris/lutris/blob/6b968e858955c0638bf93b3a72fec5ae650f0932/lutris/settings.py#L20-L25
+        lutris_game_config_dir = os.path.join(os.path.expanduser(self.install_loc.get('config_dir')), 'games')
+        if not lutris_game_config_dir:
             return {}
     
+        if not os.path.isdir(lutris_game_config_dir):
+            # Lutris 'install_dir' will be '/path/to/lutris/runners/wine', go two directories up to get the root Lutris install folder
+            lutris_game_config_data_dir = os.path.abspath(
+                os.path.join(
+                    os.path.expanduser(self.install_loc.get('install_dir')),
+                    '..', '..',
+                    'games'
+                )
+            )
+
+            if not os.path.isdir(lutris_game_config_data_dir):
+                return {}
+
+            lutris_game_config_dir = lutris_game_config_data_dir
+
         # search a *.yml game configuration file that contains either the install_slug+installed_at or, if not found, the game slug
         fn = ''
-        for game_cfg_file in os.listdir(os.path.join(os.path.expanduser(lutris_config_dir), 'games')):
+        for game_cfg_file in os.listdir(lutris_game_config_dir):
             if str(self.installer_slug) in game_cfg_file and str(self.installed_at) in game_cfg_file:
                 fn = game_cfg_file
                 break
         else:
-            for game_cfg_file in os.listdir(os.path.join(os.path.expanduser(lutris_config_dir), 'games')):
+            for game_cfg_file in os.listdir(lutris_game_config_dir):
                 if self.slug in game_cfg_file:
                     fn = game_cfg_file
                     break
 
-        lutris_game_cfg = os.path.join(os.path.expanduser(lutris_config_dir), 'games', fn)
+        lutris_game_cfg = os.path.join(lutris_game_config_dir, fn)
         if not os.path.isfile(lutris_game_cfg):
             return {}
         with open(lutris_game_cfg, 'r') as f:


DavidoTek:
Thanks!

> I wonder if we should have LutrisGame#get_game_config prefer ~/.config/lutris if it exists *(Issue 495)*

That makes sense since it is the same behavior as the Lutris code.

> ~/.local/share/lutris/runners/wine/../..games/. We get the abspath and expanduser of this so we end up with a clean path, like /home/gaben/.local/share/lutris/games.
> and it should be safe to do for Flatpak paths as well, since install_dir for Lutris should always be lutris/runners/wine

Seems to work fine with the Flatpak version as well.

PR #499, opened 2025-03-06
Proton-CachyOS: Remove deprecated Type Hinting

sonic2kk:
This PR removes the deprecated type hints from the `typing` module. They exist sprinkled throughout the codebase, and should be removed, so I don't mean to single out this ctmod. I spotted this as I'm eyeing some changes to help it further inherit from the GE-Proton ctmod in the future, so I was in and around here and noticed it. I have also removed a couple of unused imports from this Ctmod.

I also added the `@override` annotation to methods we override from GE-Proton, which we currently don't do elsewhere in the codebase, although it is something I also did in #498. Like with removing the old deprecated type hints, I hope to go through and add these around the codebase where relevant (I think only the Ctmods, maybe there's a Qt component we subclass for the Shortcut Editor but I'm not 100% sure).

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_protoncachyos.py
+++ b/pupgui2/resources/ctmods/ctmod_protoncachyos.py
@@ -3,7 +3,6 @@
 # Copyright (C) 2021 DavidoTek, partially based on AUNaseef's protonup
 
 import os
-from typing import override
 
 from PySide6.QtCore import QCoreApplication
 
@@ -33,7 +32,6 @@ class CtInstaller(ProtonGECtInstaller):
     CT_URL = 'https://api.github.com/repos/CachyOS/proton-cachyos/releases'
     CT_INFO_URL = 'https://github.com/CachyOS/proton-cachyos/releases/tag/'
 
-    @override
     def __fetch_github_data(self, tag: str, arch: str) -> dict | None:
         """
         Fetch GitHub release information
@@ -74,7 +72,6 @@ class CtInstaller(ProtonGECtInstaller):
             hwcaps.add('x86_64_v2')
         return hwcaps
 
-    @override
     def fetch_releases(self, count: int = 100, page: int = 1) -> list:
         """
         List available releases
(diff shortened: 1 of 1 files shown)


sonic2kk:
Actually, this annotation may only be available in Python 3.10. Looking at the PEP ([PEP 698](https://peps.python.org/pep-0698/)), I believe it's only in 3.12, but we currently build against Python 3.10 for the AppImage. Python 3.10 is still getting security updates [until October 31st 2026](https://endoflife.date/python), so it is not really worth bumping just to get this extra feature.

I should really have my ProtonUp-Qt virtualenv based on at Python 3.10, not 3.12... Would need to test this against 3.12 to make sure this works, although looking at the docs, [the Python 3.10 docs](https://docs.python.org/3.10/library/typing.html#typing.override) don't mention `@override`, but [the Python 3.12 docs do](https://docs.python.org/3.12/library/typing.html#typing.override).

Will need to confirm if this decorator is actually available to us.

loathingKernel:
If the AppImage is the only block to update the Python version, cx-freeze offers and appimage target that will include later versions of python with the image.

DavidoTek:
Thanks!

> Looking at the PEP (PEP 698), I believe it's only in 3.12, but we currently build against Python 3.10 for the AppImage

I just confirmed that it is not available to Python 3.10, the import fails.

The Pytest workflow succeeded. Maybe we should extend it beyond the test cases to verify the code is actually working for Python 3.10.
It is a bit tricky with the ctmods as they are dynamically imported. I assume we can add a test case that imports them.

> If the AppImage is the only block to update the Python version, cx-freeze offers an appimage target that will include later versions of python with the image. It should include whatever the build virtualenv contains really

I wasn't aware cx_freeze supports AppImage packaging. That is good to know.

sonic2kk:
I have removed the `@override` decorators for now. In future we can revisit potentially adding them :-) No reason to make this PR wait for changes to the AppImage side given it's the only thing we'd be using from 3.12 right now anyway.

PR #509, opened 2025-03-31
DXVK Nightly: Simplify by Inheriting from DXVK Ctmod

### LINKED ISSUE #500: DXVK Nightly Listing Test Windows Builds

**Describe the bug**  
DXVK's GitHub Actions has two workflows: The Linux builds ([`artifacts.yml`](https://github.com/doitsujin/dxvk/actions/workflows/artifacts.yml)) and the test workflow to build on Windows ([`test-build-windows.yml`](https://github.com/doitsujin/dxvk/actions/workflows/test-build-windows.yml)).

These runners build against the same commit, resulting in two problems:
- DXVK Nightly will list the same commit SHA twice, one for each time that commit has had an Action run against it.
- Only one of these versions is valid (should always be the "first" occurrence, but not guaranteed).

To fix this, we need to check which workflow the build was run against, so as to only pick the one for the Linux builds' `artifact.yml`. The way DXVK Nightly currently works is that it runs an API request against `artifacts`, meaning we get _all_ build artifacts from the DXVK GitHub Actions and so from all workflows. This is why we get the Windows builds included.

<hr>

Fortunately, we've already solved this problem around the codebase! The Proton-tkg Ctmod does exactly this, although it is considerably less elegant than what we currently have for DXVK Nightly:

https://github.com/DavidoTek/ProtonUp-Qt/blob/7264188f6475016fc1e348092334f11945e05a1a/pupgui2/resources/ctmods/ctmod_protontkg.py#L146-L164

Note that Proton-tkg fetches the _workflows_ and not _artifacts_, meaning its `workflow` object is different than what we have for DXVK Nightly. This means we cannot simply check `workflow['path']` in DXVK Nightly (I checked 😦).

Proton-tkg can fetch from multiple workflows, and child classes will override `PROTON_PACKAGE_NAME` to determine which workflow to download build artifacts from. It also has some extra logic to fetch historical artifacts if all builds on the current page failed.

So we can take this logic from Proton-tkg and apply it to DXVK Nightly: Telling it to check for the `artifacts.yml` workflow name, and only fetch artifacts that are a part of that workflow. This will solve the problem of DXVK Nightly builds listening seemingly duplicate versions, and so will only list valid versions.

We could break out this logic that we have for Proton-tkg into a util function that we can re-use between Proton-tkg and DXVK Nightly (and any future builds we download from GitHub Actions).

<hr>

I plan to take a look at this in the near future, but opened this issue so I don't lose track of the problem :-)

**To Reproduce**  
1. Select a Launcher that supports DXVK Nightly (i.e. Lutris).
2. Click on "Add Version".
3. Select "DXVK Nightly"
4. In the versions dropdown, the same commit SHA will be listed twice.
5. One of these (usually the 2nd occurrence) will be an invalid DXVK version as it is the Windows build and not the Linux build we expect. 

**Expected behavior**  
ProtonUp-Qt only lists valid and non-seemingly-duplicate versions of DXVK Nightly.

**Screenshots**  
Image

 - **Platform**: Laptop
 - **System**: Arch Linux
 - **Version**: 7264188f6475016fc1e348092334f11945e05a1a
 - **How did you install ProtonUp-Qt?**: Git via virtualenv

**Additional context**  
N/A.

**Terminal output**  
N/A.

sonic2kk:
Fixes #500.

## Overview
This PR refactors the DXVK Nightly Ctmod to inherit from the DXVK Ctmod, as there is quite a bit of shared logic between these Ctmods.

This PR also fixes #500 by refactoring the logic of how we fetch releases. Much of the complexity in this PR comes from fixing #500, not from the refactor itself. This complexity would have arisen regardless of the refactor.

#### CtInstaller Dialog Contains Invalid Versions

image

#### Expected Folder Structure

image

#### CtInstaller Dialog Contains Only Valid Versions

image

#### Expected Folder Structure Preserved

image

### Ctmod Refactoring
In order for DXVK Nightly to inherit from the main DXVK Ctmod, a slight rework was needed to the main DXVK Ctmod.
- We needed to create a `__get_data` method in order to allow the DXVK Nightly child class to override this method.
    - It needs to override the `data` that comes back, so that it returns the data from `__fetch_github_data` (which works with `__fetch_artifact_from_commit` to get the GitHub Actions workflow artifact information from the given commit the user selects in the version dropdown) instead of `__fetch_data`. This ultimately allows us to use the `nightly.link` link that gets set on `data` and then is downloaded via the DXVK main class's `get_tool` when it calls the overridden `__get_data`.
    - This is essentially the same idea as what we did in #498.
- For cleanliness, I broke out the extraction logic for the downloaded archive into an `__extract` method. This puts all of the conditional checks for if we're working with a ZIP or a tarball into this method.
    - However, since this is broken out as a separate method, we could let the DXVK Nightly ctmod handle overriding the extract logic as well. Instead of the main DXVK Ctmod handling this, Ctmods that inherit and need different functionality can override this method. Happy to discuss this, and will leave a comment about this after posting the PR :-)

I have tested the main DXVK Ctmod, DXVK Async (which also inherits from the main DXVK Ctmod), and of course DXVK Nightly, and the extracted folder (name and directory structure) works just how it does on `main` @ 44ec5917f6c5096795d9e6455e262048b9ce6935 right now.

### Fixing Workflow Fetching Logic
Onto how #500 was solved: Previously, we were fetching all releases from all workflows. That is, going to `https://api.github.com/repos/doitsujin/dxvk/actions/artifacts?per_page=50&page=1` and taking all of the Commit SHAs from the runs. This worked before, when DXVK only had one workflow, and so each commit only had one artifact. However now DXVK has two workflows: `artifacts.yml` for our Linux DXVK DLLs (and DXVK Native, which is an archive inside of these workflows) and `test-build-windows.yml` which builds DXVK with MSVC. This results in two workflows per commit, so we end up with DXVK versions being listed twice most of the time.

To fix this:
- **When fetching releases**, we need to instead only fetch runs that occurred in the `artifacts.yml` workflow. This means we will only populate the releases dropdown with valid commit hashes.
    - This unfortunately adds a little bit of overhead to the length of time it takes to fetch releases that is imo noticeable, but I don't think we can avoid it.
    - **I encountered some segfaults in initial development**, but have not been able to reproduce it in the extensive testing that has followed. 
- **When downloading**, we want to make sure the artifact we select does not end in `-msvc-output`. This means when we select a commit to download, instead of just picking the first one that matches the commit, we will pick the one that does not end in `-msvc-output` which should currently always correspond to the valid `artifacts.yml` workflow.
    - This is needed in addition to only fetching releases from a corresponding workflow to ensure we fetch the correct amount of artifacts. More on this in "[Checking For `-msvc-output` Suffix](#checking-for-msvc-output-suffix)".
    - In future, if other workflows were added or if this suffix changed, we would need to also ignore these. It may be sufficient to just check if our package ends with the expected commit hash and nothing else, we can discuss this though :-) I will leave a comment about this after posting this PR.

#### Using Commit Hashes
The solution of fetching runs from only a specific workflow comes from the solution we used for the Proton-tkg ctmod, as outlined in #500. In the DXVK Nightly Ctmod I have implemented a `__get_workflows` method [very similar to what we used for Proton-tkg](https://github.com/DavidoTek/ProtonUp-Qt/blob/7264188f6475016fc1e348092334f11945e05a1a/pupgui2/resources/ctmods/ctmod_protontkg.py#L146-L164), but the main difference is that I chose to return Commit SHAs instead of Workflow Run IDs. There are benefits to this:
- It preserves the existing behaviour for DXVK Nightly, which displays Commit SHAs.
- It preserves the functionality of `get_info_url` method which will allow the "info" button on the CtInstaller dialog to open at that commit.

#### Reusing `__fetch_workflows`
I would really like `__fetch_workflows` to be a generic function, inside of a `githubutils.py` file. The ultimate goal here would be to allow this to be re-used between DXVK Nightly, Proton-tkg, and anything else that might require it. In addition, it would allow us to create unit tests for this function that we could test against pre-defined JSON files of test data. We _could_ write unit tests eventually for our Ctmods (and hopefully will do this) that would cover this, but I would also like if this was a standalone function that could be tested.

I had actually refactored this method out into a standalone function, inside of a `githubutils.py` file, but I scrapped that because I realised that in its current state we cannot re-use it between DXVK Nightly and Proton-tkg because DXVK Nightly works with commit hashes but Proton-tkg works with Workflow IDs. If we want to make this function generic, we would need a way to define what it should return in its `tags` list (`tags` is not the best name, either, and is likely a holdover from working with release tags for tagged, non-nightly artifacts).

I decided not to tackle that in this PR, as I feel a separate PR could cover making the function generic and reusable between DXVK Nightly and Proton-tkg, and also that same PR could handle writing the unit tests.

#### Checking For `-msvc-output` Suffix
We check for the suffix to ensure we only download artifacts with commits that correspond to the artifact we want to use. Since we aren't working with Workflow IDs like with Proton-tkg, which are unique, we are getting all workflow runs with the commit. In this case there are usually two: One for `artifacts.yml`, and one for `test-windows-build.yml`. We can tell these builds apart based on a suffix that is appended to the MSVC builds, `-msvc-output`. This output comes from the [`test-windows-build.yml`](https://github.com/doitsujin/dxvk/blob/20a6fae8a7f60e7719724b229552eba1ae6c3427/.github/workflows/test-build-windows.yml#L80) GitHub Actions workflow file.

When I discovered this, I did consider that just checking the artifact name for `.endswith('-msvc-output')` would be enough, however this would result in returning fewer artifacts than we should. If a page returns 50 artifacts, and each commit has two artifacts (one for `artifacts.yml`, one for `test-windows-build.yml`), and we then filter to exclude the artifacts that end with `-msvc-output`, we end up with only 25 artifacts. This is why checking _all_ the artifacts in the `count` variable is important: We want to make sure we count all artifacts that we know are in a workflow that is valid, in this case checking all workflows in `artifacts.yml`.

Also, as a result of using `__fetch_workflows`, we are able to skip expired artifacts and move backwards through pages to make sure we collect as many artifacts as we can.

**Side Note**: It is conceivable (though not observed in my testing) that a DXVK `test-build-windows.yml` build would be the only present artifact for a commit (i.e. no Linux-compatible artifact at all), therefore even if only one version was listed, that listed version would always be invalid. With the current approach I have taken in this PR by only fetching commit SHAs for runs inside of the `artifacts.yml` workflow, and filtering out workflow runs that have `-msvc-output` in the name, we can guarantee that the versions we list are only successfully built artifacts from the `artifacts.yml` workflow, and we will never get Windows downloads (unless the `-msvc-output` suffix changes for MSVC builds).

<hr>

Some significant changes here with fixing DXVK Nightly to only download artifacts from the correct workflow. Let me know if any of the rationale here or justification was not explained properly, happy to elaborate however I can! And as usual, all feedback is welcome :-)

Thanks!

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_z0dxvk.py
+++ b/pupgui2/resources/ctmods/ctmod_z0dxvk.py
@@ -9,7 +9,7 @@ import requests
 from PySide6.QtCore import QObject, QCoreApplication, Signal, Property
 
 from pupgui2.networkutil import download_file
-from pupgui2.util import extract_tar, get_launcher_from_installdir, fetch_project_releases
+from pupgui2.util import extract_tar, extract_zip, get_launcher_from_installdir, fetch_project_releases
 from pupgui2.util import fetch_project_release_data, build_headers_with_authorization
 from pupgui2.datastructures import Launcher
 
@@ -117,6 +117,27 @@ class CtInstaller(QObject):
 
         return (data, dxvk_dir)
 
+    def __extract(self, archive_path: str, extract_dir: str) -> bool:
+
+        """
+        Extract the tool archive at the given path.
+        Return Type: bool
+        """
+
+        if not archive_path or not extract_dir:
+            return False
+        
+        # DXVK and DXVK Async are 'tar.gz'
+        if self.release_format.startswith('tar'):
+            tar_type = self.release_format.split('.')[-1]
+
+            return extract_tar(archive_path, extract_dir, mode=tar_type)
+        # DXVK Nightly is 'zip'
+        elif self.release_format == 'zip':
+            return extract_zip(archive_path, extract_dir)
+
+        # If unknown archive format, cannot extract, so default to fail
+        return False
 
     def get_tool(self, version, install_dir, temp_dir):
         """
@@ -129,11 +150,11 @@ class CtInstaller(QObject):
             return False
 
         # Should be updated to support Heroic, like ctmod_d8vk
-        dxvk_tar = os.path.join(temp_dir, data['download'].split('/')[-1])
-        if not self.__download(url=data['download'], destination=dxvk_tar, known_size=data.get('size', 0)):
+        dxvk_archive = os.path.join(temp_dir, data['download'].split('/')[-1])
+        if not self.__download(url=data['download'], destination=dxvk_archive, known_size=data.get('size', 0)):
             return False
 
-        if not dxvk_dir or not extract_tar(dxvk_tar, dxvk_dir, mode='gz'):
+        if not dxvk_dir or not self.__extract(dxvk_archive, dxvk_dir):
             return False
 
         self.__set_download_progress_percent(100)
--- a/pupgui2/resources/ctmods/ctmod_z0dxvk.py
+++ b/pupgui2/resources/ctmods/ctmod_z0dxvk.py
@@ -102,14 +102,30 @@ class CtInstaller(QObject):
         """
         return fetch_project_releases(self.CT_URL, self.rs, count=count, page=page)
 
+    def __get_data(self, version: str, install_dir: str) -> tuple[dict | None, str | None]:
+
+        """
+        Get needed download data and path to extract directory.
+        Return Type: diple[dict | None, str | None]
+        """
+
+        data = self.__fetch_data(version)
+        if not data or 'download' not in data:
+            return (None, None)
+
+        dxvk_dir = self.get_extract_dir(install_dir)
+
+        return (data, dxvk_dir)
+
+
     def get_tool(self, version, install_dir, temp_dir):
         """
         Download and install the compatibility tool
         Return Type: bool
         """
 
-        data = self.__fetch_data(version)
-        if not data or 'download' not in data:
+        data, dxvk_dir = self.__get_data(version, install_dir)
+        if not data:
             return False
 
         # Should be updated to support Heroic, like ctmod_d8vk
@@ -117,8 +133,7 @@ class CtInstaller(QObject):
         if not self.__download(url=data['download'], destination=dxvk_tar, known_size=data.get('size', 0)):
             return False
 
-        dxvk_dir = self.get_extract_dir(install_dir)
-        if not extract_tar(dxvk_tar, dxvk_dir, mode='gz'):
+        if not dxvk_dir or not extract_tar(dxvk_tar, dxvk_dir, mode='gz'):
             return False
 
         self.__set_download_progress_percent(100)
--- a/pupgui2/resources/ctmods/ctmod_z0dxvk.py
+++ b/pupgui2/resources/ctmods/ctmod_z0dxvk.py
@@ -21,20 +21,20 @@ CT_DESCRIPTION = {'en': QCoreApplication.instance().translate('ctmod_z0dxvk', ''
 
 class CtInstaller(QObject):
 
-    BUFFER_SIZE = 65536
-    CT_URL = 'https://api.github.com/repos/doitsujin/dxvk/releases'
-    CT_INFO_URL = 'https://github.com/doitsujin/dxvk/releases/tag/'
+    BUFFER_SIZE: int = 65536
+    CT_URL: str = 'https://api.github.com/repos/doitsujin/dxvk/releases'
+    CT_INFO_URL: str = 'https://github.com/doitsujin/dxvk/releases/tag/'
 
-    p_download_progress_percent = 0
-    download_progress_percent = Signal(int)
-    message_box_message = Signal((str, str, QMessageBox.Icon))
+    p_download_progress_percent: int = 0
+    download_progress_percent: Signal = Signal(int)
+    message_box_message: Signal = Signal((str, str, QMessageBox.Icon))
 
     def __init__(self, main_window = None):
         super(CtInstaller, self).__init__()
-        self.p_download_canceled = False
-        self.release_format = 'tar.gz'
+        self.p_download_canceled: bool = False
+        self.release_format: str = 'tar.gz'
 
-        self.rs = requests.Session()
+        self.rs: requests.Session = requests.Session()
         rs_headers = build_headers_with_authorization({}, main_window.web_access_tokens, 'github')
         self.rs.headers.update(rs_headers)
 
@@ -95,7 +95,7 @@ class CtInstaller(QObject):
         """
         return True
 
-    def fetch_releases(self, count=100, page=1):
+    def fetch_releases(self, count: int = 100, page: int = 1):
         """
         List available releases
         Return Type: list[str]
@@ -139,7 +139,7 @@ class CtInstaller(QObject):
         # If unknown archive format, cannot extract, so default to fail
         return False
 
-    def get_tool(self, version, install_dir, temp_dir):
+    def get_tool(self, version: str, install_dir: str, temp_dir: str) -> bool:
         """
         Download and install the compatibility tool
         Return Type: bool
@@ -150,7 +150,7 @@ class CtInstaller(QObject):
             return False
 
         # Should be updated to support Heroic, like ctmod_d8vk
-        dxvk_archive = os.path.join(temp_dir, data['download'].split('/')[-1])
+        dxvk_archive: str = os.path.join(temp_dir, data['download'].split('/')[-1])
         if not self.__download(url=data['download'], destination=dxvk_archive, known_size=data.get('size', 0)):
             return False
 
@@ -161,7 +161,7 @@ class CtInstaller(QObject):
 
         return True
 
-    def get_info_url(self, version):
+    def get_info_url(self, version: str) -> str:
         """
         Get link with info about version (eg. GitHub release page)
         Return Type: str
--- a/pupgui2/resources/ctmods/ctmod_z0dxvk.py
+++ b/pupgui2/resources/ctmods/ctmod_z0dxvk.py
@@ -9,7 +9,7 @@ import requests
 from PySide6.QtCore import QObject, QCoreApplication, Signal, Property
 
 from pupgui2.networkutil import download_file
-from pupgui2.util import extract_tar, extract_zip, get_launcher_from_installdir, fetch_project_releases
+from pupgui2.util import extract_tar, get_launcher_from_installdir, fetch_project_releases
 from pupgui2.util import fetch_project_release_data, build_headers_with_authorization
 from pupgui2.datastructures import Launcher
 
@@ -128,16 +128,9 @@ class CtInstaller(QObject):
             return False
         
         # DXVK and DXVK Async are 'tar.gz'
-        if self.release_format.startswith('tar'):
-            tar_type = self.release_format.split('.')[-1]
+        tar_type = self.release_format.split('.')[-1]
 
-            return extract_tar(archive_path, extract_dir, mode=tar_type)
-        # DXVK Nightly is 'zip'
-        elif self.release_format == 'zip':
-            return extract_zip(archive_path, extract_dir)
-
-        # If unknown archive format, cannot extract, so default to fail
-        return False
+        return extract_tar(archive_path, extract_dir, mode=tar_type)
 
     def get_tool(self, version: str, install_dir: str, temp_dir: str) -> bool:
         """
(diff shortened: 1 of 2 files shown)


pupgui2/resources/ctmods/ctmod_z0dxvk.py
```diff
@@ -95,37 +95,73 @@ def is_system_compatible(self):
...
+            return False
+        
+        # DXVK and DXVK Async are 'tar.gz'
+        if self.release_format.startswith('tar'):
+            tar_type = self.release_format.split('.')[-1]
+
+            return extract_tar(archive_path, extract_dir, mode=tar_type)
+        # DXVK Nightly is 'zip'
+        elif self.release_format == 'zip':
+            return extract_zip(archive_path, extract_dir)
```
sonic2kk, line 137: As noted in the PR, since this is a separate method, we could override this method inside of the DXVK Nightly Ctmod and allow it to handle extracting the ZIP archive. Something like this:

```python3
# ctmod_z0dxvk.py
def __extract(self, archive_path: str, extract_dir: str) -> bool:

    """
    Extract the tool archive at the given path.
    Return Type: bool
    """

    if not archive_path or not extract_dir:
        return False
    
    # DXVK and DXVK Async are 'tar.gz'
    if self.release_format.startswith('tar'):
        tar_type = self.release_format.split('.')[-1]

        return extract_tar(archive_path, extract_dir, mode=tar_type)

    # If unknown archive format, cannot extract, so default to fail
    return False

# ctmod_z2dxvknightly.py
def __extract(self, archive_path: str, extract_dir: str) -> bool:

    """
    Extract the tool archive at the given path.
    Return Type: bool
    """

    if not archive_path or not extract_dir:
        return False

    # DXVK Nightly is 'zip'
    if self.release_format == 'zip':
        return extract_zip(archive_path, extract_dir)

    # If unknown archive format, cannot extract, so default to fail
    return False
```

The more I think about it, the more I lean towards this approach, but more than happy to discuss it :-)

I think having a dedicated `__extract` method is also a good idea as it allows us to move our Ctmods closer towards a clean, common structure that we can begin to pick apart into a Ctmod base class (I've been harping on about this for like a year now lol, but I'm dedicated to the idea! :smile:)

pupgui2/resources/ctmods/ctmod_z2dxvknightly.py
```diff
@@ -3,99 +3,112 @@
...
         Get artifact from commit
         Return Type: str
         """
-        for artifact in self.rs.get(f'{self.CT_URL}?per_page=100').json()["artifacts"]:
-            if artifact['workflow_run']['head_sha'][:len(commit)] == commit:
+
+        for artifact in self.rs.get(f'{self.CT_ALL_ARTIFACTS_URL}?per_page=100').json()["artifacts"]:
+            # DXVK appends '-msvc-output' to Windows builds
+            # See: https://github.com/doitsujin/dxvk/blob/20a6fae8a7f60e7719724b229552eba1ae6c3427/.github/workflows/test-build-windows.yml#L80
+            if artifact['workflow_run']['head_sha'][:len(commit)] == commit and not artifact['name'].endswith('-msvc-output'):
```
sonic2kk, line 96: The `artifact['name']` ending with `-msvc-output`comes from the DXVK `test-build-windows.yml` Workflow file, as noted.

However this _could_ change _or_ other workflows could be added and we'd need to refactor this code to exclude those as well.

Perhaps it may be sufficient and more desirable to just check if our `artifact['name].endswith(artifact['workflow_run']['head_sha'])`. That way we can assume that workflow run names that end with just the commit SHA correspond to the `artifact.yml` builds that we want to download.

Something like this:

```python3
commit = artifact['workflow_run']['head_sha']
if artifact[:len(commit)] == commit and artifact.endswith(commit):
    # ...
```

Although it is also worth pointing out that just as the `test-build-windows.yml` suffix could change, or that new workflows could be added, so too could the `artifacts.yml` build naming change and get its own suffix. So either way this is going to be a little bit brittle imo, since we're depending on inferring from the artifact name.

Proton-tkg doesn't have this issue, since it works based on unique workflow IDs, but for DXVK Nightly I think it is more desirable to display the commit hashes.

Ideally we could return both the Workflow ID and the Commit SHA, but I don't think we have such a mechanism for a selected release to map information like that i.e. we can't have a "display name" for releases and then extract out the corresponding, say, Workflow ID from the selected version; we only support the releases dropdown returning to us the selected string.

DavidoTek:
Thanks!

I will do a complete review later.

> For cleanliness, I broke out the extraction logic for the downloaded archive into an __extract method. This puts all of the conditional checks for if we're working with a ZIP or a tarball into this method.  However, since this is broken out as a separate method, we could let the DXVK Nightly ctmod handle overriding the extract logic as well. Instead of the main DXVK Ctmod handling this, Ctmods that inherit and need different functionality can override this method. Happy to discuss this, and will leave a comment about this after posting the PR :-)
> The more I think about it, the more I lean towards this approach, but more than happy to discuss it :-)

Okay, feel free to do it with overwriting the `__extract` method.
We should, at some point, indicate which functions are supposed to be overwritten and which ones are not (`@final`-decorator since Python 3.8).

> I encountered some segfaults in initial development, but have not been able to reproduce it in the extensive testing that has followed.

That should not happen with pure Python. This is likely caused by something related to Qt (Signal, QMessageBox, Property, ...). I will see if I find anything.

> Reusing __fetch_workflows: If we want to make this function generic, we would need a way to define what it should return in its tags list (tags is not the best name, either, and is likely a holdover from working with release tags for tagged, non-nightly artifacts).

Is returning a dict an option? e.g. `{'commit': str(run['head_commit']['id'][:7]), 'id': str(run['id'])}`

> Perhaps it may be sufficient and more desirable to just check if our artifact['name].endswith(artifact['workflow_run']['head_sha']). That way we can assume that workflow run names that end with just the commit SHA correspond to the artifact.yml builds that we want to download.
> Although it is also worth pointing out that just as the test-build-windows.yml suffix could change, or that new workflows could be added, so too could the artifacts.yml build naming change and get its own suffix

Is there any way to get the workflow id like 6218184 from the artifact and then get the `path` from https://api.github.com/repos/doitsujin/dxvk/actions/workflows/6218184 or is there any other method that does not require many API calls?

I think the current implementation with checking the suffix for `-msvc-output` is fine for now.

sonic2kk:
I noticed a couple of oddities here...
- There are still sometimes duplicate hashes in this PR (currently I can see it listing `5caa867` twice, but that is the only one), I'll look into why and see if they are harmless. If they are, we can just ignore duplicate tags when adding.
- The hashes displayed on `main` (minus the duplicates) versus the ones displayed on this PR do not match. Some of the same commits are there, but in a different order.

Hashes listed on `main` are (we can't see as many because of the duplicates):

```
2e20d73
a1f6ea1
20a6fae
6cee1c1
62b99d7
6bdca15
24d9118
8df6a4d
```

Hashes listed on this PR are:

```
2e20d73
5caa867
a1f6ea1
9241ef2
20a6fae
62b99d7
390662f
0fdbb77
3e85146
```

### `main`

image
image


### This PR

image

<hr>

I will investigate where this discrepancy comes from.

sonic2kk:
Seems like the discrepancy noted in my above comment is because we weren't filtering by workflows run against the `master` branch. We were including all runs inside of that workflow, which included those against PRs, so we ended up with extra commits. These were _valid_ but we probably only want to include workflows run against `master`.

Now that we filter, the commits should match.

### `main` @ 44ec5917f6c5096795d9e6455e262048b9ce6935

image

image

### This PR

Note that this PR includes more commits because the list does not contain duplicate commits :-)

image

<hr>

With that fix, this PR should work. I have tested by checking that the commits should correspond to real `master` branch commits by clicking on the "Info" button on the CtInstaller dialog as well.

pupgui2/resources/ctmods/ctmod_z0dxvk.py
sonic2kk: The extraction logic has been reworked in bb3c556 so that the main DXVK ctmod only knows how to extract itself, and then DXVK Nightly, which needs different extraction logic (ZIP), overrides this method and handles ZIP extraction logic.

No ZIP extraction logic exists now in the main DXVK ctmod. It only has logic that relates to extracting itself (which overlaps with DXVK Async, but if this ever needed to change, we could override `__extract` in the DXVK Async ctmod as well).

sonic2kk:
PR is in a good state now for review. :slightly_smiling_face:

DavidoTek:
Thanks, looks very good! :tada:

The correct versions were downloaded in my testing and DXVK is still working as expected as well. I did not encounter any bugs while testing.

PR #514, opened 2025-04-01
util: Remove deprecated `Dict`, `List`, `Union`, `Tuple`, `Optional` type hints

sonic2kk:
Similar to #513, I thought this would be a trickier one to do and we would have a ton of usages, but going slow and by hand, it wasn't so bad :-)

This should be all of the deprecated type hints I think. I did a Find in Files for `from typing import` and the remaining usages outside of these last few PRs are all valid imports of `Any` and `Callable`.

### DIFF
--- a/pupgui2/util.py
+++ b/pupgui2/util.py
@@ -14,7 +14,7 @@ import random
 import zstandard
 
 from configparser import ConfigParser
-from typing import Any, Union, Tuple, Optional, Callable
+from typing import Any, Tuple, Optional, Callable
 
 import PySide6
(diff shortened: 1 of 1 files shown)


DavidoTek:
Thanks!

> `Callable | None`

I find it interesting thatit  is actually the official notation for the Option type: https://docs.python.org/3/library/stdtypes.html#types-union

I somehow would have expected something like `str?` from Python, but on the other hand, this is a very simple and unambiguous way to describe it.

> "Optional types can be spelled as a union with None: `str | None == typing.Optional[str]`"

PR #517, opened 2025-04-06
Ctmod: Add Proton-tkg (Wine Master NTSYNC)

### LINKED ISSUE #516: ProtonUpQT has ProtonTKG (Wine Master) but does not have the NTSYNC version.
ProtonUpQT has ProtonTKG (Wine Master) but does not have the NTSYNC version.

Add a Proton TKG (Wine Master) NTSYNC option 
https://github.com/Frogging-Family/wine-tkg-git/actions/workflows/proton-arch-ntsync-nopackage.yml

Manually just downloading and installing

**Additional context**
For use with NTSYNC kernels only.

sonic2kk:
Close #516.

This PR implements a Ctmod for downloading Proton-tkg with the NTSYNC patches for Wine Master.

I have tested that this successfully downloads and extracts a Proton-tkg version that is the same size as a manually downloaded archive. I also tested that the workflow ID in the dropdown matches that of the same verified Proton-tkg (Wine Master NTSYNC) ctmod (https://github.com/Frogging-Family/wine-tkg-git/actions/runs/14285965734).

However, I don't have a Kernel with the NTSYNC patches yet. I believe this is only included in Kernel 6.14+, and I am also not sure how to verify if NTSYNC is being used. But I have taken steps to verify that the archive downloaded is sane and matches what would be downloaded manually.

Note: It takes an absolute _age_ to download these archives from ProtonUp-Qt, probably because of `nightly.link` (as it is faster direct from GitHub, but not much we can do there :sweat_smile:).

Thanks!

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_protontkg_ntsync.py
+++ b/pupgui2/resources/ctmods/ctmod_protontkg_ntsync.py
@@ -1,0 +1,23 @@
+# pupgui2 compatibility tools module
+# Proton-Tkg https://github.com/Frogging-Family/wine-tkg-git
+# Copyright (C) 2022 DavidoTek, partially based on AUNaseef's protonup
+
+from PySide6.QtCore import QCoreApplication
+
+from pupgui2.resources.ctmods.ctmod_protontkg import CtInstaller as TKGCtInstaller  # Use ProtonTKg Ctmod as base
+
+
+CT_NAME = 'Proton Tkg (Wine Master NTSYNC)'
+CT_LAUNCHERS = ['steam', 'heroicproton', 'advmode']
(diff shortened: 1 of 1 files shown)


DavidoTek:
Thanks!

> I have tested that this successfully downloads and extracts a Proton-tkg version that is the same size as a manually downloaded archive. I also tested that the workflow ID in the dropdown matches that of the same verified Proton-tkg (Wine Master NTSYNC) ctmod (https://github.com/Frogging-Family/wine-tkg-git/actions/runs/14285965734).

Great. I also compared the `compatibilitytool.vdf` and it looks good.

> However, I don't have a Kernel with the NTSYNC patches yet. I believe this is only included in Kernel 6.14+, and I am also not sure how to verify if NTSYNC is being used. But I have taken steps to verify that the archive downloaded is sane and matches what would be downloaded manually.

Yeah, I guess there are quite a few \*sync options available by now. It should work then or at least is not something we need to worry about.

Issue #464, opened 2024-09-25
Add Sarek


pythonlover02:
**Is your feature request related to a problem? Please describe.**
Nop

**Describe the solution you'd like**
Adding Sarek or Proton-For-Old-Vulkan Builds: https://github.com/pythonlover02/Proton-Sarek/tree/Stable

**Describe alternatives you've considered**
Manually downloading the Sarek builds

**Additional context**
Sarek (Proton-For-Old-Vulkan) is a custom Proton build featuring DXVK 1.10.3 and DXVK 1.10.3 Async, tailored for users with GPUs supporting Vulkan 1.1+ (but not 1.3+) and non-Vulkan setups. It offers patched versions of Proton-GE for improved performance on old hardware, with little enhancements to WineD3D. Originally created for two friends with GT 710 GPUs, the goal is to allow users with older GPUs to run Proton-GE without needing to use older versions.

Im the creator and maintainer of the repo, i come here because some people would really like those builds to be on ProtonUp-Qt for a more easy access, anything that i need to know, or do i am at the orders 

pd: while i have a decent experience modifying code and using GitHub a im new to this of doing a public request, if i mess up anywhere just say it and i will fix it. And thx for reading :)

sonic2kk:
Sounds like a good idea, can you give more information on the technical details? Including:
- What archive type is used, and is this likely to change, or has this changed in the past? This could impact implementation a bit, if the extension has changed before.
- Does this project use standard GitHub releases only? If there are nightlies would we also want to add support for downloading these as well?
- What launchers are supported here? Steam seems like the obvious one of course, but what about Lutris, Heroic, and Bottles, which usually support Proton-based tools as well as those using Wine with patches and tweaks.
    - While this tool may work with these launchers I'm asking this more to get an understanding ofan "official" standpoint.
    - Lutris may detect this tool already if installed for Steam so we may get Lutris for free / whether we like it or not, I'd need to check though.
- Anything else you think might be relevant! 😄 

The repository can probably answer some of these but getting them directly from someone familiar with the project is also good, and having the answers here gives a good single reference point during implementation. :-)

> while i have a decent experience modifying code and using GitHub

I'd wait until the maintainer weighs in, but if you'd like you're free to implement this yourself! Using the GE-Proton ctmod as a base is probably good, it could probably be subclassed and used in a new file called something like `ctmod_sarek.py`.

[...]

PR #518, opened 2025-04-07
Ctmod: Add Proton-Sarek

sonic2kk:
Implements #464.

~~**NOTE**: [Proton-Sarek has regular and `-async` releases](https://github.com/pythonlover02/Proton-Sarek/releases/tag/Proton-Sarek9-27) for each of its tags, similar to what [Lutris-Wine does with fshack builds](https://github.com/lutris/wine/releases/tag/lutris-wine-7.2).~~

~~This PR does **not** currently display these, and will be in draft until I can implement this. Right now it seems to default to downloading the "-async" builds, but ideally we'd want to display "-async" as a separate tag just like we do for Lutris-Wine fshack builds. A similar approach may be a possible drop-in but I will work on it :-)~~

Support for downloading async and non-async builds was added in 517cdc0. The above note has been fully addressed :slightly_smiling_face: 

## Overview
This PR implements [Proton-Sarek](https://github.com/pythonlover02/Proton-Sarek) as a Ctmod (gated behind Advanced Mode, as per https://github.com/DavidoTek/ProtonUp-Qt/issues/464#issuecomment-2399641754) for Steam, Heroic Proton, and Bottles.

I tested that the ctmod downloaded and extracted with the expected file structure. The compatibility tool displays inside of Steam as expected and I tested using it with "Hammerwatch: Anniversary Edition". The game loaded fine using "Proton-Sarek9-27-async".

## Implementation
This is implemented by subclassing the GE-Proton ctmod and overriding the methods to fetch releases and fetch GitHub data, as they use different logic. A lot of this borrows from how we do things with Lutris-Wine, where we need to support fshack and non-fshack builds (see "Concerns" for more details).

Most of this ctmod is largely copy-pasted from NorthStar Proton, as it also subclasses GE-Proton. However I didn't believe it was good practice to subclass NorthStar Proton, in case any changes were made to that ctmod that made it no longer compatible. In other words, it just happens that they both take the same approach (and maybe eventually we'll have a better ctmod "type" that this Proton version and others can base themselves on).

## Concerns
The logic we use in `fetch_releases`, `__fetch_github_data`, `__get_data`, and `get_info_url` is almost identical to what we use for Lutris Wine, except with Lutris Wine we use `fshack` instead of `-async` as our identifier for  the rogue releases. We also need an extra `asset_condition` by default for Proton-Sarek to make sure we don't download the `-async` builds when we aren't using an `-async` tag.

My primary concern with this PR is that I would like to get rid of this duplication. I would like a way for us to keep this PR lean without the duplicated logic. I don't think tweaking and subclassing LutrisWine so that we can have a generic extra tag identifier (i.e. `fshack` for LutrisWine, `-async` for Proton-Sarek) is the solution. Maybe I need to sleep on it and think about it some more but in my head this smells of coupling the Proton-Sarek ctmod too "low down".

It is worth noting that if we do go the road of adding DXVK-Sarek, then it would hit the same issue as it has -async and non-async builds. It can likely be solved in the exact same manner, just subclassing the main DXVK Ctmod instead of GE-Proton, but we would again end up with a bunch of repeated code that I'd like to avoid if possible.

Perhaps there is just not a better solution for now. But I would like to discuss finding one before merging this PR!

I left a note about the behaviour requiring `asset_condition` in both cases of `is_async` being `True` and `False`, as I wonder if we should be stricter in [`get_download_url_from_asset`](https://github.com/DavidoTek/ProtonUp-Qt/blob/a82eeabea2714ea9f3b8031b70593f2d1c0ec517/pupgui2/util.py#L651) in the tag name that we check against to use exact matches by default, but I'm not sure if this would break anything. I haven't tested, but i have a hunch that it will cause problems. Then again, an `asset_condition` when this is needed does suffice, so it may not be worth the extra effort.

## Future Work
I wonder if it is worth exposing [DXVK-Sarek](https://github.com/pythonlover02/DXVK-Sarek) for Lutris and Heroic Wine. If so, that  may be better suited to wait for #509 to be merged.

Lastly, I wonder if this should be made available for Lutris. Now that we make GE-Proton available for Lutris, since Proton itself is more broadly compatible with Non-Steam environments thanks to UMU (discussion in #191, implementation in #496), should all of our Proton forks be made available for installation into Lutris, or even Heroic Proton? In my opinion, making these available for Lutris makes sense, but for Heroic Proton I'm not sure as I don't know if it uses UMU / Steam Linux Runtime (iirc it did maybe have some integration to use the Steam Linux Runtime, but I'm not sure of the current state).

<hr>

As usual, all feedback is welcome. Thanks!

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_protonsarek.py
+++ b/pupgui2/resources/ctmods/ctmod_protonsarek.py
@@ -2,6 +2,9 @@
 # pythonlover02's Proton-Sarek
 # Copyright (C) 2025 DavidoTek, partially based on AUNaseef's protonup
 
+from typing import Callable
+
+
 from PySide6.QtCore import QCoreApplication
 
 from pupgui2.util import fetch_project_release_data, fetch_project_releases
@@ -28,13 +31,13 @@ class CtInstaller(GEProtonInstaller):
         self.async_suffix = '-async'
 
     def fetch_releases(self, count: int = 100, page: int = 1) -> list[str]:
-        
+
         """
         List available releases
         Return Type: str[]
         """
 
-        include_extra_asset = lambda release: release['tag_name'] + self.async_suffix if any(self.async_suffix in asset['name'] for asset in release['assets']) else None
+        include_extra_asset: Callable[..., list[str]] = lambda release: [str(release.get('tag_name', '')) + self.async_suffix for asset in release.get('assets', {}) if self.async_suffix in asset.get('name', '')]
         return fetch_project_releases(self.CT_URL, self.rs, count=count, page=page, include_extra_asset=include_extra_asset)
 
     def __fetch_github_data(self, tag: str) -> dict:
--- a/pupgui2/resources/ctmods/ctmod_protonsarek.py
+++ b/pupgui2/resources/ctmods/ctmod_protonsarek.py
@@ -2,11 +2,9 @@
 # pythonlover02's Proton-Sarek
 # Copyright (C) 2025 DavidoTek, partially based on AUNaseef's protonup
 
-import os
-
 from PySide6.QtCore import QCoreApplication
 
-from pupgui2.util import fetch_project_release_data, ghapi_rlcheck
+from pupgui2.util import fetch_project_release_data, fetch_project_releases
 
 from pupgui2.resources.ctmods.ctmod_00protonge import CtInstaller as GEProtonInstaller
 
@@ -36,21 +34,10 @@ class CtInstaller(GEProtonInstaller):
         Return Type: str[]
         """
 
-        tags: list[str] = []
-        for release in ghapi_rlcheck(self.rs.get(f'{self.CT_URL}?per_page={count}&page={page}').json()):
-            if not 'tag_name' in release:
-                continue
+        include_extra_asset = lambda release: release['tag_name'] + self.async_suffix if any(self.async_suffix in asset['name'] for asset in release['assets']) else None
+        return fetch_project_releases(self.CT_URL, self.rs, count=count, page=page, include_extra_asset=include_extra_asset)
 
-            tags.append(release['tag_name'])
-            if 'assets' not in release or len(release['assets']) <= 0:
-                continue
-
-            if any(self.async_suffix in asset['name'] for asset in release['assets']):
-                tags.append(release['tag_name'] + self.async_suffix)
-
-        return tags
-
-    def __fetch_github_data(self, tag: str, is_async: bool) -> dict:
+    def __fetch_github_data(self, tag: str) -> dict:
 
         """
         Fetch GitHub release information
@@ -60,30 +47,8 @@ class CtInstaller(GEProtonInstaller):
         """
 
         # Exclude async builds by default -- Maybe 'get_download_url_from_asset' needs to be stricter?
-        asset_condition = lambda asset: self.async_suffix not in asset['name']
-        if is_async:
-            asset_condition = lambda asset: self.async_suffix in asset['name']
-
-        return fetch_project_release_data(self.CT_URL, self.release_format, self.rs, tag=tag, asset_condition=asset_condition)
-
-    def __get_data(self, version: str, install_dir: str) -> tuple[dict | None, str | None]:
-
-        """
-        Get needed download data and path to extract directory.
-        Return Type: tuple[dict | None, str | None]
-        """
-
-        is_async = self.async_suffix in version
-        if is_async:
-            version = version.replace(self.async_suffix, '')
-
-        data = self.__fetch_github_data(version, is_async)
-        if not data or 'download' not in data:
-            return (None, None)
-
-        protondir = os.path.join(install_dir, data['version'])
-
-        return (data, protondir)
+        asset_condition = lambda asset: self.async_suffix in asset['name'] if self.async_suffix in tag else self.async_suffix not in asset['name']
+        return fetch_project_release_data(self.CT_URL, self.release_format, self.rs, tag=tag.replace(self.async_suffix, ''), asset_condition=asset_condition)
 
     def get_info_url(self, version: str) -> str:
 
--- a/pupgui2/resources/ctmods/ctmod_protonsarek.py
+++ b/pupgui2/resources/ctmods/ctmod_protonsarek.py
@@ -13,7 +13,7 @@ from pupgui2.resources.ctmods.ctmod_00protonge import CtInstaller as GEProtonIns
 
 
 CT_NAME = 'Proton-Sarek'
-CT_LAUNCHERS = ['steam', 'heroicproton', 'bottles', 'advmode']
+CT_LAUNCHERS = ['steam', 'lutris', 'heroicproton', 'bottles', 'advmode']
 CT_DESCRIPTION = {'en': QCoreApplication.instance().translate('ctmod_protonsarek', '''A custom Proton build with <a href="https://github.com/pythonlover02/DXVK-Sarek">DXVK-Sarek</a> for users with GPUs that support Vulkan 1.1+ but not Vulkan 1.3, or for those with non-Vulkan support who want a plug-and-play option featuring personal patches.''')}
 
 
--- a/pupgui2/resources/ctmods/ctmod_protonsarek.py
+++ b/pupgui2/resources/ctmods/ctmod_protonsarek.py
@@ -2,9 +2,11 @@
 # pythonlover02's Proton-Sarek
 # Copyright (C) 2025 DavidoTek, partially based on AUNaseef's protonup
 
+import os
+
 from PySide6.QtCore import QCoreApplication
 
-from pupgui2.util import fetch_project_release_data, fetch_project_releases
+from pupgui2.util import fetch_project_release_data, ghapi_rlcheck
 
 from pupgui2.resources.ctmods.ctmod_00protonge import CtInstaller as GEProtonInstaller
 
@@ -25,6 +27,7 @@ class CtInstaller(GEProtonInstaller):
         super().__init__(main_window)
 
         self.release_format: str = 'tar.gz'
+        self.async_suffix = '-async'
 
     def fetch_releases(self, count: int = 100, page: int = 1) -> list[str]:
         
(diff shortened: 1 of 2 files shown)


pythonlover02:
> I wonder if it is worth exposing [DXVK-Sarek](https://github.com/pythonlover02/DXVK-Sarek) for Lutris and Heroic Wine. If so, that may be better suited to wait for #509 to be merged.

Maybe? I stopped using Wine + DXVK when UMU became stable, though I have friends who still use the old wine-ge-8-26 with some DXVK version, so I made [this](https://github.com/pythonlover02/wine-ge-custom-sk/releases/tag/wine-ge-8-26-sk) for them. I also think that a lot of people will continue to use Wine instead of Proton with UMU for a long time, so having the option would not hurt.

> Lastly, I wonder if this should be made available for Lutris. Now that we make GE-Proton available for Lutris, since Proton itself is more broadly compatible with Non-Steam environments thanks to UMU (discussion in #191, implementation in #496), should all of our Proton forks be made available for installation into Lutris, or even Heroic Proton? In my opinion, making these available for Lutris makes sense, but for Heroic Proton I'm not sure as I don't know if it uses UMU / Steam Linux Runtime (iirc it did maybe have some integration to use the Steam Linux Runtime, but I'm not sure of the current state).

There should be no problem with Proton-Sarek and UMU, after all its just GE-Proton patched with [DXVK-Sarek](https://github.com/pythonlover02/DXVK-Sarek), an older VKD3D and modifications to the `proton` file to add the rest of the features. 

About UMU on Heroic, as far as I remember UMU is available and enabled by default since [#3724](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3724), that being said I never had a pleasant experience with Heroic using Proton, it might have changed since the last time I tried it.

sonic2kk:
I investigated reducing code duplication and moving some of our complexity into `fetch_project_releases`, giving us a callable function to append extra releases based on a conditional. I came up with something along these lines:

```diff
diff --git a/pupgui2/util.py b/pupgui2/util.py
index 56cd74e..c8f9ec1 100644
--- a/pupgui2/util.py
+++ b/pupgui2/util.py
@@ -602,7 +602,7 @@ def is_online(host='https://api.github.com/rate_limit/', timeout=5) -> bool:
 
 
 # Only used for dxvk and dxvk-async right now, but is potentially useful to more ctmods?
-def fetch_project_releases(releases_url: str, rs: requests.Session, count=100, page=1) -> list[str]:
+def fetch_project_releases(releases_url: str, rs: requests.Session, count: int = 100, page: int = 1, include_extra_asset: Callable | None = None) -> list[str]:
 
     """
     List available releases for a given project URL hosted using requests.
@@ -621,7 +621,21 @@ def fetch_project_releases(releases_url: str, rs: requests.Session, count=100, p
     else:
         return []  # Unknown API, cannot fetch releases!
 
-    return [release[tag_key] for release in releases if tag_key in release]
+    releases_list: list[str] = []
+    for release in releases:
+        if tag_key in release:
+            releases_list.append(release[tag_key])
+        
+        if 'assets' not in release or len(release['assets']) <= 0:
+            continue
+        
+        if not include_extra_asset:
+            continue
+
+        if extra_asset := include_extra_asset(release):
+            releases_list.append(extra_asset)
+
+    return releases_list
 
 
 def get_assets_from_release(release_url: str, release: dict) -> dict:
```

I am not super attached to the `include_extra_asset` name, and additional testing would be needed to ensure this doesn't break anything (it shouldn't; we skip if `include_extra_asset` is not defined).

This allows our `fetch_releases` to look something like this:

```python3
def fetch_releases(self, count: int = 100, page: int = 1) -> list[str]:
    
    """
    List available releases
    Return Type: str[]
    """

    include_extra_asset = lambda release: release['tag_name'] + self.async_suffix if any(self.async_suffix in asset['name'] for asset in release['assets']) else None
    return fetch_project_releases(self.CT_URL, self.rs, count=count, page=page, include_extra_asset=include_extra_asset)
```

And this simplification can be carried over to LutrisWine as well.

<hr>

This removes the biggest section of duplicated logic, and is probably the only place we can actually cut down on the repeated complexity:
- `__fetch_github_data` is probably fine, ctmods can define their own `asset_conditions` and the implementation here, while very similar to LutrisWine, is not super complex. We could cut down complexity by only defining the lambda once, and using a common condition, but we can't avoid overriding it.
- `__get_data` is a little annoying to have to override, this is the main method I'd like to do something about to make it a bit cleaner. I wonder if `__fetch_github_data` can do the check itself for `is_async` instead of it needing to be passed.
- `get_info_url` is fine to override.

pupgui2/resources/ctmods/ctmod_protonsarek.py
```diff
@@ -0,0 +1,95 @@
...
+
+from PySide6.QtCore import QCoreApplication
+
+from pupgui2.util import fetch_project_release_data, ghapi_rlcheck
+
+from pupgui2.resources.ctmods.ctmod_00protonge import CtInstaller as GEProtonInstaller
+
+
+CT_NAME = 'Proton-Sarek'
+CT_LAUNCHERS = ['steam', 'heroicproton', 'bottles', 'advmode']
```
sonic2kk, line 15: As per https://github.com/DavidoTek/ProtonUp-Qt/pull/518#issuecomment-2784778814, once our implementation is finalised, we can implement support for Lutris as Proton-Sarek supports UMU.

sonic2kk:
I made some significant changes to the implementation in the last two commits. I implemented the changes I noted in `util#fetch_project_releases`. Feedback absolutely welcome on the implementation. I am not married to it, but I think this is the cleanest approach and will allow us to simplify LutrisWine in future, and we will get the same benefits if we implement DXVK-Sarek.

These changes allowed us to simplify the Proton-Sarek ctmod's `fetch_releases` by just using a lambda to define how to include extra assets. As well as the standard assets we also want to append release names ending with `-async` if releases contain an asset with `-async` in their name. Since Proton-Sarek (and other tools) include variations like this as release assets rather than separate tagged releases, we can't display them in ProtonUp-Qt. To fix this, we want to include extra tags in our dropdown that correspond to these assets. But we only want to do this if the release we're looking at has an asset with that extra variation. In this case, only if Proton-Sark has a release asset with `-async` in the title (historic releases don't have `-async` in the title, for example). So this allows us to display, for example, `Proton-Sarek9-27-async` in our dropdown even though it doesn't have a tag. This is the same idea of what we do for LutirsWine with displaying the `fshack` builds.

In `__fetch_github_data`, the `tag` that we get will contain this suffix. But this doesn't correspond to a release, so we won't be able to find it. So if it looks for `Proton-Sarek9-27-async`, it won't find it! In order to get around this, we need to give `fetch_project_release_data` a tag that doesn't include our suffix. This means it won't return the `-async` asset even if this is the release that was selected! To get around this we need to provide an `asset_condition` to `fetch_project_release_data`. This condition says "if we selected a tag with `-async` in it, then fetch_project_release_data select assets that have `-async` in their name. Otherwise, **exclude** assets that have `-async` in their name.". Effectively this means we call `fetch_project_release_data` with the a valid release tag, i.e. Proton-Sarek9-27 with no `-async` suffix, but if we selected a release that has `-async` (from the versions dropdown on the CtInstaller dialog) then we will _filter the assets on that release_ to make sure we _only get back_ the `-async` releases. Otherwise, _filter out releases that include `-async` in their name_.

This is the same idea as what we had before, but we were able to remove the override of `__get_data` since all it did was check if the `-async` suffix was in the `version` it gets called with, and then use that to feed a boolean `is_async` into a call to `__fetch_github_data`. However, there is no reason for us to do this, as we can just check on the `tag` we get from `__get_data` and allow `__fetch_github_data` to perform the check for if the selected version is `async` or not. It may be worth revisiting LutrisWine to see if we can simplify the logic there as well. However that can be done after this PR, especially if we go forward to the changes to `util#fetch_project_releases`, then we can do a small refactor to LutrisWine and simplify it a bit to take advantage of that change  and simplify its `__get_data` too in the same way we did here. 

<hr>

Some more testing is needed, however:
- In my testing, this appears to work, although I think I am running into network/rate limit issues as sometimes the download will stop right after starting, but on a small handful of runs, it will work. So I will need to test tomorrow.
- I want to make sure these changes to `fetch_project_releases` have not broken anything for the other ctmods. It should be safe, but I want to test more in-depth.

<hr>

All feedback welcome on the changes made! I realise these are substantial enough, but I think the recent commits simplified what we need well enough. Compared to the naive first commit before we did any release filtering, eb58b2e, we have implemented logic to download both async and non-async releases [with straightforward changes and only five new lines](https://github.com/DavidoTek/ProtonUp-Qt/compare/eb58b2ef6363d3edeaa4759caa5f8089151e775c...a715885e69e86a65935486af6367e67d74daae34#diff-7a73e01eda6898d95f18a77de28c062f6354bbfa157c0fb01b7fdfad866a0fb5) (excluding adding `get_info_url`, which is trivial).

pupgui2/util.py
```diff
@@ -621,7 +621,21 @@ def fetch_project_releases(releases_url: str, rs: requests.Session, count=100, p
...
+        if tag_key in release:
+            releases_list.append(release[tag_key])
+
+        if 'assets' not in release or len(release['assets']) <= 0:
+            continue
+
+        if not include_extra_asset:
+            continue
+
+        if extra_asset := include_extra_asset(release):
```
DavidoTek, line 635: One thing that came to my mind: What if there are multiple variations like `x`, `x-async`, `x-blah`?
Would it make sense to have `include_extra_asset` return a `list[str]` instead of `str`?

We already have the `asset_condition` with `fetch_project_release_data`. Since that only returns a boolean, it already allows having multiple suffixes.

DavidoTek:
Thanks!

Looks good, I like the implementation using `include_extra_asset`. I wonder if it would make sense to have that return a list instead (see my comment).

I still need to do the testing.

sonic2kk:
> I wonder if it would make sense to have that return a list instead (see my comment).

I plan to take a deeper look ASAP but the comment makes sense :+1: Makes this more robust and versatile for other tools that may need us to include an extra asset!

> Is there anything preventing the ctmod from supporting Lutris as of now?

I believe we would extract into the wrong directory for Lutris, since currently we would expand into the `install_dir` for Lutris which is i.e. `~/.local/share/lutris/runners/wine`, whereas we should be using i.e. `~/.local/share/lutris/runners/proton`. See https://github.com/DavidoTek/ProtonUp-Qt/issues/527#issuecomment-2811078457 for background and discussion, and #536 for _a_ solution as well as a proposed alternative.

Apart from that this Ctmod should be ready for Lutris :-)

pupgui2/util.py
sonic2kk: This was much more straightforward than I thought! I have made the relevant change and tested that I could download `-async` and non-`-async` builds. I did not test them with games but the folder names, structures, and the name in `compatibilitytool.vdf` looks sane. Woohoo! :smile:

sonic2kk:
Rebased this PR on top of `main` which should now include #536. This should allow us to extract for Lutris correctly now into the `runners/proton` folder, but I have not yet tested.

sonic2kk:
Got Lutris installed on this new installation and when installing, the `runners/proton` folder was created, and Proton-Sarek was downloaded. I tested Proton-Sarek9-27 and Proton-Sarek9-24-async, and both installed and extracted fine for Lutris. So the rebase ontop of a8532a7 was successful and does allow this Ctmod to install correctly for Lutris! :partying_face:

PR #525, opened 2025-04-09
Proton-tkg: Refactor to use `download_file`

sonic2kk:
This PR refactors the Proton-tkg ctmod to use the `download_file` util function. This is in a similar vein to other PRs that migrate compatibility tools to using 

I am eyeing up a larger refactor and cleanup of this ctmod (improved type hinting, reducing code duplication between it and DXVK Nightly following #509, extracting the archive extraction logic into an `__extract` method that each ctmod can implement instead of the Proton-tkg ctmod having to be aware of how to extract all of its child classes, maybe some other fun things in future too!) and this PR is the beginnings of that :-)

This PR is opened as a draft as it is largely untested, and I want to make sure it works against all the Proton-tkg ctmods. I tested briefly with the main Proton-tkg Ctmod and it appeared to work, but we have many child ctmods and I want to ensure each download and extract properly :-)

Thanks!

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_protontkg.py
+++ b/pupgui2/resources/ctmods/ctmod_protontkg.py
@@ -7,7 +7,9 @@ import glob
 import requests
 
 from PySide6.QtCore import QObject, QCoreApplication, Signal, Property
+from PySide6.QtWidgets import QMessageBox
 
+from pupgui2.networkutil import download_file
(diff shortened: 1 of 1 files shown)


sonic2kk:
Taking this out of draft as the code is ready for review. Kind of forgot about this and have not comprehensively tested it yet, but it should work okay enough that it's ready for review :-)

PR #526, opened 2025-04-09
vkd3d-proton: Refactor to use `download_file`

sonic2kk:
This PR refactors the vkd3d-proton Ctmod to use the `download_file` util function. Previously it was using the legacy download logic that `download_file` was created to replace, which cleans up the file a little :-)

I tested that vkd3d-proton and vkd3d-lutris ([which subclasses the vkd3d-proton ctmod](https://github.com/DavidoTek/ProtonUp-Qt/blob/69fa871a25dae1838090cf2b433afef8e620f5e0/pupgui2/resources/ctmods/ctmod_vkd3dlutris.py#L14)) both download and extract correctly following this change.

Thanks!


DavidoTek:
Thanks! :tada:

PR #528, opened 2025-04-11
test_util: Parametrize some tests

sonic2kk:
This PR refactors two tests in `test_util` - `test_get_dict_key_from_value` and `test_get_launcher_from_installdir` - to use PyTest's `parametrize` feature. The rationale here is that instead of constructing test data inside of the tests for all different scenarios (in this case, many different launcher scenarios), we can instead pass in a list of scenarios to test.

For example in `test_get_launcher_from_installdir`, we pass in the Steam launcher list, then the Lutris launcher list, etc, and then for each of the scenarios passed in, the test will run against those without us having to define all of that logic. This allows the test itself to be lean and generic. This also makes it easy to add additional scenarios.

We can even name these scenarios using an `id`, so that when we run tests we can see the exact parametrized scenario. Previously we wouldn't have had any insight, whereas now if one parametrized test fails we can see the name (ID) of the failed test.

<hr>

Note that `test_get_random_game_name` was left out of this, as we'd probably want in future to have some kind of PyTest fixture for games from each launcher that we can re-use. So for now the implementation was left alone :-)

<hr>

I pushed for testing in general in ProtonUp-Qt, and I've been slacking. I've been trying to find the best pattern to go with for writing and thus submitting tests, and have been trying to learn as much as I can about writing "good" tests so that we can end up with a robust test suite that will give us confidence in that green "tick".

Hopefully this is a good first step in improving our tests and giving a good pattern on writing tests going forward. I'm already looking into writing some tests for the non-happy-path scenarios for the existing tests, and expanding the test suite. But that will come at a later date. :wink: 

Thanks!

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -76,38 +76,44 @@ def test_get_dict_key_from_value(test_dict: dict[str | int | Launcher, str | str
         assert type(retrieved_key) is expected_type
 
 
-def test_get_launcher_from_installdir() -> None:
+@pytest.mark.parametrize(
+    'launcher_paths, expected_launcher', [
+        pytest.param(
+            [ install_location['install_dir'] for install_location in POSSIBLE_INSTALL_LOCATIONS if install_location['launcher'] == 'steam' ],
+            Launcher.STEAM,
+            id='Steam Launcher'
+        ),
+        pytest.param(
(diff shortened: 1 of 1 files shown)


sonic2kk:
I found a way to parametrize `get_random_game_name`, so I will push that to this branch in a sec :-)

PR #529, opened 2025-04-12
test_util: Implement tests for `is_online`

sonic2kk:
This PR implements unit tests for the `is_online` util function. We test the "happy path" (a successful response where `is_online` returns `True` and the response we get matches a sane successful response) and the "error path" (`is_online` returns `False` when we catch a `requests.ConnectionError` or a `requests.Timeout` error, and the response matches an expected stubbed error response).

Initially, I had written these tests out using [`requests-mock`](https://github.com/jamielennox/requests-mock), but [it seems that library is unmaintained](https://github.com/jamielennox/requests-mock/issues/262). So I rewrote the tests to use the [Responses library](https://github.com/getsentry/responses) which is more actively maintained.

For convenience, I opted to use `pytest-responses` which is a dependency on top of `responses` for use with PyTest. It allows `responses` to be used as a fixture, which to my understanding means we don't have to do the setup/teardown of a Responses RequestMock ourselves.

So we use the Responses library in order to mock the responses of `requests.get` in our `is_online` tests. But in order for this to run in CI, I had to make a couple of extra changes:
- I created a `tests/requirements.txt` file that defines dependencies that are _only_ required for running tests. We already have a few of these, including PyTest itself. They're defined in [GitHub Actions](https://github.com/DavidoTek/ProtonUp-Qt/blob/0c418459dd0dc27d9b20fd3757371a54b2b0e3cf/.github/workflows/testing-ci.yml#L29). I took these dependencies and put them into the `tests/requirements.txt` file.
- Now that we have this file, we can replace the GitHub Actions step of installing PyTest and friends with `pip install -r tests/requirements.txt`, replacing the hardcoded list of requirements. This should also make it easier for contributors to install and get up to speed with the requirements needed to run and write unit tests for the project.

I realise this PR introduces a new dependency _and_ touches the CI, so feedback is welcome on what I've done and if anything should be done differently. I think using Responses is sane and will set us up nicely for testing and mocking other network calls (maybe even testing `ghapi_rlcheck` and `glapi_rlcheck`). But if there's a better way to make sure these dependencies are used in CI and/or having `tests/requirements.txt` is undesirable, I'm happy to look into other approaches :-) 

Thanks!

### DIFF
--- a/.github/workflows/testing-ci.yml
+++ b/.github/workflows/testing-ci.yml
@@ -26,7 +26,7 @@ jobs:
     - name: Install dependencies
       run: |
         sudo apt install -y libegl1 libxkbcommon0
-        pip install pytest pytest-md pytest-emoji
+        pip install -r tests/requirements.txt
 
     - name: Install ProtonUp-Qt
       run: pip install -e .
--- a/pupgui2/util.py
+++ b/pupgui2/util.py
@@ -587,7 +587,7 @@ def is_gitlab_instance(url: str) -> bool:
     return any(instance in url for instance in GITLAB_API)
 
 
-def is_online(host='https://api.github.com/rate_limit/', timeout=5) -> bool:
+def is_online(host: str = 'https://api.github.com/rate_limit/', timeout: int = 5) -> bool:
     """
     Attempts to ping a given host using `requests`.
     Returns False if `requests` raises a `ConnectionError` or `Timeout` exception, otherwise returns True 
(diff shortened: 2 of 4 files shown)


DavidoTek:
Thanks! Looks good :tada:

> CI and/or having tests/requirements.txt

I think that's fine. Alternatively, we could put it in the top folder and call it `test-requirements.txt` or similar, but I guess we can leave it in the tests folder.

PR #530, opened 2025-04-12
test_util: Implement test for `is_gitlab_instance`

sonic2kk:
This PR implements a test for the `is_gitlab_instance` utility function.

We test against each element in our `GITLAB_API` list constant using [PyTest's Parametrize feature](https://docs.pytest.org/en/7.1.x/example/parametrize.html). Currently we only have one entry in this list, but if we had more in future, this would allow us to check to make sure if we add one that it this function should always work with all items in that list. We also then check a failing test case with the GitHub API to ensure that the function returns `False` when we expect it to.

Thanks!

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -1,6 +1,8 @@
+import pytest
+
 from pupgui2.util import *
 
-from pupgui2.constants import POSSIBLE_INSTALL_LOCATIONS
+from pupgui2.constants import POSSIBLE_INSTALL_LOCATIONS, GITLAB_API, GITHUB_API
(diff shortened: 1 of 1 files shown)


DavidoTek:
Awesome to see that the number of tests is growing, thanks!

PS: I may be a bit slow at reviewing currently due to some other projects...

sonic2kk:
No problem! I spent a lot of time trying to figure out PyTest and good testing standards / practices, and when I finally felt like I understood, I decide to try and make good on my promise from a while back to actually write some tests.

The PRs for the tests are not testing functionality that I foresee changing, so there's no need to hurry on review :-) I did test merging them into `main` locally and it seems there'll be conflicts at some point so I'll resolve those once we get to them!

PR #533, opened 2025-04-12
test_util: Implement test for `get_combobox_index_by_value`

sonic2kk:
This PR implements a unit test for the `get_combobox_index_by_value` utility function.

I tested this test with and without the `QT_QPA_PLATFORM=offset` environment variable. It's worth noting that [we set this environment variable in CI](https://github.com/DavidoTek/ProtonUp-Qt/blob/0c418459dd0dc27d9b20fd3757371a54b2b0e3cf/.github/workflows/testing-ci.yml#L37).

When running multiple tests that use a `QApplication`, I was running into an error along the lines of `Please destroy the QApplication singleton before creating a new QApplication instance.` This would happen if a test created a `QApplication()` and then another test tried to do the same thing.

Trying to fix this took a bit of research. I stumbled across [this comment on a thread](https://forum.qt.io/topic/153620/please-destroy-the-qapplication-singleton-before-creating-a-new-qapplication-instance/6?_=1744426079662&lang=en-US) that suggested using `QApplication.shutdown()`. This resolved the problem.

Thanks!

### DIFF
--- a/tests/test_ctmods.py
+++ b/tests/test_ctmods.py
@@ -14,7 +14,7 @@ def test_ctmod_loader() -> None:
     """
     Test loading of ctmods using CtLoader.
     """
-    QApplication()
+    app = QApplication()
     
     dummy_main_window = DummyMainWindow({})
 
@@ -29,3 +29,5 @@ def test_ctmod_loader() -> None:
 
     # Ensure that no advanced mode ctmods are loaded when advanced_mode is False
     assert(all(["advmode" not in ctmod.CT_LAUNCHERS for ctmod in ct_loader.get_ctmods(launcher=None, advanced_mode=False)]))
+
+    QApplication.shutdown(app)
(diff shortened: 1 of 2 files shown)


sonic2kk:
I was taking a look at another test that requires a `QApplication` instance (`ghapi_rlcheck` util) and it needs the `message_box_message`. To solve this we can instantiate a `PupguiApp` object, which is a child of `QApplication`.

I wonder if this is a better approach, and if we should have a `PupguiApp` as a fixture for all tests?

DavidoTek:
> QT_QPA_PLATFORM=offset

Do you mean `QT_QPA_PLATFORM=offscreen`? Then that's correct.

> I wonder if this is a better approach, and if we should have a PupguiApp as a fixture for all tests?

I think we have to do that. `ghapi_rlcheck`/`glapi_rlcheck` would crash otherwise, in the case of a rate limit.

While reviewing the "test" PRs, I just noticed how many side effects some of our functions have. That's making testing a bit complicated and possibly unpredictable...

PR #534, opened 2025-04-12
Add styling for QScrollBar/QComboBox controls for the Steam Deck theme

loathingKernel:
Adds theming for QScrollBar/QComboBox controls for the Steam Deck theme

Preview:
image

The icon file paths are injected in to the CSS string dynamically right now. I would like to move them into a QResources file but I wanted to get the ok from you before going forward with that. The QResources as far as the theme is concerned should include the QSS file along with any images it uses.


loathingKernel:
I've added example of using QResources and generating the stylesheet through `qstylizer` so it is more clear what kind of handling I was referring to previously. The qstylizer code is probably over-complicated for this use-case but I wanted to demonstrate what is possible with it.

loathingKernel:
> I'm already not very happy with the way the translation files are handled since both the source code and the compiled .qm files are in the repository.

I have the same issue in Rare, but it is what it is until I get some time to figure out how to use buildtools to build them at package creation.

> Since the style is defined imperatively in the stylesheet.py, wouldn't it be possible to directly import it from that file without going the route with the .qss file?

You could, but you would need to depend on `qstylizer` at runtime to generate the `qss` stylesheet string. So it's what you consider less of an issue. For the icons to work with the stylesheet tho, they need to be either absolute paths or be part of the same QResource file as the stylesheet. So in the case of the former you do need extra manipulation to inject them. Likely not an issue if you are generating the style at runtime, but in my opinion makes for a more complicated setup overall.

So it boils down to whether you want to include another dependency that doesn't do anything but generate the stylesheet.

Issue #527, opened 2025-04-11
Lutris and Steam GE-Proton conflict also Lutris not detecting GE-Proton


mc711:
Please fill out following when reporting a new bug:

**Describe the bug**  
If GE-Proton is installed on both Steam and Lutris, there is a conflict. both during selection and deletion.

**To Reproduce**  
1. Install GE-Proton-27 (as of this bug) on both Steam and Lutris through ProtonUp-Qt
2. try to make GE-Proton-27 default wine runner in lutris (Preferences -> Runners -> Wine -> Options -> Runner Options-> Wine Version) , it will select the Steam version (verified by selecting the Manage wine versions button next to the wine runner options button, GE-Proton-27 will not show any games)
3. uninstall GE-Proton-27 from steam with ProtonUp-Qt, lutris will no longer show GE-Proton-27 (GE-Proton-27 folder is still in lutris wine runner folder, even after reinstallation of GE-Proton on lutris, still not showing on dropdown list)
4. 

**Expected behavior**  
Lutris should be able to select the installed GE-Proton-27 and also show the amount of games using this runner (in the manage wine versions)

**Screenshots**  
![Image](https://github.com/user-attachments/assets/77d5a4ae-80c5-45ea-832d-c8609128baa3)

![Image](https://github.com/user-attachments/assets/2da07808-faea-4236-8c1a-2db2f1bd6024)

![Image](https://github.com/user-attachments/assets/bf6c528b-3d41-4e6a-9f86-d0bfebaec14f)

![Image](https://github.com/user-attachments/assets/54ab3368-58d3-4fc6-a48c-c6c76ccacc1d)

![Image](https://github.com/user-attachments/assets/2d1429e8-00b6-4a6c-be02-5adf0e0a1266)

**Desktop (please complete the following information):**  
 - **Platform**: Steam Deck
 - **System**: SteamOS 3.6.24
 - **Version**: ProtonUp-Qt 2.12.1
 - **How did you install ProtonUp-Qt?**: Flatpak

DavidoTek:
I think this are multiple issues.

> uninstall GE-Proton-27 from steam with ProtonUp-Qt, lutris will no longer show GE-Proton-27 (GE-Proton-27 folder is still in lutris wine runner folder, even after reinstallation of GE-Proton on lutris, still not showing on dropdown list)
> Lutris should be able to select the installed GE-Proton-27

[...]

PR #536, opened 2025-04-17
GE-Proton: Use correct extract path for Lutris

sonic2kk:
See also: #527

## Overview
This PR changes the extract directory for GE-Proton, extracting it into Lutris' `runners/proton` folder instead of into `runners/wine`. Lutris expects Proton-based compatibility tools that run with UMU to be installed into `runners/proton` (this path being relative to the Lutris install location, i.e. `~/.local/share/lutris/runners/proton`).

Note that this is **one** approach, and I will detail an alternative approach at the bottom of this PR.

### Use Correct Extract Dir for GE-Proton
I added a \`get_extract_dir\` method to the GE-Proton Ctmod to set the value of `install_dir` to the expected extract dir for the launcher. Since for every other launcher we want to trust `install_dir`, we can return that as the default value from `get_extract_dir`. However if we're using Lutris then we need to tell it to use the `runners/proton` folder, so we return that.

This pattern of using `os.path.abspath(os.path.join(install_dir, ...))` is used by other ctmods as well, such as [DXVK](https://github.com/DavidoTek/ProtonUp-Qt/blob/0c418459dd0dc27d9b20fd3757371a54b2b0e3cf/pupgui2/resources/ctmods/ctmod_z0dxvk.py#L164-L176).

A nice benefit of this approach is that if we enable downloading other Proton-based compatibility tools for Lutris, if they subclass GE-Proton, they will use the correct extract directory for free! However, if Wine-based compatibility tools inherit from GE-Proton (such as Lutris-Wine), then they will **not** and they would need to override `get_extract_dir`.

### Use Correct Extract Dir for Lutris-Wine and Kron4ek Wine
Lutris-Wine inherits from the GE-Proton ctmod, as does Kron4ek Wine. Since now when GE-Proton detects a Lutris `install_dir` it tries to extract into `runners/proton`, this breaks these two Wine-based compatibility tools  that should instead go into the default `install_dir` path, which is `runners/wine`. Therefore, we override the `get_extract_dir` method in these Ctmods and just return `install_dir` always. Since they are Wine-based, they can always use the default path.

We would probably instead at some point want a higher level ctmod base class that by default extracts into `install_dir`, and tools like GE-Proton, DXVK, and vkd3d can all inherit from this and override the install path as needed, rather than the Wine-based Lutris tools inheriting from GE-Proton and needing to override its default.

## Testing
I thought about implementing a test suite for this, to test that the correct install directory is used for GE-Proton and all Ctmods that inherit from it. However, I ran into a couple of snags.

The overall conclusion I came to was that writing tests for ctmods can wait for now. :sweat_smile: There are probably some wider design questions of our ctmod classes that would need revisiting before we start writing tests for methods that we shouldn't directly test.

### Displaying Proton versions on Main Menu Tool List
Since GE-Proton is not installed into the standard Lutris `install_dir`, it is not visible by default on the Main Menu tool list. We had the same problem for DXVK and vkd3d-proton, and to address it, we have a pattern in place to fetch tools in a given directory if their their display name contains a substring. For example, for DXVK we search `runtime/dxvk` and list all directories in that folder that have `dxvk` in their name. We do the same for vkd3d.

The same pattern can be re-used here for GE-Proton, and it was a straightforward enough change to re-use the `MainWindow#get_installed_versions` method to do it.

### Needing to Stub `MainWindow`
We need a `main_window` parameter to initialise the ctmod. We could create one as a fixture to pass around many of our tests, which probably makes sense. But I'm not sure if there are parts we'd want to stub out. We would maybe want a `DummyMainWindow` fixture. We do have a `DummyMainWindow` in our `test_ctmods` but we may need something more thorough than that in future, something closer to a real instance of `MainWindow`, or a stubbed-out instance of `MainWindow` where required. We could even have a "`MainWindow`-ish" fixture only for ctmod tests by using a `conftest.py` inside of a `tests/ctmods` directory.

The only time we may not want a fixture is if we're testing the `MainWindow` itself. So setting up the tests could be a bit of legwork that might be better served as a separate endeavour.

### Visibility of `CtInstaller#get_extract_dir`
However, even supposing that we do that legwork in this PR, we run into a problem. Should we be directly testing `get_extract_dir`, and more to the point, should `get_extract_dir` be "public"? It's "public" in other classes but it should probably not be, since it shouldn't be used by outside classes. Access in Python is of course not as strict as in other languages, it's convention, but we still use `__method_name` for other methods in Ctmods.

If it should be "private", then even though we _can_  test this method, whether or not we _should_ test "private" methods up for debate. In other languages while you _can_ do it you usually don't want to, and instead want to test it implicitly through methods that would use it. So in our case we'd want to test `get_tool`.

Trying to figure this out led me down a rabbit hole of, _should private methods be overridden_. This is not something I got a good answer to in the context of Python. In other languages like Java, private methods can't be overridden afaik. Python is probably more flexible, but I still don't know what the best practice is.

## Alternative Approach
This PR is a little... _hacky_. A more sustainable approach may be to take inspiration from what we do with Heroic and have a `lutriswine` and `lutrisproton` install location. For `lutrisproton` we would add new entries into our `POSSIBLE_INSTALL_LOCATIONS` constant, something like `{'install_dir': '~/.local/share/lutris/runners/proton/', 'display_name': 'Lutris Proton', 'launcher': 'lutrisproton', 'type': 'native', 'icon': 'lutris', 'config_dir': '~/.config/lutris'}`. Equivalent for Flatpak as well. Then for the existing Lutris entries, we would refactor them to be `lutriswine` instead of just `lutris`.

If we implement such a thing, right now there would only be one Proton-based Ctmod. However we could easily then add a bunch of Proton-based compatibility tools, since they could just use `install_dir`. We also wouldn't need any implementation of `get_extract_dir` for GE-Proton, and we wouldn't need.

This approach is one that I thought of after implementing this PR. It would be a fairly significant refactor as well. If we want, we can go with this PR for now, and maybe create an issue detailing breaking Lutris out into two separate launcher types. Maybe we'd want to wait until we have tests in place for ctmods before doing such a refactor to give us a bit more confidence that our refactor won't screw something up. Or perhaps we want to scrap this PR's approach, prioritise Ctmod unit tests, and then look at a refactor without having this throwaway work.

Taking this approach also means we wouldn't need to do the `self.get_installed_versions('proton', proton_dir)` call, since a `lutrisproton` ctmod would have its `install_dir` as `runners/proton` already and would be able to find these tools.

I guess if we wanted to make a release quickly with a fix for using the correct extract dir, we could use this PR to do it as a quick-fix and then spend time looking at a solution like having `lutriswine` and `lutrisproton`.

<hr>

If we want to go with the alternative approach, I'm happy to close this PR and we can discuss how we want to approach breaking Lutris out into `lutriswine` and `lutrisproton`.

Either way, all feedback is welcome on this PR. Thanks!

### DIFF
--- a/pupgui2/pupgui2.py
+++ b/pupgui2/pupgui2.py
@@ -242,9 +242,11 @@ class MainWindow(QObject):
         if install_loc.get('launcher') == 'lutris':
             dxvk_dir = os.path.join(install_directory(), '../../runtime/dxvk')
             vkd3d_dir = os.path.join(install_directory(), '../../runtime/vkd3d')
+            proton_dir = os.path.join(install_directory(), '../../runners/proton')
 
             self.get_installed_versions('dxvk', dxvk_dir)
             self.get_installed_versions('vkd3d', vkd3d_dir)
+            self.get_installed_versions('proton', proton_dir)
         # Launcher specific (Steam): Number of games using the compatibility tool
         elif install_loc.get('launcher') == 'steam' and 'vdf_dir' in install_loc:
             steam_app_list = get_steam_app_list(install_loc.get('vdf_dir'), cached=False)  # update app list cache
--- a/pupgui2/resources/ctmods/ctmod_00protonge.py
+++ b/pupgui2/resources/ctmods/ctmod_00protonge.py
@@ -9,7 +9,8 @@ import hashlib
 from PySide6.QtWidgets import QMessageBox
 from PySide6.QtCore import QObject, QCoreApplication, Signal, Property
 
-from pupgui2.util import fetch_project_releases, ghapi_rlcheck, extract_tar
+from pupgui2.datastructures import Launcher
+from pupgui2.util import fetch_project_releases, get_launcher_from_installdir, ghapi_rlcheck, extract_tar
 from pupgui2.util import build_headers_with_authorization
 from pupgui2.networkutil import download_file
 
@@ -148,12 +149,17 @@ class CtInstaller(QObject):
         Download and install the compatibility tool
         Return Type: bool
         """
+
+        install_dir = self.get_extract_dir(install_dir)
+
         data, protondir = self.__get_data(version, install_dir)
         if not data:
             return False
 
+        # Note: protondir is only used for checksums
         if not protondir or  not os.path.exists(protondir):
-            protondir = os.path.join(install_dir, 'Proton-' + data['version'])
+            protondir = os.path.join(install_dir, 'Proton-' + data['version'])  # Check if we have an older Proton-GE folder name
+
         checksum_dir = f'{protondir}/sha512sum'
         source_checksum = self.rs.get(data['checksum']).text if 'checksum' in data else None
         local_checksum = open(checksum_dir).read() if os.path.exists(checksum_dir) else None
@@ -182,6 +188,26 @@ class CtInstaller(QObject):
         self.__set_download_progress_percent(100)
 
         return True
+
+    def get_extract_dir(self, install_dir: str) -> str:
+
+        """
+        Return the directory to extract GE-Proton archive based on the current launcher
+        Return Type: str
+        """
+
+        launcher = get_launcher_from_installdir(install_dir)
+        extract_dir: str = install_dir
+
+        if launcher == Launcher.LUTRIS:
+            # GE-Proton for Lutris needs to go into 'runners/proton' and not 'runners/wine'
+            extract_dir = os.path.abspath(os.path.join(install_dir, '../../runners/proton'))
+            
+            # Lutris may not be guaranteed to make this path, so ensure it exists
+            if not os.path.exists(extract_dir):
+                os.mkdir(extract_dir)
+
+        return extract_dir  # Default to install_dir
 
     def get_info_url(self, version: str) -> str:
         """
--- a/pupgui2/resources/ctmods/ctmod_kron4ekvanilla.py
+++ b/pupgui2/resources/ctmods/ctmod_kron4ekvanilla.py
@@ -56,3 +56,15 @@ class CtInstaller(GEProtonInstaller):
         ldd_min = int(ldd_ver.split(b'.')[1])
         return False if ldd_maj < 2 else ldd_min >= 27 or ldd_maj != 2
 
+    def get_extract_dir(self, install_dir: str) -> str:
+
+        """
+        Return the directory to extract Lutris-Wine archive based on the current launcher
+        Return Type: str
+        """
+
+        # GE-Proton ctmod figures out if it needs to into a different folder
+        #
+        # kron4ek can use default 'install_dir' always because it is Wine and not Proton,
+        # so override to return unmodified 'install_dir'
+        return install_dir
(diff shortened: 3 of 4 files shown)


sonic2kk:
Whoops, turns out I forgot to push the change for Kron4ek Wine until now... Fixed :-)

DavidoTek:
> We would maybe want a DummyMainWindow fixture
> We could even have a "MainWindow-ish" fixture only for ctmod tests by using a conftest.py inside of a tests/ctmods directory

Yeah, it likely makes sense to have a DummyMainWindow fixture in conftest.py. Not sure which features we would exactly need, except the `web_access_tokens` and `get_installed_versions` (do we actually need the latter in the fixture or is a simple `assert os.path.exists(...)` sufficient?)

> Visibility of CtInstaller#get_extract_dir
> Should we be directly testing get_extract_dir, and more to the point, should get_extract_dir be "public"?

Good thought.
I feel like having that as a public function would make sense. For example, this would allow displaying the user where the tool will be installed based on the install_dir before it is installed (not sure if anyone needs that but it would be possible).
In that case, we can just test it.

> In other languages while you can do it you usually don't want to, and instead want to test it implicitly through methods that would use it. So in our case we'd want to test get_tool.
> Trying to figure this out led me down a rabbit hole of, should private methods be overridden

Hmm. Some of the private functions are becoming more universal and I guess you already suggested moving them into a util file at some point. In that case, we can just add a test function for that util file.

It's a bit more complicated for the other functions that depend on overwritten private functions. Take `__get_data`, for example. It behaves differently when it is run in GE-Proton or Kron4ek since `__fetch_github_data` is overwritten there.
I'm worrying less about convention and more about clarity. I think a clean approach here would be to have `__get_data` as a utility function and then pass `__fetch_github_data` as a parameter like with the `asset_condition` although I don't think changing that is worth the effort.

So I'm not settled on how we should handle that and we should see if it becomes necessary to test the private functions.

---

> This PR is a little... hacky. A more sustainable approach may be to take inspiration from what we do with Heroic and have a lutriswine and lutrisproton install location

I guess the cleanest approach would be to use a list/dict for install_dir like `'install_dir': {'proton': '...', 'wine': '...'}` and then have a function for getting the correct install directory. Ideally, there would be an object/class for each launcher where we could do something like `mylauncher.get_install_dir(CtoolTypes.WINE)`
That would require a lot of refactoring.

> If we want, we can go with this PR for now
> Maybe we'd want to wait until we have tests in place for ctmods before doing such a refactor to give us a bit more confidence that our refactor won't screw something up

That sounds like a good plan :smile:

I will merge this for now. The solution is simple and sticks to our existing solution for "different tool directories".

> See also: #527

Does this fix #527?


Thanks

sonic2kk:
(Meant to reply yesterday, sorry!)

> Does this fix https://github.com/DavidoTek/ProtonUp-Qt/issues/527?

It's been a lil bit but looking back, I think it should fix that issue. The problem in that issue I believe were twofold:
- That Lutris would display GE-Proton builds installed for Steam only (intended Lutris functionality)
- That Lutris couldn't see GE-Proton builds installed just for Lutris unless they were also installed for Steam (this was caused by us extracting into `runners/wine` instead of `runners/proton`, which this PR fixes)
    - We can't do anything about Lutris displaying _both_ for example, GE-Proton9-27 if it's installed for Lutris _and_ Steam at the same time. A user would just have to know that this is not intended for Lutris to handle. I believe this is outside the scope of ProtonUp-Qt.

So the piece of that issue that I believe to be user error is not handled by this PR, but the functionality issue that we are extracting to the wrong directory (very kindly investigated and pointed out by https://github.com/DavidoTek/ProtonUp-Qt/issues/527#issuecomment-2811078457) should be solved by this PR. If we take the latter as solving the issue then I am happy to say that this PR fixes #527.

PR #538, opened 2025-04-23
test_util: Parametrize `get_random_game_name`

sonic2kk:
Related: #528 

Uses PyTest Parametrize for the `get_random_game_name` function. Also adds some more test scenarios for ensuring we always get a `str` back, and to ensure we properly handle unknown objects and empty lists :-)


sonic2kk:
Failing because the environment doesn't have `typing_extensions`. No idea how or why I have it, I only used `LiteralString` because `basedPyright` didn't want me typing as `list[str]`. I'll fix it when I'm back at my PC. :-)

PR #540, opened 2025-04-23
Update appimage-ci.yml

scgm0:


sonic2kk:
Seems the Action we currently used was Archived some time ago, and the one used in this PR is an offering from GitHub. Looks good to me.

DavidoTek:
One thing I did not think of when merging this is that it does not automatically create a new release and attach the files. See https://github.com/DavidoTek/ProtonUp-Qt/releases/tag/v2.13.0

We should change this.

PR #542, opened 2025-04-25
test_dbusutil: Add tests for dbusutil functions

sonic2kk:
Depends on: #529 (this PR requires `pytest-mock` introduced in the linked PR).

## Overview
This PR implements unit tests for our dbusutil functions. This file consists of two utility functions: A generic one to create and send a DBus message on a given bus or the session bus, and a specific one to create a progress message to send to DBus to display an expected progress bar and count on supported Task Managers.

While coverage isn't always a metric to obsess over, `pytest-cov` indicates that this test gives us 100% coverage on `dbusutil.py`!

## Implementation
We test `create_and_send_dbus_message` to ensure it sends an expected constructed message. Since other tests will test multiple message variations, I didn't feel the need to Parametrize this test with multiple message cases. We build a DBus message to send and then we check that the `QDBusConnection.send` method was actually called and called with the values that we expect from our constructed message.

We check inside of `create_and_send_dbus_message` if it has a DBus connection and we only send the message if it does. If it doesn't have a connection, we will default to returning `False`. I implemented a test to cover this specific case, where we just mock the `isConnected` method of `QDBusConnection` to always return `False`. This means we never get to the point of calling `QDBusConnection.send`, so we check to make sure that `send` is not called and that we return `False` from our call (if `send` was not called, the only return value that we have should be from returning `False` ourselves).

Finally we test `dbus_progress_message` where we call it with an expected `count` and `progress`. This function is an abstraction on top of `create_and_send_dbus_message`. Since the progress message we want to send has a fixed structure and we don't need to care about the underlying manual message components. We just give it a `progress` value between `0` and `1`, and a `count` value, and it does the rest. For our test I decided to Parametrize it because I wanted to check that using different values for `progress` and `count` work as expected. When `progress >= 1` we don't display the progress anymore, and when `count <= 0` we don't display the count either. So I wanted to test these scenarios and uses Parametrization to do it.

## Gaps
The only gap in the test that I can see is that we don't have a test case to cover manually providing a `bus` object, but I don't think we need to test that case too much. Plus, coverage is already at 100%, so we should be fine :-)

We also don't check in `dbus_progress_message` that `create_and_send_dbus_message`'s call to `QDBusConnection.send` is sent with the expected parameters. I chose not to get this granular because I feel that testing that `create_and_send_dbus_message` here is unnecessary since it has its own test. 

<hr>

Like with #541, I opened this PR despite it having dependencies because this is all "net new". This also entirely covers the tests in dbusutil, so since the tests in this PR are also quite standalone I think it is worth opening separately :-)

As usual, all feedback is welcome. Thanks!

### DIFF
--- a/tests/test_dbusutil.py
+++ b/tests/test_dbusutil.py
@@ -1,3 +1,5 @@
+import pupgui2.dbusutil  # needed for spies
+
 import pytest
 
 from pytest_mock import MockerFixture
@@ -5,10 +7,8 @@ from pytest_mock import MockerFixture
 from PySide6.QtDBus import QDBusMessage, QDBusConnection
 
 from pupgui2.dbusutil import *
-from pupgui2.constants import DBUS_DOWNLOAD_OBJECT_BASEPATH
 
-
-# NOTE: WOOHOO this mostly works!
+from pupgui2.constants import DBUS_DOWNLOAD_OBJECT_BASEPATH, DBUS_APPLICATION_URI
 
 
 def test_create_and_send_dbus_message(mocker: MockerFixture) -> None:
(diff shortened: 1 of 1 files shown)


sonic2kk:
CI is failing because it doesn't have `pytest_mock` as noted. However even when it does, I'll be interested to see how it feels about importing QtDBus :-) We mock the "real" calls so we shouldn't ever be sending any real data (we mock it to just be a dummy function that returns `True`), but still, I wonder if importing the module in CI has any Unforeseen Consequences™!

PR #546, opened 2025-05-06
Add entry point for GUI script

pastalian:
This makes it easier to launch the GUI when installing from source.

https://setuptools.pypa.io/en/latest/userguide/entry_point.html

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,3 +30,7 @@ install_requires =
     zstandard>=0.19.0
 
 packages = find:
+
+[options.entry_points]
+gui_scripts =
+    protonup-qt = pupgui2.pupgui2:main


sonic2kk:
So from reading, this allows us to run `python3 -m protonup-qt`, right? We can already run `python3 -m pupgui2`. What is the difference here, unless I'm missing something?

pastalian:
No, this entry point adds a convenient script named protonup-qt.
For example, if you install ProtonUp-Qt using `pip install .`, it will place a script to an appropriate location in your PATH, which looks like this:

```python
#!/usr/bin/python3
# -*- coding: utf-8 -*-
import re
import sys
from pupgui2.pupgui2 import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```

This allows users to run the GUI with just the `protonup-qt` command, without having to type `python3 -m pupgui2`.

sonic2kk:
Gotcha, my bad! I misread the docs as this allowing for `python -m timmins`, but missed the part about being able to run `hello-world` :sweat_smile:

Issue #555, opened 2025-07-20
add Proton-EM


ahjolinna:
Proton-EM focuses on wayland testing so it would be nice to have it for those who are interested 
https://github.com/Etaash-mathamsetty/Proton/releases

sonic2kk:
More of a gap in my knowledge but is this different than the Wayland support provided in GE-Proton10 currently? Perhaps it is more regularly updated?

Just wondering as if there's not too much difference, we could just keep GE-Proton. That's only my personal opinion though and personal curiosity :-)

Sn0whax:
Was juuuuust gonna make this request until I saw this. +1 for this.

PR #560, opened 2025-08-06
Ctmod: Add Proton-EM

sonic2kk:
Implements #555.

This PR implements Proton-EM for Steam, Lutris, Heroic Proton, and Bottles. We do this by subclassing the GE-Proton Ctmod installer, updating the `CT_URL` and `CT_INFO_URL` variables, and setting the Release Format to `tar.xz` (GE-Proton uses `tar.gz`).

Screenshot_20250806_173319

This Ctmod is gated behind Advanced Mode, as the main benefits of using it probably appeals more to "Advanced" users (i.e. you need to use commandline options for the features like Wine-Wayland support and so on, otherwise it's more or less just like vanilla Proton).

Thanks!


Sn0whax:
Based.

DavidoTek:
Thanks!

PR #582, opened 2025-10-08
Kron4ek: Only download amd64 builds

### LINKED ISSUE #581: wine-10.16-amd64-wow64 is downloaded under Kron4ek Wine Vanilla Builds instead of wine-10.16-amd64

**Describe the bug**  
wine-10.16-amd64-wow64 is downloaded under Kron4ek Wine Vanilla Builds instead of wine-10.16-amd64. 
**To Reproduce**  
wine-10.16-amd64-wow64 is downloaded under Kron4ek Wine-Builds Vanilla instead of wine-10.16-amd64.

**Expected behavior**  
I expected the standard amd64 build would be installed instead of the amd64-wow64 build.

 - **Platform**: Laptop
 - **System**: Mint 22.2
 - **Version**: 10.16-amd64-wow64
 - **How did you install ProtonUp-Qt?**: Flatpak

**Additional context**  
I only noticed what was off when using the amd64-wow64 build instead of amd64 broke one of my Wine prefixes, which was previously using wine-10.15-amd64, leaving me unable to even use Winetricks. I also opted to manually download wine-10.16-amd64 from GitHub after this.

sonic2kk:
Fixes #581.

This PR updates our `asset_condition` for Kron4ek Vanilla Wine builds so that it ignores the amd64-wow64 builds. When we search for assets we only check to include ones that have `amd64`, so `amd64-wow64` gets included and gets priority. Instead we should ignore these builds and only download `amd64` builds, as per #581.

Thanks!

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_kron4ekvanilla.py
+++ b/pupgui2/resources/ctmods/ctmod_kron4ekvanilla.py
@@ -38,7 +38,7 @@ class CtInstaller(GEProtonInstaller):
             'version', 'date', 'download', 'size'
         """
 
-        asset_condition = lambda asset: 'amd64' in asset.get('name', '') and 'staging' not in asset.get('name', '')
+        asset_condition = lambda asset: 'amd64' in asset.get('name', '') and not any(ignore in asset.get('name', '') for ignore in ['staging', 'wow64'])
         return fetch_project_release_data(self.CT_URL, self.release_format, self.rs, tag=tag, asset_condition=asset_condition)
 
     def is_system_compatible(self) -> bool:


DavidoTek:
Thanks!

Doc-4:
@sonic2kk Can this be changed into a "toggle visibility of wow64 versions" in future builds instead of disabling them entirely? There are some valid usecases some people have for wow64 right now, and steam frame will also need them. (iirc snapdragon 8 gen 3 is 64bit only, it'll likely need wow64 builds for said custom builds to work with 32bit applications)

DavidoTek:
> Can this be changed into a "toggle visibility of wow64 versions" in future builds instead of disabling them entirely? There are some valid usecases some people have for wow64 right now, and steam frame will also need them. (iirc snapdragon 8 gen 3 is 64bit only, it'll likely need wow64 builds for said custom builds to work with 32bit applications)

With https://github.com/DavidoTek/ProtonUp-Qt/pull/590, both amd64 and wow64 builds are displayed and downloadable by default.

PR #611, opened 2026-03-25
GameList: Add button to upgrade all GE-Proton games

solarisfire:
Won't be for everyone, but this is a serious help for anyone with a large Steam library, and a lot of games running on GE-Proton. Previously I was manually going through each row on the table, now it's one button click.

Adds a button to the Game List dialog for Steam that identifies the latest installed GE-Proton version and queues an update for all games currently using any version of GE-Proton to that latest version.

### DIFF
--- a/pupgui2/pupgui2gamelistdialog.py
+++ b/pupgui2/pupgui2gamelistdialog.py
@@ -55,6 +55,7 @@ class PupguiGameListDialog(QObject):
             self.setup_heroic_list_ui()
 
         self.ui.btnShortcutEditor.setVisible(self.launcher == 'steam')
+        self.ui.btnUpgradeGEProton.setVisible(self.launcher == 'steam')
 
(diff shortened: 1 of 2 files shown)


DavidoTek:
Thanks.

We already have the batch update feature, which allows upgrading all games from one GE-Proton version to another. It slightly differs from your implementation since it only upgrades from a single GE-Proton version to another, instead of upgrading all games using GE-Proton.
The feature can be accessed by selecting a GE-Proton version in the main window, clicking "show info" and then clicking "batch update" to move all games from the selected GE-Proton version to a newer version.

Were you aware of this? Do you think your approach has advantages over the current approach? I like that the button is in the game list since it is easier to find there, although it is specific to GE-Proton and not generic. It can also upgrade all GE-Proton versions, whereas the current implementation allows upgrading only from a certain version.

PR #615, opened 2026-05-12
Ctmods: Remove Proton-Sarek

### LINKED ISSUE #612: Remove Proton-Sarek (unsupported, replaced by proton-cachyos)
Proton-Sarek is no longer supported by me. It has been replaced by [proton-cachyos](https://github.com/CachyOS/proton-cachyos) as the supported Proton build shipping DXVK-Sarek.

sonic2kk:
Fixes #612.

## Overview

This PR removes Proton-Sarek from the list of available Ctmods. As noted in #612 it is no longer supported.

<hr>

It's been a while but I think this is all we needed to do to remove Ctmods. Let me know if I missed anything. Thanks! :smile:


sonic2kk:
I am not sure why tests are failing on CI, it seems like it may be a CI issue rather than a PR issue (the Translations sync PR is also failing on CI with the same error).

DavidoTek:
Thanks!

> I am not sure why tests are failing on CI, it seems like it may be a CI issue rather than a PR issue (the Translations sync PR is also failing on CI with the same error).

Hmm, I need to check why it's failing. It does not fail when I run the tests locally so I will merge it anyway.

> It's been a while but I think this is all we needed to do to remove Ctmods. Let me know if I missed anything. Thanks! 😄

I think there hasn't been much to do lately, but it's nice to see that everything is stable so far, and no formats have changed. Also, great to hear from you again.

DavidoTek:
> I am not sure why tests are failing on CI, it seems like it may be a CI issue rather than a PR issue (the Translations sync PR is also failing on CI with the same error).

The issue is fixed with https://github.com/DavidoTek/ProtonUp-Qt/pull/616

PR #624, opened 2026-06-24
Fix GE-Proton download

### LINKED ISSUE #623: GE-Proton download doesn't filter architecture
Just a heads up -- GE-Proton now has -aarch64 builds, I got a bunch of misreports due to ProtonPlus and ProtonUp-Qt downloading the aarch64 version instead of the standard x86 version. aarch64 builds and shasums end in -aarch64.

Jan200101:
fixes #623

The x86_64 builds don't mark themselves as such so its not possible to solve this in a nice way, instead we check if the release contains `aarch64` or not.

The logic is a bit primitive and could be simplifed into this:
```py
is_aarch64 = platform.machine().lower() in ("arm64", "aarch64")
is_aarch64_download = "aarch64" in asset["browser_download_url"]
return is_aarch64 == is_aarch64_download
```

but what this PR has is more readable.

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_00protonge.py
+++ b/pupgui2/resources/ctmods/ctmod_00protonge.py
@@ -5,6 +5,7 @@
 import os
 import requests
 import hashlib
+import platform
 
 from PySide6.QtWidgets import QMessageBox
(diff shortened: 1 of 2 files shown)


DavidoTek:
Thanks for the quick fix!

PR #644, opened 2026-09-10
fix: use correct ctmod configuration for Proton-RTSP

rs189:
Due to an invalid ctmod configuration, the Proton-RTSP versions would not show up in the compatibility tool selector. This fixes it.

### DIFF
--- a/pupgui2/resources/ctmods/ctmod_rtspgeproton.py
+++ b/pupgui2/resources/ctmods/ctmod_rtspgeproton.py
@@ -1,22 +1,22 @@
 # pupgui2 compatibility tools module
-# SpookySkeleton's RTSP-GE-Proton
+# SpookySkeleton's Proton-RTSP
 # Copyright (C) 2021 DavidoTek, partially based on AUNaseef's protonup
 
 from PySide6.QtCore import QCoreApplication
 
-from pupgui2.resources.ctmods.ctmod_00protonge import CtInstaller as GEProtonInstaller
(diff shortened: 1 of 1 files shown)


DavidoTek:
What exactly is the fix here? Proton-RTSP is only displayed when advanced mode is enabled. The installation works fine on my side.

The rename from `RTSP Proton` to `Proton-RTSP`, the fixed URL, and the added support for Lutris and Heroic are fine.

What about the removal of `advmode`, the rename from `GEProtonInstaller` to `ProtonGECTInstaller`, and the changed description?

rs189:
My apologies, I was not aware of there being an “advanced mode”, and perhaps that is another issue. 
The only reason I had to use ProtonPlus in addition to ProtonUp (I much rather prefer ProtonUp), was that, I was under the assumption that there was no support for Proton-RTSP, and so were others I know who rely on Proton-RTSP.
Now that you mention it, I am wondering what determines whether a specific Proton version should be hidden behind the advanced options or not?

DavidoTek:
No worries. Yeah, advanced mode is a bit hidden.

There is no real rule behind it. Basically, new compatibility tools and forks were added as "advanced" tools. Maybe we should remove it at some point since other users were also confused by it.

Can you add `advmode` again for now and change it back to `GEProtonInstaller`? I will merge this then, thanks.

rs189:
That should be done now. 
And yeah, looking at this again, it took me a minute to realize there was an advanced mode in the "About" section.

